PR #61, opened 2019-01-21
Change key-gen command back to simpler version

brockfanning:
Going through the quick-start again, it seems that this PEM change actually does cause a problem with Github. So I think we need to revert it. More investigation may be needed into when/where it is needed.


dougmet:
If I remember rightly the issue here was with Windows? I've had no trouble from Windows creating the keys without pem but where I fell over was the `travis encrypt` stage. It never works and I need to use either the linux sub-system, Docker, or find a Linux box/Mac

brockfanning:
Thanks @dougmet. The weird thing is that I'm getting conflicting reports of whether or not this is a problem, in both cases from people using Windows.

PR #65, opened 2019-01-23
move data acquisition to top of layout fixes #64

### LINKED ISSUE #64: title text not finding goal info
As shown in the UK version the title text is pulling through the wrong goal information:

image

Here the short text is for goal 1 not goal 2.

The reason is that on line https://github.com/open-sdg/open-sdg/blob/2d04ac1dad5af4e2fd0925dc9259149e720d6169/_includes/head.html#L17 where we assign the goal information, it uses `page.sdg_goal` which no longer exists at the page level. The page only knows:

```yaml
indicator: 2.3.2
layout: indicator
permalink: /2-3-2/
```

And the rest comes from the meta data which first appears below the `head.html` include https://github.com/open-sdg/open-sdg/blob/2d04ac1dad5af4e2fd0925dc9259149e720d6169/_layouts/indicator.html#L8

So we either need to remove the goal info or pull in the meta data higher up so that the `head.html` include can use it too.

dougmet:
The indicator page has to get a lot of data that's been imported from the data sets and the metadata. This PR loads that data earlier so that the metadata about goals can be pulled through to the page head.

### DIFF
--- a/_includes/head.html
+++ b/_includes/head.html
@@ -1,7 +1,6 @@
 <!DOCTYPE html>
 <html lang="en" class="no-js">
     <head>
-        {%- include multilingual.html -%}
         <script>var translations = {};</script>
         <script src="https://cdnjs.cloudflare.com/ajax/libs/modernizr/2.8.3/modernizr.min.js"></script>
         <!-- Basic Page Needs
--- a/_includes/head.html
+++ b/_includes/head.html
@@ -14,9 +14,8 @@
         <meta name="viewport" content="width=device-width, initial-scale=1.0">
         <!-- Title and meta description
         ================================================== -->
-        {% assign goal = site.goals | where: "sdg_goal", page.sdg_goal | first %}
         {% capture page_title %}
-        {{ t.general.indicator }} {{ page.indicator }} - {{ t.general.goal }} {{ goal.goal }} - {{ goal.short }}
+        {{ t.general.indicator }} {{ page.indicator }} - {{ t.general.goal }} {{ goal_number }} - {{ goal.short }}
         {% endcapture %}
         <title>{% if page.indicator %}{{ page_title | escape }}{% else %}{{ site.title | escape }}{% endif %}</title>
         <meta name="description" content="">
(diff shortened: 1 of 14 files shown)


dougmet:
This will need a little testing. It fixed the issue in my local build.

dougmet:
Yes good point. I'm not sure I totally like how global variables pass around the various includes.

I do like the idea of the variables include. Should make the layout more readable too.

brockfanning:
Agreed, the global variables are not great developer experience. Maybe we call it "global_variables.html" to make it very clear what it is? They could probably also have a naming convention, like maybe starting them with "_" (like, _meta and _t). Meanwhile we can start thinking about whether there is a possible way to avoid using them.

brockfanning:
@dougmet Are you leaning towards waiting for the more involved fix we're discussing, or starting now with the quick fix in this PR?

dougmet:
Yes I think so. But it doesn't need to be too involved. At the very least I'll move things to that `global_variables.html` file and that will make things more readable. As this bug is live I should probably get you something asap.

I vaguely recall there's a way to "pass" variables into include files. I'll double check, that might be one way around this.

dougmet:
Yep. I noticed that as soon as I'd done it. Yes OK, then that would go in `default.html` and then in `indicator.html` it would go:

```
{% include global-variables.html %}
{% include indicator-variables.html %}
```

dougmet:
OR

We could just have global variables and then some control logic. I think I prefer the first one.

brockfanning:
@dougmet I agree with your preference there.

brockfanning:
@dougmet One note though - I don't think that just putting it in default.html would be enough. I think it would need to be added to the top of all the layouts.

dougmet:
hi, ignore commits and future builds until I say so. My testing system currently involves pushing to this branch and then building the site from it. If necessary I can close the PR and reopen when ready.

brockfanning:
Cool, no need to close/reopen.

dougmet:
ok good. I need a better flow for this.

dougmet:
Closing for #76

PR #68, opened 2019-01-30
High contrast loading gifs

### LINKED ISSUE #4: Missing "loading" image in high-contrast mode
When loading an indicator page in high-contrast mode, an asset can't be found: `/assets/img/high-contrast/loading.gif`

We should add this asset if still needed, or otherwise remove the link to it.

brockfanning:
Fixes #4


brockfanning:
@dougmet I came across this gif which might work for the high-contrast background. I used some online tools to resize/crop, so hopefully the quality is OK. I actually think it's not a huge deal though -- it's kind of a javascript race condition thing: There is javascript that shows the loading gif, and there is also javascript that converts all images to high-contrast versions. This only really matters if the latter executes before the former. But we might as well get the images in there.

PR #69, opened 2019-01-30
Two column goal pages, with targets on left

### LINKED ISSUE #54: Incorporate "targets" into the platform
This idea was brought up on https://github.com/dawncomer/open-sdg-site-starter/issues/3 and would make sense as an addition to the core platform. Right now we are not displaying "targets" -- only goals and indicators.

Probably the most straightforward way to incorporate targets would be to update the goal.html layout to group the listed indicators by target:

```
homepage
└─goal 1 page
  └─all indicators in goal 1, grouped by target
└─goal 2 page
  └─all indicators in goal 2, grouped by target
└─etc...
```

A more involved alternative might be to create new pages for each target, and have the goal pages list targets, and the target pages list indicators.

```
homepage
└─goal 1 page
  └─target 1.1 page
    └─all indicators in target 1.1
  └─target 1.2 page
    └─all indicators in target 1.2
  └─etc...
└─etc...
```

We might start with the first "grouping" approach, since it is less drastic of a change? There are also mockups on the issue linked above.

brockfanning:
This is one approach to integrating targets into the platform - by changing the goal pages from a grid layout to a two-column layout, with targets on the left and indicators on the right. This is what it would look like:
two-column-goals-uk

This is the kind of change that would probably warrant a major version bump, or at least a minor version bump, since it changes a layout and CSS. For example this could cause a problem for any site already overriding the goal.html layout, such as the UK's site.

This also depends on the latest sdg-translations build, for a translation of "Targets".

Fixes #54

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -23,7 +23,7 @@ test.prep:
 	cd tests && npm install
 
 test.serve:
-	# Serve the Jekyll site at http://127.0.0.1:4000/
+	# Serve the Jekyll site at http://127.0.0.1:4000/open-sdg-site-testing/
 	cd starter && bundle exec jekyll serve --detach --skip-initial-build
 
 test.html:
--- a/Makefile
+++ b/Makefile
@@ -23,7 +23,7 @@ test.prep:
 	cd tests && npm install
 
 test.serve:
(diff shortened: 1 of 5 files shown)


brockfanning:
Closing this in favor of the optional approach in #118 which includes an alternative layout for goal pages.

PR #72, opened 2019-02-01
Initial text for FAQ page

brockfanning:


dougmet:
Hi @brockfanning is there something we need to kick to get readthedocs to rebuild? The FAQ isn't pulling through at the moment. Haven't used RTD before so not sure how it works.

brockfanning:
@dougmet I just re-built manually. I'm guessing that since I signed up with RTD with my own account, it only auto-builds when I'm the one doing the merging. I'll look into how to add the rest of the team.

dougmet:
Awesome, thanks. I see that now.

AnnCorp:
Hi @brockfanning if you have any suggestions before you go on leave on how we can rebuild this manually then we can try to add more to the FAQ page while you're away.

PR #76, opened 2019-02-05
move indicator and multilingual variables to own files #64

### LINKED ISSUE #64: title text not finding goal info

dougmet:
- This removes some unused variables from indicators.html
- Multilingual.html is not directly included in the head. It must be
added to the top of any layout now.
- Indicator variables move about the head.html include to fix #64
- layouts adjusted to directly include multilingual.

Replacing #65

### DIFF
--- a/_includes/head.html
+++ b/_includes/head.html
@@ -1,7 +1,6 @@
 <!DOCTYPE html>
 <html lang="en" class="no-js">
     <head>
-        {%- include multilingual.html -%}
         <script>var translations = {};</script>
         <script src="https://cdnjs.cloudflare.com/ajax/libs/modernizr/2.8.3/modernizr.min.js"></script>
         <!-- Basic Page Needs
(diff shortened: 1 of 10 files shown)


dougmet:
As discussed in #65 some refactoring of global variables has been done. In the end I have kept `multilingual.html` which goes at the top of every layout, and then `indicator-variables.html` which is only used on the indicator layout. There didn't seem need yet of a `global-variables.html`.

I do like the idea of some further refactoring, reducing the number of repeated variables, and using a global naming convention such as `_t`, `_meta` etc. However, as this is a bug fix for a live bug I wonder if we can come back to that?

PR #79, opened 2019-02-06
Update changelog.md

dougmet:
Getting prepared for a patch release. 

I've documented changes to the documentation, but I wonder if we really need this going forwards? The docs don't affect the theme and they are built separately for the readthedocs site.

We could consider only discussing code changes to keep it simpler.

### DIFF
--- a/docs/changelog.md
+++ b/docs/changelog.md
@@ -1,5 +1,15 @@
 <h1>Change Log</h1>
 
+### 0.3.1
+
+Patch release with some accessibility and bug fixes.
+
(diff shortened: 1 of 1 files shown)


brockfanning:
@dougmet That's a great point, I would be in favor of ignoring docs changes.

dougmet:
Mustn't merge until #2 and #4 are resolved

dougmet:
Have rebased from master so that the tests should work now. Also shrank 3 commits down to 1

PR #82, opened 2019-02-13
add q= to search query url

### LINKED ISSUE #81: Can search query follow standard get string?
Right now if you search for, say gold, you get:
`https://sustainabledevelopment-uk.github.io/search/?gold`

A more standard string would be:
`https://sustainabledevelopment-uk.github.io/search/?q=gold`

And indeed Google Analytics needs it in this format to understand searches (apparently). So could we alter `_includes/assets/js/search.js`, and maybe `search.html` if required, to use this format?

dougmet:
* Fixes #81
* Allows Google Analytics to understand searches
* Also update tests in relevant places

### DIFF
--- a/_includes/assets/js/search.js
+++ b/_includes/assets/js/search.js
@@ -31,7 +31,7 @@ var indicatorSearch = function(inputElement, indicatorDataStore) {
 
     var results = [],
         that = this,
-        searchString = unescape(location.search.substring(1));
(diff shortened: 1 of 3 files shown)


dougmet:
Main reason to block this is if it would break anything for anyone else?

dougmet:
image

dougmet:
Great, thanks!

PR #88, opened 2019-02-22
Support for multilingual tags

brockfanning:
I realized too late that the recently-added support for "tags" on indicators does not get translated. This update fixes that. It relies on the current method for translating metadata (subfolders of "meta" in the data repository). To see it in action:
* Go [here](http://brock.tips/sdg-site-maptest/no-poverty/) and switch to Spanish
* Also see the [English metadata](https://github.com/brockfanning/sdg-data-maptest/blob/develop/meta/1-1-1.md) and the [Spanish metadata](https://github.com/brockfanning/sdg-data-maptest/blob/develop/meta/es/1-1-1.md)

### DIFF
--- a/_layouts/goal.html
+++ b/_layouts/goal.html
@@ -26,6 +26,10 @@
 
   {% assign goal_indicators = site.data.meta | where: 'sdg_goal', page.sdg_goal | sort: 'indicator_sort_order' %}
   {% for indicator in goal_indicators %}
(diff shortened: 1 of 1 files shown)


brockfanning:
Thinking more about how this would actually work, I'm going to close it and come up with a different approach. The multi-file approach is way too cumbersome in this case.

PR #90, opened 2019-02-22
Fix image path of logo in docs

brockfanning:


brockfanning:
@AnnCorp Apparently I messed up the image path on the homepage of the docs. This should fix it.

PR #94, opened 2019-02-23
Switch to using the new 't' filter for menu item labels.

brockfanning:
This uses a new "t" Liquid filter provided by the jekyll-open-sdg-plugins Ruby Gem to translate the menu item labels, instead of the rather confusing Liquid code that's doing it now.

**This is a breaking change!** It requires sites to be loading the jekyll-open-sdg-plugins Ruby Gem.


brockfanning:
Tests here will fail until #95 is merged.

PR #96, opened 2019-02-23
Run tag labels through the t filter

### LINKED ISSUE #93: Indicators tags don't get translated
Indicator "tag" functionality was recently added, but I forgot to work them into the multilingual setup. Let's figure out a good way to make these tags easily translatable, preferably without any duplication of code.

brockfanning:
This allows for translated indicator tags. It relies on a Liquid filter "t" which is provided by the jekyll-open-sdg-plugins Ruby Gem. But the way Jekyll treats missing filters is to just fail gracefully, so if the filter is not there then nothing will change. So, this is not a breaking change.

Fixes #93

### DIFF
--- a/_layouts/goal.html
+++ b/_layouts/goal.html
@@ -63,7 +63,7 @@
               <ul class="tags">
               {% for tag in indicator.tags %}
                 {% assign tag_class = tag | slugify %}
-                <li class="tag-{{ tag_class }} warning">{{ tag }}</li>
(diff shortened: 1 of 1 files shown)


brockfanning:
That's right - more details in #101. Also if you're curious, here's that part of the code in the plugin: https://github.com/open-sdg/jekyll-open-sdg-plugins/blob/master/lib/jekyll-open-sdg-plugins/translate_key.rb#L45

PR #97, opened 2019-02-23
Gemspec file

### LINKED ISSUE #92: Make this theme a Ruby gem
This repository is a Jekyll theme which, so far, is being used with the help of the `jekyll_remote_theme` plugin. But we could also make this theme a Ruby gem, and allow users to use it in the normal way, without using `jekyll_remote_theme`. Is there any reason we should not add the required .gemspec file and push this to rubygems.org?

brockfanning:
Fixes #92


brockfanning:
Just a note that I've been holding off on this because I want include actual email addresses in the "authors" section.

brockfanning:
Actually on looking at some other projects, it seems like that field can be arbitrary.

PR #98, opened 2019-02-24
Avoid double-quotes in the manually-generated json files

### LINKED ISSUE #66: Double-quotes in indicator title translations break the search function
If there are double-quotes in the translated titles of indicators, the JSON at /indicators.json is invalid, which breaks the search function. We may need to replace double-quotes with single quotes.

brockfanning:
Fixes #66

### DIFF
--- a/_layouts/indicator-json.html
+++ b/_layouts/indicator-json.html
@@ -7,7 +7,7 @@
   {
     "goal": {
       "id": "{{goal_number}}",
-      "title": "{{translated_goal.title}}",
+      "title": "{{translated_goal.title | strip_newlines | strip | replace: '"', "'"}}",
       "indicators": [
         {%- assign all_meta = site.data.meta -%}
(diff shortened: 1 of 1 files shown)


brockfanning:
As I've got it in the PR this is catching double-quotes in the translations of indicator titles, goal titles, as well as the metadata field "national_indicator_description".

I did try converting `"` to `\"`, but that resulted in an error during the Jekyll build. I also tried running those strings through the "jsonify" Liquid filter, but that apparently only works for objects, and not for individual strings. I even went the route of writing a plugin called "jekyll_jsonify_string" but in the end decided that just catching the double-quotes was accomplishing the same thing.

I think in general it would be best if we didn't have to manually construct the json here, and instead could just run an already-built object through the "jsonify" filter. So this is more of a quick fix so that search will work even if there are double-quotes in the content.

PR #99, opened 2019-02-25
Remote goal icons

### LINKED ISSUE #83: Should goal icons be moved to sdg-translations?
Right now we're including translated goal icons in this project, under assets/img/goals. As we add more languages, there are going to be many more there. Is this OK? Or should we consider moving them out? One alternative might be to move them to a separate repository, like sdg-translations, and use it like a CDN. I've started a PR over there in case this is a good idea: https://github.com/open-sdg/sdg-translations/pull/49

brockfanning:
This is to show an approach for removing the goal icons from this project. This points to sdg-translations for the icons.

**This is a breaking change!** Sites that are already overriding these layouts would need to update them accordingly, and also add the `goal_image_base` setting to their _config.yml files.

Fixes #83


brockfanning:
Note that this PR needs discussion before merging. @dougmet @philipashlock any thoughts are welcome. Would there be any policy (or other) problems with pointing to `open-sdg.github.io/sdg-translations` for goal icons, on the production sites.

dougmet:
I can't think of any major objections. It's just moving things around within the project as a whole.

_layouts/frontpage.html
```diff
@@ -15,13 +15,13 @@ <h1>{{site.country.name}} {{ t.frontpage.heading }}</h1>
         {% cycle 'add row' : '<div class="row no-gutters">', '', '', '', '', '' %}
             <div class="col-xs-4 col-md-2">
                 <a href="./{{ default_goal.short | slugify }}/">
-                  <img src="{{ site.baseurl }}/assets/img/goals/{{ current_language }}/{{ goal_number }}.png" id="goal-{{ goal_number }}" alt="{{ translated_goal.short }} - {{ t.general.goal }} {{ goal_number }}" />
+                  <img src="https://open-sdg.github.io/sdg-translations/assets/img/goals/{{ current_language }}/{{ goal_number }}.png" id="goal-{{ goal_number }}" alt="{{ translated_goal.short }} - {{ t.general.goal }} {{ goal_number }}" />
```
dougmet: Would it be better to have this as a parameter? I even wonder if a whole translations section of the `_config.yml` would be useful.

brockfanning:
@dougmet That's a great idea. I've created a pull-request at https://github.com/open-sdg/open-sdg-site-starter/pull/23 to add that parameter to the starter site, and adjusted this PR to use it. Tests will fail here until that other PR is merged.

brockfanning:
Actually, I keep forgetting that the config for the tests is in this repository. Tests are passing now.

brockfanning:
I'm thinking it is OK to merge this, so it can go into the next release, which is looking like it's going to have a few breaking (but manageable) changes. Any countries that want the previous behavior can add the icons to their own site repositories and choose their own `goal_image_base` setting.

PR #101, opened 2019-02-25
Run country name/adjective through the "t" filter.

### LINKED ISSUE #62: Country name/adjective is not translated
The _config.yml variable for country name/adjective is inserted into a translated "tag line" on the homepage (among other places) but the name/adjective itself is not translated. We need to think about how to translate _config.yml variables like this, or alternatively, to move these kinds of variables into the translation system, instead of _config.yml.

brockfanning:
Also, include the country adjective IN the frontpage header, rather than putting them side by side.

This depends on https://github.com/open-sdg/sdg-translations/pull/56 so it should not be merged until that issue has been merged.

The "t" filter referenced here is provided by the jekyll-open-sdg-plugins Gem. If a site is missing that plugin, the "t" filter will do nothing (fail gracefully).

Fixes #62

### DIFF
--- a/_layouts/frontpage.html
+++ b/_layouts/frontpage.html
@@ -4,10 +4,9 @@
 <div id="main-content" class="container goal-tiles" role="main">
 
     {% assign country_name = site.country.name | t %}
-    {% assign country_adjective = site.country.adjective | t %}
-    <h1>{{ t.frontpage.heading | replace: '%adjective', country_adjective }}</h1>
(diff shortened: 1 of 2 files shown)


brockfanning:
@dougmet Right, if the string is not actually a translation key, then the string is output as-is. So, `{{ "general.goal" | t }}` will get output as a translation of the word "Goal", according to the current language. But `{{ "Goal" | t }}` will always be output as "Goal", regardless of the current language.

PR #102, opened 2019-02-26
Translate the output of metadata fields

brockfanning:
This change adds a new way to have multilingual metadata. First I'll describe the existing way:

The `meta` folder in the data repository can have subfolders for each language, and corresponding .md files can be put there to extend/override an indicator's metadata for that particular language. This is a fairly straightforward way to do it, and is great for translating one-off bits of text, like descriptions, keywords, etc.

The case where that is **not** great is for translating common words/phrases. If every single indicator needs "Office of Statistics" translated, it is a bit duplicative to have to copy/paste the translations of "Office of Statistics" for every single indicator.

Now I'll describe what this PR adds:

This PR offers a new option which works better in that use-case. Instead of making subfolders and extra files, and country could replace "Office of Statistics" with something like "custom.office_of_statistics". Then if they have a translation file (ala SDG Translations) called "custom.yml" which includes a key called "office_of_statistics", then the phrase will automatically get translated into all the languages.

So essentially this is a way to use the sdg-translations-style workflow instead of the metadata subfolders.

Note also that the "t" filter being used here fails gracefully -- if it does not exist, then behavior will be the same as before. (The "t" filter is part of the jekyll-open-sdg-plugins Gem.)

### DIFF
--- a/_includes/components/metadata.html
+++ b/_includes/components/metadata.html
@@ -13,7 +13,7 @@
           {% if indicator_metadata.name contains "_url" or indicator_metadata.name contains "_link" %}
             {% assign url_text_name = indicator_metadata.name | replace_first:'_url','_url_text'  %}
             {% assign url_text_name = url_text_name | replace_first:'_link','_link_text'  %}
-            {% assign url_text = meta[url_text_name] %}
+            {% assign url_text = meta[url_text_name] | t %}
             {% unless url_text %}
               {% assign url_text = 'Link' %}
             {% endunless %}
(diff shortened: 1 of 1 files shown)


dougmet:
Just so I understand correctly. If you have small phrases that need translation, you can use a translation key in that field. If you have a long description, then you can still use a language override. Is that correct?

brockfanning:
@dougmet Right, both approaches -- the "subfolder" approach, and the "translation key" approach -- are available to be used. A country could translate their metadata entirely with the subfolder approach, or entirely with the translation key approach. My distinction above is more an opinion about when one approach might be preferable to another.

Basically, if a particular translation is going to be repeated across many many indicators, then I personally would prefer the translation key approach -- because you can avoid copy-pasting (duplicating) the translated text into many different files.

But the subfolder approach is probably a bit more straightforward and easy to visualize/understand. So many people might prefer that approach in a lot of cases.

PR #111, opened 2019-03-07
Reporting Status from data service

### LINKED ISSUE #112: Allow different reporting statuses and colours
This is the open-sdg version of ONSdigital/sdg-indicators#2851

Requirement is to allow countries to use different reporting statuses without needing to override open-sdg. In the first instance they will need to be statuses that have a translation in sdg-translations.

To fulfil this we're configuring open-sdg/sdg-build#8 to output reporting statistics at data-build so that the Jekyll site can just pull these in.

dougmet:
**Before this is merged**:

* open-sdg/sdg-build#14 fixing open-sdg/sdg-build#8 needs to be merged and released.
* the backlog of translation PRs need to be merged in here and smoothed over.

**What's in this PR**:

The data service can now build all of the reporting statistics on the backend and serve them in a new `stats` route. This PR uses this to build the reporting page without needing JavaScript. This brings a couple of main benefits:

* Allows for countries to pick and choose statuses to use (e.g. UK doesn't want `inprogress`)
* Works for users who don't have JavaScript enabled.
* Faster page load.

The reporting service can also be used by other sites. For example someone may wish to integrate it into their own site as an SDG widget.

**Breaking Changes**

* The `prose.yml` file in the data service needs to have statuses in the right order and preferably with translations keys listed (although I'll ship with defaults in the first release).
* Must be using the latest version of `sdg-build` otherwise this will break.

Fixes #112 
cf ONSdigital/sdg-indicators#2851


brockfanning:
@dougmet To get the tests to pass, it might be necessary to add the stats route to [this file](https://github.com/open-sdg/open-sdg/blob/master/tests/_config.yml). I just updated open-sdg-data-testing to sdg-build 0.3.0, so the file is now available here: https://open-sdg.github.io/open-sdg-data-testing/staging/stats/reporting.json

dougmet:
Ready for review. But do give it a thorough read as it does mean breaking changes.

dougmet:
In the search results I've used the status as a css class. Everywhere else I use the slugified status (because it's easy in Jekyll but not with the underscore templating. But all of our allowed statuses don't need slugify so it works without this. I personally think this is OK because search is likely to change before the statuses change. But it's one to look out for if any weird status strings appear.

dougmet:
Would it help if I start the 0.5.0 changelog here before merging?

brockfanning:
@dougmet It would definitely help to document the breaking changes here. But it would be fine as just a comment here that we use in the release notes later.

brockfanning:
@dougmet Or maybe you could start the PR for the changelog update?

PR #117, opened 2019-03-10
Translate disaggregation filters/columns

### LINKED ISSUE #5: Allow for translation of data and metadata
Right now the platform can be shown in multiple languages, but the metadata for each indicator is only available in one language. We should think about the best way to allow multilingual metadata per indicator.

Along the same lines, the data (column names and row values) is not yet translated either, when it appears in the form of the disaggregation dropdown filters. Ideally whatever solution we use for metadata should also work for data.

brockfanning:
This allows for the translation of the disaggregation data columns/values, including the filters on the left, the columns of the data tables, the CSV exports, and the chart legends.

For indicators this depends on the translation keys being in a "data" translation group. I've started an [issue](https://github.com/open-sdg/sdg-translations/issues/70) in the SDG Translations project to do some translating of common disaggregation-related words.

This has a breaking change in the `head.html` include. If any implementer has overridden head.html then they will need to update it.

Also included are relevant documentation updates and Cucumber tests.

Fixes #5

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -14,6 +14,8 @@ test.prep:
 	cp tests/_config.yml starter
 	# Copy all the theme files into the starter.
 	cp -r -t starter/ _includes _layouts assets
+	# Copy any custom data into the starter.
+	cp -r tests/data starter/
 	# Add a second language.
 	cd starter && python scripts/batch/add_language.py es
 	# Build the Jekyll site.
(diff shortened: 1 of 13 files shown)


Norric1Admin:
@SavvasStephanides Could you give your thoughts on this please?

dougmet:
Couldn't quite trace through to answer this myself. Does it pull in all languages for a given topic, or just the ones that the site uses?

brockfanning:
@dougmet It only renders the translations for the current page. The mechanism for this is already in place [here](https://github.com/open-sdg/open-sdg/blob/master/_includes/multilingual-js.html). That include can be used either to render entire "translation group" (eg, "general") or even just a specify translation key (eg, "general.goals"). [Here's an example](https://github.com/open-sdg/open-sdg/blob/master/_layouts/indicator.html#L7) of where that include is used.

brockfanning:
@dougmet @SavvasStephanides I'm hoping we can get this into 0.5.0. Any chance you would have time to review before Friday? If not, no worries.

PR #118, opened 2019-03-14
Add an optional layout for a grouped-by-target goal page

### LINKED ISSUE #54: Incorporate "targets" into the platform

brockfanning:
This adds an optional goal layout for displaying targets in the left column and indicators in the right column.

Fixes #54

### DIFF
--- a/_layouts/goal-by-target.html
+++ b/_layouts/goal-by-target.html
@@ -1,0 +1,96 @@
+{%- include multilingual.html -%}
+{% include head.html %}
+{% include header.html %}
+{% assign goal_number = page.sdg_goal %}
+{% assign translated_goal = t.global_goals[goal_number] %}
+
+<div class="heading goal-{{page.sdg_goal}}">
+  <div class="container">
+    <div class="row">
+      <div class="col-xs-4 col-md-3 col-lg-2">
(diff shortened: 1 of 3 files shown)


brockfanning:
@Norric1Admin Great point, I just added documentation to this PR.

Norric1Admin:
Thanks!

PR #120, opened 2019-03-15
Better character handling in search code

### LINKED ISSUE #115: Search: no results when non-Latin characters are used
When a user performs a search, if there are non-Latin characters in the search query, there are no results found. It's not giving any error; it's just not finding results when it should.

brockfanning:
This use decodeURIComponent instead of unescape. Apparently this handles non-latin characters better, as it fixes search for Armenian/Russian.

Fixes #115


brockfanning:
It's straightforward, yes, basically just replacing one function with another. The old one, "unescape", is actually deprecated, so it's a good idea in general.

PR #123, opened 2019-03-15
Change to numbers for goal paths

### LINKED ISSUE #84: Goal URLs - should we switch to numbers?
The tests are failing now, with htmlproofer errors: https://circleci.com/gh/open-sdg/open-sdg/165

I believe it's due to some [goal short titles changing](https://github.com/open-sdg/sdg-translations/commit/67a90d0a8960c9395cf3d27bb0eca5c423258a0a) in the sdg-translations project. Some ways to deal with it:

1. Update the goal paths in the [open-sdg-site-starter](https://github.com/open-sdg/open-sdg-site-starter/tree/develop/_goals) repo to match the new short titles. We would also want to contact implementing countries to make sure their paths get updated too.
2. Revert the changes referenced above that affected the URL slugs - namely goals [11](https://github.com/open-sdg/sdg-translations/commit/67a90d0a8960c9395cf3d27bb0eca5c423258a0a#diff-bd649b1a46574edaedde33a51f9541faL8) and [16](https://github.com/open-sdg/sdg-translations/commit/67a90d0a8960c9395cf3d27bb0eca5c423258a0a#diff-bd649b1a46574edaedde33a51f9541faL26)
3. Update open-sdg in some way to lock down the goal URLs so that they are not dependent on the sdg-translations project. One obvious way would be to change them to numbers. Eg: `example.com/1` instead of `example.com/no-poverty`. This would also involve contacting implementing countries to help them update.

The second option above is the quickest fix, so I'll go ahead and do a PR for that. But thinking longer-term this is worth discussing.

brockfanning:
In #84 we discussed switching to numbers for goal paths. This PR makes this change.

Obviously this is a major breaking change, but the switch-over is not difficult to make. It just involves updating the 17 files in the `_goals` folder. For each file in the `_goals` folder, should replace the line:

```
permalink: /no-poverty/
```

to...

```
permalink: /1/
```

..and so on, for each of the 17 goal pages.

I also tested out the [jekyll-redirect-from](https://github.com/jekyll/jekyll-redirect-from) plugin, and it works great! Here are the basic instructions for that:

1. Add `gem "jekyll-redirect-from"` to the Gemfile
2. Add `- jekyll-redirect-from` in the list of `plugins` in `_config.yml`
3. Update each of the 17 goal files with a `redirect_from` field, like so:
    ```
    redirect_from:
      - /no-poverty/
    ```

This is a super easy way to make sure that the old goal paths keep working.

Fixes #84

### DIFF
--- a/_includes/indicator-variables.html
+++ b/_includes/indicator-variables.html
@@ -14,7 +14,7 @@
 {%- assign goal_number = meta.sdg_goal | downcase -%}
 {%- assign goal = site.data.translations[default_language].global_goals[goal_number] -%}
 {%- assign translated_goal = t.global_goals[goal_number] -%}
-{% capture goal_uri %}{{ site.baseurl }}{{ baseurl_folder }}/{{ goal.short | slugify }}{% endcapture %}
+{% capture goal_uri %}{{ site.baseurl }}{{ baseurl_folder }}/{{ goal_number }}{% endcapture %}
 {% capture goal_href %}{{ goal_uri }}{% endcapture %}
 {% capture goal_title %}{{ goal.title }}{% endcapture %}
 
--- a/_layouts/frontpage.html
+++ b/_layouts/frontpage.html
@@ -15,7 +15,7 @@
 
         {% cycle 'add row' : '<div class="row no-gutters">', '', '', '', '', '' %}
             <div class="col-xs-4 col-md-2">
-                <a href="./{{ default_goal.short | slugify }}/">
+                <a href="./{{ goal_number }}/">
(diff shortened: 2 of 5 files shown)


brockfanning:
Tests will actually fail because of the permalinks in the open-sdg-site-starter repository. Whenever we get closer to the time for the next Open SDG release (0.5.0) I will update open-sdg-site-starter and re-trigger these tests.

brockfanning:
@dougmet @SavvasStephanides If you're wondering why tests are failing, it's because I went ahead and [updated goal paths in the open-sdg-site-starter repo](https://github.com/open-sdg/open-sdg-site-starter/commit/f421fa542687dfba2fcf93c209a8bdf9d4498c0a), which is what this uses for running tests. So we'll need to merge this PR to get tests working again. (Assuming we want this change in 0.5.0, which I think is what we decided on.)

PR #124, opened 2019-03-16
Set scope=row on cells in the Year column

### LINKED ISSUE #106: Accessibility: Year cells should have scope=row
Moved from ONSdigital/sdg-indicators#1893

For screen readers a table is much easier to navigate if columns and rows are labelled. We're lucky because all of our datasets are time series so the Year column can be used as a row label (think about how quarters would affect this). To do this we change them to `<th>` tags.

```
<th scope="row">2015</th>
```

as here https://webaim.org/techniques/tables/data#th

brockfanning:
Fixes #106 

One note though: I feel like we should be using the jquery.dataTables library to generate table markup, instead of looping through the data to generate it with custom code. My first thought in implementing this was to use the jquery.dataTables options "columnDefs", "cellType", and "createdCell". This should have worked, but we're generating our table markup outside of jquery.dataTables. Not a pressing issue, but in the future it might be nice if we could trim our custom code and leverage jquery.dataTables more. All that said, there may well be a good reason for the way it is now, that I don't know about.

### DIFF
--- a/_includes/assets/js/indicatorView.js
+++ b/_includes/assets/js/indicatorView.js
@@ -439,7 +439,7 @@ var indicatorView = function (model, options) {
   };
 
   var setDataTableWidth = function(table) {
-    table.find('th').each(function() {
+    table.find('thead th').each(function() {
       var textLength = $(this).text().length;
       for(var loop = 0; loop < view_obj._tableColumnDefs.length; loop++) {
         var def = view_obj._tableColumnDefs[loop];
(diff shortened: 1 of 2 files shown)


dougmet:
Yes I'll take a look (mostly testing). Re `jquery.dataTables` I would support this. I can't say why it wasn't used, but my guess is one of 1. Keep external dependencies down 2. Something to do with progressive enhancement (although I think they just get a Jekyll built headline data table) 3. Tighter control over some element. 4. Didn't think of it.

Basically I'm up for it, although it would be nice to test and merge this first to get that accessibility feature out as early as possible.

brockfanning:
@dougmet I'm totally fine with waiting until later to do any refactoring related to jquery.dataTables.

PR #126, opened 2019-03-19
Sort the years array in maps, to fix first/last year functionality

brockfanning:
There is functionality, particularly in the year slider, that assumes the years array is sorted from first to last. But in practice it isn't necessarily sorted that way, so let's ensure that.


brockfanning:
@philipashlock @SavvasStephanides @dougmet @Norric1Admin Would be nice if we could get this into release 0.5.0, since we're already editing this file.

Issue #125, opened 2019-03-18
Support easy implementation of custom Google Analytics events


brockfanning:
Right now adding custom Google Analytics events involves a lot of overriding of Open SDG files. Can we allow for custom events without any overrides?

This was brought up [here](https://github.com/ONSdigital/sdg-indicators/issues/2921).

dougmet:
First step is to move the GA code out of the footer into an include. EIther a `ga.html` inside a script tag, or a `ga.js` which we include in `scripts.html` (so long as it's loaded after page render)

SavvasStephanides:
So we've had a lengthy session with @dougmet yesterday. Here are a few ideas (Doug, you can expand if I've left anything behind):

### 1. Element attributes

The main idea is to add the non-constant  `ga()` parameters (category, action and label) as attributes to the element we want to track. For example:

```html
<a data-toggle="tab" ga-event ga-category="tab change" ga-action="Change data view" ga-label="Change to Table tab"></a>
```

We then create a `Javascript` file that will pick these up and add event listeners to the elements:

```javascript
var elementsTrackedByGoogleAnalytics = document.querySelectorAll("[ga-event]");

elementsTrackedByGoogleAnalytics.forEach(addGoogleAnalyticsClickEventToElement);

function addGoogleAnalyticsClickEventToElement(element){
    var category = eachElement.getAttribute("ga-category");
    var action = eachElement.getAttribute("ga-action");
    var label = eachElement.getAttribute("ga-label");

    element.addEventListener("click", function(){
        ga("send", "event", category, action, label);
    });
}
```

### 2. Add IDs to elements

Another idea is to add IDs to elements and take care of the `ga()` in the `Javascript` file. This could be the default `html` ID (`<a id="">`) or we could implement our own "special" IDs. We would then create a `JSON` "dictionary" which would match the IDs to the `ga()` attributes. An advantage of this over method (1) is that we can avoid repetition for elements like tabs.

Example:

HTML:

```html
<a data-toggle="tab" ga-id="tabChange" ga-label="Change to Table tab">Table</a>
<a data-toggle="tab" ga-id="tabChange" ga-label="Change to Chart tab">Chart</a>
```

JSON:

```js
{
    "tabChange": {
        "category": "Tab change",
        "action":"Change data view"
    }
}
```
`Javascript` will add the relevant `ga()` click events similar to method (1) above.

Let us know what you think @brockfanning

brockfanning:
[...]

PR #128, opened 2019-03-25
Refactor Google Analytics code

dougmet:
This PR moves the google analytics code into a separate file `googleAnalytics.js` which can then be overridden without having to override the whole page footer.

I did sneak in another change, which I want to check with you @brockfanning. Rather than disabling `ga` for staging I've put it so that it checks for a `ga_prod` string in the config. If it has one, it'll use `ga`, if not, it won't. This is set from the line:

```
{% if site.analytics.ga_prod %}
initialiseGoogleAnalytics();
{% endif %}
```

This makes no changes for the UK. How about the US and other countries?

See #125


brockfanning:
@dougmet I believe empty strings are "truthy" in Liquid, so what would happen is that even if `site.analytics.ga_prod` is `''`, then everything would still run. Of course, without that GA number, it wouldn't impact anyone's analytics. So I think this is fine.

dougmet:
oh I see. OK, I thought they were falsy (desired effect) so that code does nothing. Then I should change it.

dougmet:
So in summary. Now, if you have `ga_prod` set, then the page view will get sent. Even on staging. So it's important to make sure that `site.analytics.ga_prod` is only set on the sites you actually want to monitor analytics on.

The bonus is that you can use a different GA ID to _test_ GA on the staging environment.

PR #133, opened 2019-03-27
Ignore SDMX-related fields

brockfanning:


dougmet:
ah. this implies we need something equivalent on the build side to remove the edges. Although seeing as we're moving everything over to the build eventually, perhaps we just note it on the ticket.

Not relevant here but need to think about what tickets we want on sdg-build to support column metadata. It's related to everything! Outputting to csvw, sdmx etc.

dougmet:
I've made a start with open-sdg/sdg-build#20

brockfanning:
Cool. I think we dismissed this possibility, but just to make sure we're considering everything, could we do this entirely in sdg-build? I think it would be great if we could, but I think there was a downside I'm not remembering?

dougmet:
The downside was that we want to ignore attributes on the chart, but other consumers may want it. So for downloading data sources, data apis etc. We should definitely ignore them for edges because the chart is the only user of that.

My ideal scenario would be that all the columns are there but they're labelled so that the chart only looks for things it understands (TimePeriod, Units, Dissaggregation, Value) and ignores things it doesn't know what to do with. It would make it more robust if data owners add weird things to the data as well.

brockfanning:
Cool, thanks for the explanation. Makes sense to me.

dougmet:
I've added a quick fix on the build side if we can line up in time open-sdg/sdg-build#22

brockfanning:
Closing, no longer needed.

PR #141, opened 2019-04-03
Added autotrack.js for Google Analytics

SavvasStephanides:
I believe this is the only change required in `open-sdg` to add eventTracker. The rest of the changes are in `sdg-indicators`.


brockfanning:
@SavvasStephanides Are the other changes the HTML attributes? I was thinking those would be included in this update.

Also do you think autotrack.js could be loaded via a CDN, so that it doesn't have to be committed to this codebase?

SavvasStephanides:
I think the HTML attributes are part of `sdg-indicators`. Either that or I can't locate them in `open-sdg`. 

As for loading the `autotracker.js` via CDN, there is the possibility of loading it as a raw file from GitHub but I'm not sure that would be recommended/safe.

brockfanning:
For the CDN, how about CDNJS? https://cdnjs.com/libraries/autotrack

I was thinking we would be adding the HTML attributes to open-sdg - do I have that wrong? That goal I think would be to prevent you from having to override a lot of stuff in sdg-indicators.

SavvasStephanides:
Yes that would work. Thanks for the CDN link. As for the HTML attributes, can you point me to the right direction of where the `html` files are in `open-sdg`? I can only find them in `sdg-indicators`. Thanks.

brockfanning:
Sure @SavvasStephanides, they should basically mirror the files in sdg-indicators -- ie, the same filenames in the same folders. For example, the open-sdg equivalent of [this file](https://github.com/ONSdigital/sdg-indicators/blob/develop/_layouts/indicator.html) is [this file](https://github.com/open-sdg/open-sdg/blob/master/_layouts/indicator.html). The `indicator.html` in sdg-indicators is currently "overriding" the `indicator.html` in open-sdg, by virtue of being the same filename in the same folder.

brockfanning:
@SavvasStephanides Just following up on this one. Were you waiting on anything from me?

SavvasStephanides:
Sorry @brockfanning. I've put this on hold for now to focus on the accessibility and other stuff.

PR #144, opened 2019-04-05
Survey text

Norric1Admin:
Adds survey header functionality

### DIFF
--- a/_includes/components/survey-message.html
+++ b/_includes/components/survey-message.html
@@ -1,0 +1,3 @@
+<div class="surveyMessage">
+    {{ t.header.survey_text | replace_first: '%survey_url', site.survey_url }}
+</div>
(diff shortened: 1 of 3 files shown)


brockfanning:
@SavvasStephanides @Norric1Admin I'm not sure about this one. We're not really adding any functionality here - we're just creating a new section for the header and calling it "survey-message". It could just as easily be called "header-extra" or anything else.

Personally I would like to know more about the experience of adding this to the UK platform. Since you've done it by overriding the header.disclaimer translation -- was that acceptable? Did it feel too hacky? An alternative might have been to override the [Jekyll include for the "dev-disclaimer"](https://github.com/open-sdg/open-sdg/blob/master/_includes/components/dev-disclaimer.html).

What I'm getting at is that this may be a feature that is best left to individual countries to implement, by either overriding a Jekyll template (or a translation, if preferred).

Norric1Admin:
Hello Brock, I had forgotten about this PR. We were trying to add a new survey message to the header, which has now been hard-coded as this approach didn't work on the UK site as hoped. Long-term, we'd like to find a more elegant method as i) this overlaps the menu tab when viewed in small windows or mobile devices and ii) similar links will be required in future, and we want consistency.

Closing this PR, but we'll continue to work on it from the UK side.

Thanks,
Chris

PR #145, opened 2019-04-08
Exclude column 'Unit measure'

Norric1Admin:
@brockfanning can we also exclude the 'Unit measure' field please?
P.S. Am not getting the option to update this branch, but I can see it is 33 commits behind master.


Norric1Admin:
Also requires https://github.com/open-sdg/sdg-build/pull/27
Related issue https://github.com/ONSdigital/sdg-indicators/issues/2997

PR #147, opened 2019-04-09
Configurable frontpage headings

### LINKED ISSUE #146: Make front page heading configurable
Right now we've got the front page heading (eg, "UK data for Sustainable Development Goal indicators") being generated by dynamically by inserting the config's `site.country.name` into a hardcoded phrase. This is proving to be problematic for translations. In some languages, the grammar/spelling/etc of the country name depends on how it is used, and so can't be dynamically "built" like that, with much accuracy.

Also, given how prominent that heading is, it is a common need to tweak it. The easiest way to do this is to override the frontpage.html layout, which is a bit overkill. So I'd like to propose that we make it into a _config.yml option. In the interests of backwards-compatibility we could make the current text show up by default, if the config option is not set.

brockfanning:
Fixes #146 

## Breaking changes

None.

## Description

See #146 for details of the problem.

## Changes needed in other projects

After merging, the [open-sdg-site-starter config file](https://github.com/open-sdg/open-sdg-site-starter/blob/develop/_config.yml) should have this added to it:

```
# To customise the frontpage heading, uncomment and change the following line:
#frontpage_heading: This is my frontpage heading

# To customise the frontpage instructions, uncomment and change the following line:
#frontpage_instructions: These are my frontpage instructions
```

## Files affected

* _layouts/frontpage.html
* assets/css/default.scss

## Other notes
* believe at one point, probably before the open-sdg revamp, the jump-to-search was truly a "progressive enhancement", in that the whole phrase was hidden until javascript revealed it. At the moment it's only part of the phrase that's hidden, and the markup is tied up in sdg-translations, so is difficult to change. So I'm not worrying too much here about progressive enhancement. Long-term I think we should remove this whole phrase from sdg-translations and eventually just rely on _config.yml for this phrase. It can be much more easily tweaked from there.

## Documentation added

No, the additions to the starter repo's _config.yml afterwards will serve as documentation.

## Tests added

Yes.

### DIFF
--- a/_layouts/frontpage.html
+++ b/_layouts/frontpage.html
@@ -4,9 +4,11 @@
 <div id="main-content" class="container goal-tiles" role="main">
 
     {% assign country_name = site.country.name | t %}
-    <h1>{{ t.frontpage.heading | replace: '%name', country_name }}</h1>
+    {% assign heading_default = t.frontpage.heading | replace: '%name', country_name %}
+    <h1>{{ site.frontpage_heading | default: heading_default }}</h1>
 
-    <p>{{ t.frontpage.instructions | replace_first: '%before_link', '<span style="display:none" id="jump-to-search"><a href="javascript:void(0)">' | replace_first: '%after_link', '</a></span>' | replace_first: '%name', country_name }}</p>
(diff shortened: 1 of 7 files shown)


brockfanning:
This is ready for review.

Norric1Admin:
@brockfanning I've made a [feature branch](https://sdg.mango-solutions.com/config-header) - what is the best way to test this?

PR #155, opened 2019-04-18
FAQ: mapping update

Norric1Admin:
@brockfanning are you happy with these example links and the reference to UK step-by-step guidance?

### DIFF
--- a/docs/faq.md
+++ b/docs/faq.md
@@ -22,9 +22,8 @@ be needed to do this. The [Customisation](https://open-sdg.readthedocs.io/en/lat
 ## Are maps available in Open SDG?
 
 Yes. Maps can be automatically generated from the data if GeoCodes are included and the
-required geographical boundary files are available. Links to maps in other systems can be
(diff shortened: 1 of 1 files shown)


brockfanning:
@Norric1Admin If urgent, I'm OK with that. But if there is time do you think we could try and get the step-by-step into the Open SDG docs? I've got a few notes about those docs, but I'll follow-up offline.

Norric1Admin:
Yes certainly. Perhaps we could get this version online for now, and then follow it up next week with more details? We need to review our own guidance over April/May and are likely to have more suggestions.

brockfanning:
My only concern is the Mapbox API token issue -- maybe that should be clarified/detailed in this update?

Norric1Admin:
Thanks Brock. What is the issue with Mapbox API tokens - does each Country have to generate a new one, or do the token permissions need to be updated for each new location? Interested to know how this works.

brockfanning:
Yes, each country needs their own token, if they want to use Mapbox.

Issue #142, opened 2019-04-04
Make source field text wrap like footnote text does


LucyGwilliamAdmin:
If source spans longer than the page width, it get's cut off rather than continuing on the next line:
![image](https://user-images.githubusercontent.com/43404358/55562020-14561580-56eb-11e9-9898-4583646a0226.png)

but it doesn't seem to be a problem if the footnote is longer than the page width:
![image](https://user-images.githubusercontent.com/43404358/55562063-246df500-56eb-11e9-9414-050ad2f96d7d.png)

Is there some sort of code that is applied to the footnote field that isn't applied to the source field?

brockfanning:
@LucyGwilliamAdmin This is a great point. I think it's a problem that the text above and below the graph are not actually HTML, but instead graphics being generated by Chart.js. If we could change them to be HTML, then they'd be more easily styled and would wrap predictably. It's also better to have them as HTML for user-experience (users could drag-select the text) and for SEO (Google would be able to index that text).

brockfanning:
@LucyGwilliamAdmin @Norric1Admin @SavvasStephanides I believe I have a fix ready for this, although it would be helpful if #166 were merged first, since it changes a lot of the same files.

brockfanning:
@LucyGwilliamAdmin @Norric1Admin @SavvasStephanides Ok, I've got #179 up for review, which should fix this. Note that the UK is overriding three key files, so if you want to test #179 against the UK site, these files will need to be deleted:
* _includes/assets/js/indicatorView.js
* _layouts/indicator.html
* assets/css/default.scss

LucyGwilliamAdmin:
@brockfanning great news! we'll have to review why we're over-riding them and go from there.

Do they necessarily need to be deleted? Or can we just copy over the changes?

brockfanning:
@LucyGwilliamAdmin Great question - sorry, I meant only that for the purposes of a quick test of the PR, you could just delete those files. But you're right, to eventually incorporate this into the UK site, if the files still need to be overridden, you would need to copy over the changes.

PR #166, opened 2019-05-13
All-around cleanup, documentation, and light refactor of Chart.js stuff.

brockfanning:


brockfanning:
This is ready for review.

brockfanning:
@Norric1Admin I tested this out with the UK platform and it worked fine. I then tried overriding the head.html with an older version of that file, and it broke the graphs.

So this PR is a breaking change for any site that is overriding `_includes/head.html`. I'm not aware of any countries doing so, but regardless we would need to make this clear in the release notes.

IMO this is ready for review/merging.

Norric1Admin:
I also [tested this](https://sdg.mango-solutions.com/cleanup-test/) and it's working fine (maps and all). Neither Rwanda, Ghana or Kyrgyzstan are using head.html, so I approve the merge.
Allows: https://github.com/open-sdg/open-sdg/issues/142

PR #167, opened 2019-05-15
Accessibility: update footer as unordered list

SavvasStephanides:
The links at the footer are now an unordered list (`<ul>`). I will also make a pull request on `sdg-indicators` for the `scss` updates.

### DIFF
--- a/_includes/footer.html
+++ b/_includes/footer.html
@@ -1,7 +1,7 @@
 <footer role="contentinfo">
   <div class="container">
     <div class="row">
-      <div class="col-xs-9">
+      <div class="col-xs-12">
         <div id="footerLinks">
           <ul>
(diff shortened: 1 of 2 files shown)


Norric1Admin:
@brockfanning Many of the existing fields use `!important` (24 uses in default.scss). Do you recommend removing them all, or only the new ones that Savvas added?

SavvasStephanides:
@brockfanning I like the idea of restricting the use of `!important` as much as possible but it's already being used quite heavily in `default.scss`. As this seems like a complex issue (but an important one), I think we should tackle it as its own issue altogether, as a "refactoring" ticket.

brockfanning:
Ok, sounds like a plan.

SavvasStephanides:
Thanks for the suggestion @brockfanning. Just updated the branch.

PR #173, opened 2019-05-19
More robust way to display indicator names

brockfanning:
This refactors a few places where indicator names are needed. Mostly it changes the code from relying on global indicator names, to using a new filter in jekyll-open-sdg-plugins that figures out the indicator name in a more forgiving way.

The upside of this change is that if countries are using national/proxy/etc indicators, the custom indicator name will appear instead of the current behavior, which is that the name is either global or missing altogether.

## Breaking change

This requires that the Gemfile for a site be updated to pull in at least version 0.0.10 of the jekyll-open-sdg-plugins Gem, like [this](https://github.com/open-sdg/open-sdg-site-starter/blob/develop/Gemfile#L8).

Otherwise there are a lot of changed files, but it won't break anything if they are overridden. One unnecessary variable has been marked as "deprecated" but left in for backwards compatibility.

## Tech notes

* The mechanism here is a Liquid filter called "get_indicator_name"
* This filter has a particular series of places that it looks for indicator names. The idea is that it finds the intended translation, but always falls back to something. Here is the order of places it looks for translations:
    1. A translation of the "indicator_name" metadata field
    2. If this is the default language, the un-translated "indicator_name" metadata field
    3. If this indicator is global, the translated global indicator name
    4. Fall back to the untranslated "indicator_name" metadata field
    5. Fall back to the indicator ID

### DIFF
--- a/_includes/indicator-variables.html
+++ b/_includes/indicator-variables.html
@@ -25,3 +25,8 @@
 {% endif %}
 
 {% capture dataset_name %}indicator_{{meta.indicator | slugify }}{% endcapture %}
+
+{% comment %}
+@deprecated: The following variable will be removed for 1.0.0.
+{% endcomment %}
+{%- assign translated_indicator = t.global_indicators[meta.indicator] -%}
--- a/_includes/indicator-variables.html
+++ b/_includes/indicator-variables.html
@@ -25,5 +25,3 @@
 {% endif %}
 
 {% capture dataset_name %}indicator_{{meta.indicator | slugify }}{% endcapture %}
-
-{%- assign translated_indicator = t.global_indicators[meta.indicator] -%}
--- a/_layouts/goal-by-target.html
+++ b/_layouts/goal-by-target.html
@@ -46,8 +46,7 @@
     </div>
     <div class="indicator-cards col-md-6 row no-gutters">
     {% for indicator in group.items %}
-      {%- assign translated_indicator = t.global_indicators[indicator.indicator] -%}
-      
+
       {% assign status_css = indicator.reporting_status | slugify %}
       {% if indicator.reporting_status == 'notapplicable' %}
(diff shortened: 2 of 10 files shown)


brockfanning:
@LucyGwilliamAdmin @Norric1Admin This might be a good one to have in before Kyrgyzstan. It could avoid some confusion when custom indicator names don't appear correctly. Note the note about the breaking change.

LucyGwilliamAdmin:
@brockfanning what do you mean by custom indicator names? What metadata field would this be?

brockfanning:
I should have said "non-global indicator names" maybe. For example if a country has a custom indicator like 1.1.1z (or whatever). Currently the indicator's name will not display in some layouts, because Open SDG relies on global translations of indicator names in some layouts. Since 1.1.1z is not a global indicator, it would just show up blank.

So this change is to ensure that non-global indicator names always show up.

LucyGwilliamAdmin:
I see - me and Abdulla have agreed that while we're in Kyrgyzstan this time around we will only focus on global indicators so no need to rush this through.

That being said I'm happy to help with the testing but I'm not sure how I would do so. If you can point me in the right direction, I can take a look.

brockfanning:
@LucyGwilliamAdmin Here's an example of a minimal way to test it: [commit](https://github.com/ONSdigital/sdg-indicators/commit/d68a8b3b1a2845d970cf97cc04a22edc339590a0)

Afterwards you could confirm that, for example, indicator names still show up correctly on goal pages, and that the indicator search function still works the same as before. Also, since sdg-indicators is overriding indicator.html, you could also test an indicator page to make sure that the backwards-compatibility is working (things look the same as before).

To see an example of where it's actually displaying custom names and non-global indicators, see the [Cambodia](http://brock.tips/sdg-site-cambodia/) example.

PR #179, opened 2019-05-22
Remove all "canvas" text and replace with HTML

### LINKED ISSUE #142: Make source field text wrap like footnote text does

brockfanning:
Fixes #142 

This changes the graph title and footer so that they are normal HTML instead of "canvas" text. This allows it to be drag-selected, wrap normally, and be styled more easily. For consistency, the table footer is also changed to match. A general cleanup of indicatorView.js is also included (removing empty functions and unused variables).

Backwards-compatibility note: This changes a lot of files, but doesn't break much if files are overridden. From my testing, the worst that can happen is that, if a country is overriding both the indicator.html layout and the default.scss file, the chart legend will bump up against the chart footer.

Below is are screenshots of the new charts/tables. The footer table is not using any new CSS -- instead it's using the out-of-the-box Bootstrap treatment of "description lists" provided by the "dl-horizontal" class.

remove-canvas-text--chart

remove-canvas-text--table

### DIFF
--- a/_includes/assets/js/indicatorModel.js
+++ b/_includes/assets/js/indicatorModel.js
@@ -43,9 +43,6 @@ var indicatorModel = function (options) {
   this.chartTitle = options.chartTitle;
   this.graphType = options.graphType;
   this.measurementUnit = options.measurementUnit;
-  this.dataSource = options.dataSource;
-  this.geographicalArea = options.geographicalArea;
-  this.footnote = options.footnote;
   this.showData = options.showData;
   this.selectedFields = [];
   this.allowedFields = [];
@@ -178,12 +175,6 @@ var indicatorModel = function (options) {
       spanGaps: false
     };
 
-    that.footerFields = {
-      'Source': that.dataSource,
-      'Geographical Area': that.geographicalArea,
-      'Unit of Measurement': that.measurementUnit,
-      'Footnote': that.footnote,
-    };
   }());
 
   var headlineColor = '777777';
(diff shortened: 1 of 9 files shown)


brockfanning:
@Norric1Admin @AnnCorp @LucyGwilliamAdmin Great points Lucy. I suspected there was a good reason for the way things were before, and I think your point about the image is spot on. That title and footer were likely being done as "canvas text" purely for the purposes of getting it into the downloaded image.

So before doing any more work on this -- is the image problem a deal-breaker? I think that getting that info into the downloaded image will be inordinately difficult if we make this change. So if having the title and footer in the download image is a requirement, I think we have to close this PR. Good topic maybe for an upcoming call?

LucyGwilliamAdmin:
@brockfanning Is there any way to keep the canvas but allow the source field to wrap like the footnote field does here?
image

I agree, we should discuss whether it is a deal breaker in a call when we're back from Kyrgyzstan.

brockfanning:
@LucyGwilliamAdmin I bet that is possible - I'll submit a separate PR with only that change so we can compare.

It's a shame to have to use canvas text, because of the drawbacks. But I agree that the image download with metadata is really nice. A possible solution might be a "Download image" button using something like [this](https://www.jqueryscript.net/other/Capture-HTML-Elements-Screenshot.html), but I've never tried that before.

brockfanning:
@LucyGwilliamAdmin Well, as much as I would love to remove canvas text and slice out large chunks of code... I have to admit that #181 fixes the problem in a much less-involved way. Can you try that one out?

brockfanning:
Agree we can close this and revisit if needed.

PR #181, opened 2019-05-23
Graph source text wrap

### LINKED ISSUE #142: Make source field text wrap like footnote text does

brockfanning:
Fixes #142 

This change treats the chart's "source" footer item in the same was as the "footnote" item, to let it wrap lines.

This is a quick alternative to the much more involved #179. A benefit of this approach is that the metadata still makes it into the downloaded image of the canvas.


LucyGwilliamAdmin:
@brockfanning I've left a review but as I don't have write permission it doesn't meet the criteria for you to merge this PR.

PR #183, opened 2019-05-24
Allow either multiple or single map layers

### LINKED ISSUE #178: Map error when there is only one layer
It appears we have some javascript that only works when there are multiple layers. Here is an example line: https://github.com/open-sdg/open-sdg/blob/master/_includes/assets/js/plugins/jquery.sdgMap.js#L277

When a site has only 1 map layer, that variable is not an array. So the `[0]` gets an undefined. More investigation and confirmation is needed.

brockfanning:
Fixes #178 

This makes sure that the variable in question is always an array of responses, even when there is only one. So it will behave the same whether there is one layer or multiple.

### DIFF
--- a/_includes/assets/js/plugins/jquery.sdgMap.js
+++ b/_includes/assets/js/plugins/jquery.sdgMap.js
@@ -257,8 +257,20 @@
       });
       $.when.apply($, geoURLs).done(function() {
 
+        // Apparently "arguments" can either be an array of responses, or if
+        // there was only one response, the response itself. This behavior is
+        // odd and should be investigated. In the meantime, a workaround is a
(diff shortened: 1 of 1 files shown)


brockfanning:
@LucyGwilliamAdmin I think that's right, but right now sdg-indicators is overriding the jquery.sdgMap.js file. So to test this you'll need to either delete that file or incorporate the changes into it.

LucyGwilliamAdmin:
@brockfanning thanks, I'm getting this error now:
image

brockfanning:
@LucyGwilliamAdmin In the map config, try increasing the max_zoom to 10.

LucyGwilliamAdmin:
Perfect!
image

PR #187, opened 2019-05-26
Use the site baseurl as a defult goal image base

### LINKED ISSUE #172: How to set site.goal_image_base to site.baseurl
The image URLs for the goal tiles on the frontpage are built from `site.goal_image_base`, which based on the starter-site, points to open-sdg.github.io. However, I want the images to be hosted locally in the repo, so all assets on on our own domain. There is no way to set `site.goal_image_base` to `site.baseurl` without overriding several includes.

brockfanning:
This allows sites to use local goal images, regardless of what the site baseurl is.

Fixes #172


brockfanning:
Just realized this is incomplete.

PR #201, opened 2019-06-11
Support zero values

### LINKED ISSUE #175: Should be able to handle zero-value stats
Right now Open SDG doesn't seem to be able to display stats where the value is 0.

brockfanning:
Fixes #175 

This allows "0" to appear in both charts and tables. Previously there were some "truthy" checks that were unintentionally filtering out zeroes. This change looks specifically for "null" instead, which lets the zeroes get through.

Minor enough bug fix, no docs or tests needed.


Norric1Admin:
@brockfanning does this also allow null values (e.g. '-') or are these converted to zero?

PR #202, opened 2019-06-13
A more precise check for data (disaggregation) translations.

brockfanning:
## Breaking changes

If any countries were relying on a behind-the-scenes lower-case conversion of disaggregations before being translated, those disaggregations would no longer get translated. No countries are actually doing this, and it fails gracefully and obviously, so I would not call this a "breaking change".

## Description

Right now there is a convenience check, when displaying disaggregation labels, which looks in a translation group called "data" for a lower-case version of the whatever the label is. Documentation exists [here](https://open-sdg.readthedocs.io/en/latest/translation/#translating-data-disaggregations-and-columns).

The problem is the lower-case part. A common use-case is for a country to copy/paste _exactly_ what is in their data (CSV files) into the data.yml as the translation keys. But the lower-case conversion prevents this from working, unless the values in the CSV files happen to be lower-case already.

Let's remove the lower-case aspect of this. This "feature" is not being used by any countries, and turns out to be contrary to the way countries actually want to use this platform.

## Tests added

There are existing tests that will confirm this still works.

## Documentation added

The part of the documentation mentioning the lower-case part is being removed as part of this PR.

### DIFF
--- a/_includes/multilingual-js-base.html
+++ b/_includes/multilingual-js-base.html
@@ -10,12 +10,12 @@ var translations = {
 
     // The majority of uses of this function are to translate disaggregation
     // data. To spare data providers of needing to enter "data." in front of
-    // their disaggregation data, we do it for them here.
-    var originalKey = key;
-    if (!key.includes('.') && this.data) {
-      key = 'data.' + key.toLowerCase();
+    // their disaggregation data, we specifically look for that here.
+    if (typeof this.data === 'object' && this.data[key]) {
+      return this.data[key];
     }
 
+    var originalKey = key;
     var drilled = this;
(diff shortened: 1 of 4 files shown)


LucyGwilliamAdmin:
@brockfanning I'm not sure how I can test this as we don't have translations on the UK site.

brockfanning:
@LucyGwilliamAdmin That's a good point. If it's helpful, we are currently using this on kazstat.github.io/sdg-site-kazstat. See  https://github.com/kazstat/sdg-site-kazstat/blob/43deeec146f5074cbf92678d34a371f1bf76adc2/_config.yml#L126

PR #206, opened 2019-06-19
Expect dash-delimited translation keys for indicators/targets

brockfanning:
## Breaking changes

None (backwards-compatibility is included)

However note that after adopting the translation updates in [this PR](https://github.com/open-sdg/sdg-translations/pull/93) a country would need to update their version of jekyll-open-sdg-plugins to 0.0.12, [like so](https://github.com/open-sdg/open-sdg-site-starter/blob/develop/Gemfile#L8).

## Description

This is to support a potential change in SDG Translations, from dot-delimited translation keys to dash-delimited translation keys. See [here](https://github.com/open-sdg/sdg-translations/issues/92) for details.

The main use-case for this change is to allow countries to add translation keys for global indicators in their metadata. For example, they can put the following in an indicator's metadata file:

```
graph_title: global_indicators.1-1-1.title
```

Currently, with the dot-delimited keys, the above is impossible, because it would have to be:

```
graph_title: global_indicators.1.1.1.title
```

All of the dots break the parsing of the translation key.

## Tests added

I believe no new tests are needed.

## Documentation added

This is more of a bugfix, so existing documentation is fine.

### DIFF
--- a/_includes/indicator-variables.html
+++ b/_includes/indicator-variables.html
@@ -30,4 +30,4 @@
 @deprecated: The following variable will be removed for 1.0.0.
 {% endcomment %}
 {% assign indicator_id_dashes = meta.indicator | replace: ".", "-" %}
-{% assign translated_indicator = t.global_indicators[indicator_id_dashes] %}
+{% assign translated_indicator = t.global_indicators[indicator_id_dashes] | default: t.global_indicators[indicator_id] %}
--- a/_includes/indicator-variables.html
+++ b/_includes/indicator-variables.html
@@ -29,4 +29,5 @@
 {% comment %}
 @deprecated: The following variable will be removed for 1.0.0.
 {% endcomment %}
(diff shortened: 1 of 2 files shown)


brockfanning:
@LucyGwilliamAdmin @Norric1Admin @philipashlock Would appreciate your thoughts on this PR, and on the [corresponding PR](https://github.com/open-sdg/sdg-translations/pull/93).

Norric1Admin:
@brockfanning Will the switch to dashes in https://github.com/open-sdg/sdg-translations/pull/93 require any local changes, or does the assign/replace function in this PR do that for us?

brockfanning:
@Norric1Admin This PR itself has no breaking changes, but there is one update that would need to be made after adopting the updated translations: updating jekyll-open-sdg-plugins to 0.0.12. I've added notes in this PR under "Breaking changes".

brockfanning:
Thank you @Norric1Admin. If there are no objections, I will also merge the PR in sdg-translations.

PR #211, opened 2019-06-27
Catch another spot where zeroes are getting clobbered: the headline

brockfanning:
We recently added code to allow zeroes in data to survive into the charts/tables, but we forgot this one spot: the headline data.


brockfanning:
@LucyGwilliamAdmin Thanks for testing. These does seem to be a javascript error happening on that page. It's not technically related to this change, but I added a commit that should resolve it. Do you think you could give this another test?

PR #212, opened 2019-06-27
Remove data rounding and provide an easier way to add custom rounding

### LINKED ISSUE #199: Data rounding changes non-decimal numbers
The [data rounding function](https://github.com/open-sdg/open-sdg/blob/master/_includes/assets/js/indicatorModel.js#L19) actually changes non-decimal numbers. For example, it changes 171615 to 172000. I don't believe this is desired behavior. Or at least, if it is, it should be easier to override. (Right now, the only way to override the function is to override the entire indicatorModel.js file, or the scripts.html file.)

So I think we should either change this function as needed, or make it easier to override.

brockfanning:
This change removes the data rounding function, which has been causing issues recently. It also adds an easier way for countries to add custom rounding, without needing to override any files. Finally it adds a way to add custom javascript, through the site config, similar to the `custom_css` config option.

Fixes #199

### DIFF
--- a/_includes/assets/js/indicatorModel.js
+++ b/_includes/assets/js/indicatorModel.js
@@ -15,12 +15,6 @@ var indicatorModel = function (options) {
   this.onSelectionUpdate = new event(this);
   this.onNoHeadlineData = new event(this);
 
-  // data rounding:
-  this.roundingFunc = options.roundingFunc || function(value) {
-    var to = 3, mult = Math.pow(10, to - Math.floor(Math.log(Math.abs(value)) / Math.LN10) - 1);
-    return Math.round(value * mult) / mult;
-  };
-
   // json conversion:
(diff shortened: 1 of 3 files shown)


brockfanning:
Here is an example of how to do a custom rounding function, with this approach:
* [Rounding function in a javascript file](https://github.com/kazstat/sdg-site-kazstat/blob/develop/assets/js/custom.js)
* [Config setting to load the javascript file](https://github.com/kazstat/sdg-site-kazstat/blob/develop/_config.yml#L132)

brockfanning:
@LucyGwilliamAdmin I think the problem is that the [indicatorModel.js](https://github.com/ONSdigital/sdg-indicators/blob/opensdg-rounding/_includes/assets/js/indicatorModel.js) file is still overridden.

PR #233, opened 2019-07-04
Hide/show the sidebar depending on the active tab

brockfanning:


Norric1Admin:
Agreed, I like this feature a lot! I report the same problem on my Samsung phone.

brockfanning:
I couldn't figure out how to get the animation to work vertically, but the latest commit at least fixes the white space problem.

LucyGwilliamAdmin:
Maybe using something like [this](https://css-tricks.com/almanac/properties/s/scroll-behavior/)

brockfanning:
No prob. My first impression was that this requires some javascript. I've pushed up another commit that should address the problem.

LucyGwilliamAdmin:
One thing I have just noticed is that when on desktop version and map tab has been selected there is a lot of white space between the metadata tab panel and the bottom of the page

brockfanning:
I haven't noticed that space problem. Is this happening on a test branch that I could take a look at, by any chance?

LucyGwilliamAdmin:
This feature branch https://sdg.mango-solutions.com/remove-disaggs/3-a-1/

brockfanning:
Good catch, that was due to the indicator having a lot of disaggregations (making it very tall). I've made another commit that should help. It might be a good idea to confirm that the mobile fix still works.

LucyGwilliamAdmin:
Something else has happened now: 
Seems like the disaggregations are displaying below the bottom of the page:

brockfanning:
I just tried a quick fix, any better now?

LucyGwilliamAdmin:
@brockfanning Great, works on mobile and desktop!

Issue #221, opened 2019-07-01
Link to stats office on platform


LucyGwilliamAdmin:
Suggestion from Kyrgyzstan:
Icon on top-left of homepage which is also link to country's stats office website
![image](https://user-images.githubusercontent.com/43404358/60651053-4b001100-9e3d-11e9-8faa-1a770982fb90.png)

brockfanning:
Was the screenshot above intended to have a mockup of the feature? I'm not seeing it if so. I think that a mockup would help us decide whether this is something that should go into Open SDG, or should be custom per country.

LucyGwilliamAdmin:
@brockfanning apologies, I've updated the image

brockfanning:
Just speaking from a technical standpoint, it might be a little problematic to have the logo in the "gutter" on the left. It would definitely disappear on mobile.

Personally I feel like the most appropriate place for a link to the stats office is in the footer. Typically the footer is the place to link to related but separate stakeholders, contributors, etc. And that would also solve the mobile problem.

LucyGwilliamAdmin:
Let's close this as 1. Brock pointed out why this wouldn't be suitable, and 2. Kyrgyzstan have gone ahead and done it themselves (https://sustainabledevelopment-kyrgyzstan.github.io/)

PR #235, opened 2019-07-05
Add NSO link as image to header

LucyGwilliamAdmin:
Adds feature explained in #221  
Created test branch on UK site: https://sdg.mango-solutions.com/nso-office-test/


brockfanning:
@LucyGwilliamAdmin My only concern with this is that usually web users expect the top-left logo to be a link to the home page. I think some users may click on it expecting to go to the homepage, and then get taken off-site.

To throw out one possible alternative - do you think it would be sufficient if the logo were in the footer?

Or another alternative - could the logo link to the homepage, instead of the offsite URL? (though kind of defeats the purpose I guess)

LucyGwilliamAdmin:
Something needs to be done for high contrast - changing to high contrast currently breaks the image

PR #239, opened 2019-07-10
Refactor/cleanup of footerFields for charts/tables

### LINKED ISSUE #238: Footer fields of data table are not translated
The source, units of measurement, etc, in the footer of the data table are not getting translated.

brockfanning:
This makes the treatment of the footer fields more consistent between the table and chart, by having both use the same object. It also takes care of a TODO.

Fixes #238

### DIFF
--- a/_includes/assets/js/indicatorModel.js
+++ b/_includes/assets/js/indicatorModel.js
@@ -177,12 +177,13 @@ var indicatorModel = function (options) {
       spanGaps: false
     };
 
-    that.footerFields = {
-      'Source': that.dataSource,
-      'Geographical Area': that.geographicalArea,
-      'Unit of Measurement': that.measurementUnit,
-      'Footnote': that.footnote,
-    };
+    that.footerFields = {};
+    that.footerFields[translations.indicator.source] = that.dataSource;
+    that.footerFields[translations.indicator.geographical_area] = that.geographicalArea;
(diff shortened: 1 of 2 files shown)


brockfanning:
@LucyGwilliamAdmin @Norric1Admin @SavvasStephanides Would anyone be available to test this PR?

brockfanning:
Thanks @LucyGwilliamAdmin. I've got a feeling this changes a lot of the same code as #243 which may make this obsolete. So let's hold off on merging this.

LucyGwilliamAdmin:
Is this/should this also be fixed with this PR?
y-axis is not translated
https://sustainabledevelopment-germany.github.io/1-2-2/
image

brockfanning:
Hmm, I thought that was fixed in #228. This branch may have lacked that commit -- I've just rebased to see if that fixes it.

But again I'm guessing this will conflict with #243 so we might want to focus on #243 instead of merging this. I'll try to take a look soon.

LucyGwilliamAdmin:
Yeah, I was actually thinking something similar - I'm not sure how much further I can go with #243 as I haven't really worked with JS/JQuery before

brockfanning:
@LucyGwilliamAdmin Maybe we merge this as a short-term fix and then continue with #243 as we're able?

LucyGwilliamAdmin:
@brockfanning yes, that sounds like a good idea

Issue #223, opened 2019-07-01
Social media buttons on every page


LucyGwilliamAdmin:
Suggestion from Kyrgyzstan:
Buttons on every page when when clicked would allow you to share the link to that page via Twitter or Facebook

brockfanning:
Fixed by #242

PR #242, opened 2019-07-11
Add sticky share buttons

LucyGwilliamAdmin:
Feature for #223 

Can be see on this [test branch](https://sdg.mango-solutions.com/sticky-share/)


brockfanning:
Nice that it's so easy! Just a couple of suggestions:
1. It should probably be optional, only showing up if enabled in the site config
2. That `property` code seems like it would be specific to an account - should that be configurable?

LucyGwilliamAdmin:
@brockfanning was thinking something similar.

1. Definitely think having it should be configurable.
2. Depends if people wanted to customise it e.g. wanted to be able to share pages to different platforms other than Twitter and Facebook

LucyGwilliamAdmin:
@brockfanning I've made this optional now so if a country wants the buttons they need to add `stickyshare: true` to the config file.

I will put something together for the readthedocs about the feature - do you think that the property should be configurable or should it just be left with the share options as Facebook, Twitter and Email?

LucyGwilliamAdmin:
Have started the doc page [here]. lf you think countries should have their own `property`, I will add how to create this.

The image doesn't seem to be showing in the .md file even though I've used the same code used in other doc pages.

Even though I've created the page, I'm not actually sure how to add it to readthedocs or how to preview it on there.

LucyGwilliamAdmin:
Icons don't show when high contrast mode is activated

brockfanning:
@LucyGwilliamAdmin There's actually Open SDG code that intentionally changes the `src` of all images on the page, to add the "high-contrast" to the path. This is probably breaking the ShareThis images, since they don't exist. Here's that code: https://github.com/open-sdg/open-sdg/blob/master/_includes/assets/js/accessibility.js#L75

We might have to exempt the ShareThis images, somehow. I'm not sure how to do high-contrast for ShareThis, though.

LucyGwilliamAdmin:
@brockfanning thanks! I have added code for this now

brockfanning:
@LucyGwilliamAdmin It's looking good. A few more comments:
* My feeling is that the "property" is probably tied to the [ShareThis analytics](https://sharethis.com/support/faq/social-analytics/). I'm not sure about that, but that's what I would expect. What do you think about making the property into a configurable item, but giving it a default value (in open-sdg-site-starter)?
* I think since we're specifically using ShareThis, the site config name should include "sharethis", like maybe "sharethis_property"? The logic could check for the present of that variable. (ie, `if site.sharethis_property`)
* In order to keep the size of this repository manageable, I had adopted a general rule of not using screenshots in the docs. We could re-think this, but I think in this case a simple text description, and maybe a link to sharethis.com would be sufficient.

LucyGwilliamAdmin:
@brockfanning I've made those changes and made this [PR](https://github.com/open-sdg/open-sdg-site-starter/pull/38) in open-sdg-site-starter.

PR #243, opened 2019-07-14
Change chart attributes to HTML

### LINKED ISSUE #73: Download chart as image
This was an issue that was nearly resolved on the UK site with PR ONSdigital/sdg-indicators#2591 . As I'm closing the PR there I wanted to keep a link open to the code - even if we then close this issue.

The feature was for a user to be able to click a button and download an image of the chart they had been producing.

The reason the UK PR never got merged was that the legend was not on the chart because it had been generated outside of the chartjs. The code for downloading the rest of the image was completed however in that PR.

LucyGwilliamAdmin:
Fixes #73 

Can be see on this [test branch](https://sdg.mango-solutions.com/html-chart-attibutes/)

Not completely done - all chart attributes have been changed into HTML and a button has been created to allow the chart to be downloaded with all of the html attributes.

However still need to get the function for turning the canvas into an image working so that `URL.createObjectURL(new Blob(..))` can be used as the href for the element. 

The way it was being done [here](https://github.com/open-sdg/open-sdg/blob/49a60c233553237ec1c2b5209eaea18499746465/_includes/assets/js/indicatorView.js): 
`$(function() { $("#btn-save").click(function() {html2canvas($("#chart"), {onrendered: function(canvas) {Canvas2Image.saveAsPNG(canvas);}});});});`

uses a click function but was trying to get it so that it's in the same format as this.createDownloadButton and this.createSourceButton, which is proving quite difficult (maybe because I'm not quite sure what I'm doing)

I should also note that html2canvas and Canvas2Image aren't currently in scripts.html as I wasn't sure whether we would use those in this new function. 

Another problem you may notice is that the gap between the chart and the legend is too big - wasn't sure how to change the size of this.

### DIFF
--- a/_includes/assets/js/indicatorView.js
+++ b/_includes/assets/js/indicatorView.js
@@ -317,6 +317,8 @@ var indicatorView = function (model, options) {
 
     $(this._legendElement).html(view_obj._chartInstance.generateLegend());
   };
+  
+  
 
   this.createPlot = function (chartInfo) {
 
@@ -375,11 +377,7 @@ var indicatorView = function (model, options) {
           display: false
         },
         title: {
-          fontSize: 18,
-          fontStyle: 'normal',
-          display: this._model.chartTitle,
-          text: this._model.chartTitle,
-          padding: 20
+          display: false
         },
         plugins: {
           scaler: {}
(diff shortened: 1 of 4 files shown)


LucyGwilliamAdmin:
@brockfanning close but only working on Chrome still. No idea why/what's causing this - maybe something in the js (https://github.com/niklasvh/html2canvas/releases/download/0.4.1/html2canvas.js or https://hongru.github.io/proj/canvas2image/canvas2image.js). Could you have a look when you have a chance please as you might see what's causing it?

brockfanning:
html2canvas seems like a really good choice here. I'll take a closer look this weekend.

brockfanning:
@LucyGwilliamAdmin Actually I was having some trouble with html2canvas: the latest version only produced blank images. It seems like a really well-supported library with a large community, so I'm probably doing something wrong. However I also found [domtoimage](https://github.com/tsayen/dom-to-image) which worked great for me. I added some [more commits](https://github.com/brockfanning/open-sdg/commits/chart-html-attributes) to my fork so you can see what I'm thinking. I've tested this in Firefox but nothing else yet.

LucyGwilliamAdmin:
@brockfanning this looks great now! Works on Chrome but not on IE or Edge (download doesn't work at all) or on mobile (graph doesn't display on image)

Get this error on Edge (not sure if same error on IE as for some reason inspect element won't work for me):
image

brockfanning:
@LucyGwilliamAdmin I haven't tested, but just as a shot-in-the-dark I've added another commit to my fork to remove Bluebird. As I understand it, Bluebird is a polyfill for Promises which we should not need with this update. I could be wrong though, so appreciate any testing.

LucyGwilliamAdmin:
still seem to be getting the error but doesn't mention Bluebird:
image

brockfanning:
Ok, just pushed another commit to log the error to the console. If you could take a look at the console (in Edge at least) it should hopefully give some clue about why the download is failing.

brockfanning:
Actually it looks like [IE is not supported in domtoimage](https://github.com/tsayen/dom-to-image#browsers). And Edge seems to have the same problem. So I guess that's not a feasible approach. I'll take another look at html2canvas.

brockfanning:
@LucyGwilliamAdmin I've given html2canvas another try on [my fork](https://github.com/brockfanning/open-sdg/tree/chart-html-attributes). Strangely it seems that the latest version of html2canvas (1.0.0-rc-3) is a lot more tricky to get working than the older version (0.4.1) you were using. But after some fiddling with the [configuration options](http://html2canvas.hertzen.com/configuration) it seems to be working reasonably well. I'm curious though about IE and Edge support. Are you able to try it out?

Note that the branch is now pretty out of sync from master, so it probably will need to be re-created in a fresh branch (if we decide it's worth merging).

LucyGwilliamAdmin:
Still seem to be getting an error, although doesn't look like it's related to html2canvas:
image

LucyGwilliamAdmin:
Looking at this IE may require a vendor prefix or different name for use:
image

Would it be possible to use similar code to what is used for the Download Chart CSV button to get the blob:
i.e. .attr({
        'href': URL.createObjectURL(new Blob([this.toCsv(table)], {
          type: 'text/csv'
        }))

Although I have recently opened #313 so not sure if that's caused by that line so that might not work on IE either

LucyGwilliamAdmin:
@brockfanning I've added a polyfill [here](https://github.com/LucyGwilliamAdmin/open-sdg/blob/patch-1/_includes/scripts.html#L20) and seems to be working on IE and Edge now.

brockfanning:
@LucyGwilliamAdmin I added "toBlob" to our polyfill.io thing [here](https://github.com/brockfanning/open-sdg/tree/chart-html-attributes), I don't suppose this solves it?

brockfanning:
Awesome, glad that fixes it. Maybe also try my latest commit, which adds it to our polyfills include file?

LucyGwilliamAdmin:
@brockfanning just tested and don't get an error but the chart also doesn't download. Is [this](https://github.com/LucyGwilliamAdmin/open-sdg/blob/patch-1/_includes/scripts.html#L20) ok? Or should we not use GitHub as source?

brockfanning:
Ok, sounds like polyfill.io isn't as good as what you found. In general using a CDN is a little better, if one is available. How about this? https://cdnjs.cloudflare.com/ajax/libs/javascript-canvas-to-blob/3.15.0/js/canvas-to-blob.min.js

Also to keep things consistent, could it go in the _includes/polyfills.html file?

brockfanning:
I added a commit on my fork if that helps.

LucyGwilliamAdmin:
Ok great, that worked - just adding the changes to a new branch and will create new PR

PR #247, opened 2019-07-16
Update circle-ci.md

LucyGwilliamAdmin:


brockfanning:
Sorry to drag my feet here - I still wanted to look at this in more detail and give it a try.

brockfanning:
@LucyGwilliamAdmin Are there some missing steps here? I'm looking at the docs and it seems incomplete.

A related note - I know that this is easier to do if using Gitlab. And as for Github, I think that #398 probably makes it easier as well. In fact, we might want to postpone anything to do with CircleCI until #398 is done, as it may make CircleCI unnecessary.

LucyGwilliamAdmin:
@brockfanning I don't think anything is missing. There is a corresponding PR [here](https://github.com/open-sdg/open-sdg-data-starter/pull/16/files) where the changes take place in the data repo. Although I guess we may need some instructions for those who already have sites.

Agree we should wait until #398 is done though.

brockfanning:
I'll close this for now, since we're standardizing around Github Actions. This is still an important feature but should be revisited a bit in the future, once we've got more experience using Github Actions.

PR #255, opened 2019-07-20
Translate map regions

### LINKED ISSUE #253: Translate region labels in maps
Right now, region labels on the maps are not getting translated.

brockfanning:
Fixes #253


brockfanning:
@LucyGwilliamAdmin Thanks for testing. I just got the same results, but I think it was maybe a problem with 1-1-1? Could you give it another try? I've updated the data repository to turn off the map on some indicators where data is not available.

PR #257, opened 2019-07-20
Set a body class for each layout

brockfanning:
This will allow easier CSS selecting for particular layouts.


brockfanning:
As an example of how this might be used, see [here](https://kazstat.github.io/sdg-site-kazstat/). The corner images, and the total white background, is only in effect on the frontpage layout, by selecting for [the "layout-frontpage" body class](https://github.com/kazstat/sdg-site-kazstat/blob/develop/assets/css/custom.scss#L14).

PR #258, opened 2019-07-21
Require jekyll-open-sdg-plugins 0.0.13, and not jekyll-get-json

brockfanning:
This is a preemptive PR for the next minor release, 0.8.0. With jekyll-open-sdg-plugins version 0.0.13, the jekyll-get-json plugin is no longer required. So this updates the gemspec file accordingly.

Breaking change: After this change, all sites will need to have jekyll-open-sdg-plugins 0.0.13 or later. So this should also be accompanied by a corresponding change in open-sdg-site-starter.


LucyGwilliamAdmin:
Have just made PR in open-sdg-site-starter to upgrade jekyll-open-sdg-plugins to 0.0.13

Issue #152, opened 2019-04-12
Search symbol not showing in IE


LucyGwilliamAdmin:
![image](https://user-images.githubusercontent.com/43404358/56033209-eb9dd380-5d1b-11e9-90c2-5a2f4082b908.png)

Norric1Admin:
It would also be good to get the 'Goals' option back in the menu bar (see [UK site](https://sustainabledevelopment-uk.github.io/)) for accessibility reasons.

brockfanning:
@Norric1Admin @LucyGwilliamAdmin I'm not able to recreate the problem with 'Goals' not showing. It appears to be there for both the staging and production UK sites, as far as I can tell. Though it's clearly missing in the screenshot above. Maybe you already fixed it?

I can, however, recreate the issue with the search symbol being missing in IE. Unless you object I'd propose we rename this issue to just be about the search symbol.

LucyGwilliamAdmin:
@brockfanning the screenshot above is from the Kyrgyzstan platform - we just think it might be better for the starter sites to have the 'Goals' page.

Happy to have two separate issues

brockfanning:
@LucyGwilliamAdmin Gotcha. Adding it to the starter sites sounds like a good idea to me, and that change can be made as a pull-request in [this repository](https://github.com/open-sdg/open-sdg-site-starter). I've created an [issue](https://github.com/open-sdg/open-sdg-site-starter/issues/29) for it over there. I'll go ahead and rename this issue to focus on the IE problem.

brockfanning:
This also affects the global icon by the language switcher. I get the impression that FontAwesome doesn't really support IE anymore: https://github.com/FortAwesome/Font-Awesome/issues/8825

LucyGwilliamAdmin:
@brockfanning on that issue they managed to fix the problem:

> I tried the solutions here ... and now I found out why it doesn't show because
you didn't include fontawesome-webfont.eot in your /fonts which it was recommended font for IE11

LucyGwilliamAdmin:
^ that might not be the solution, might just be that we need a newer version of font-awesome.

That being said, I've just been on IE11 on my personal laptop and the search icon was displaying on that so will check on my work laptop when I can to see if it's also fixed on that

LucyGwilliamAdmin:
[...]

PR #263, opened 2019-07-25
Searchbtn fix

### LINKED ISSUE #259: Search bar isn't right on IE
image

LucyGwilliamAdmin:
Fixes #259 - the sizing at least.

Still have issue with font awesome icon (or at least I am @brockfanning can you confirm whether you can see the search icon on IE)

https://sdg.mango-solutions.com/search-button/

### DIFF
--- a/_includes/search.html
+++ b/_includes/search.html
@@ -1,6 +1,6 @@
 <div id="search" class="menu-target">
   <input id="indicator_search" title="{{ t.search.indicator_search }}" placeholder="{{ t.search.indicator_search }}" data-url="{{ site.baseurl }}{{ baseurl_folder }}/indicators.json" data-pageurl="{{ site.baseurl }}{{ baseurl_folder }}/search/?q=" />
   <label for="indicator_search">
-    <button id="search-btn" type="submit"><i class="fa fa-search" aria-hidden="true"></i></button><span>{{ t.search.search }}:</span>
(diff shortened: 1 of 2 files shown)


LucyGwilliamAdmin:
From #152:

Works on personal laptop bur not on work laptop. I have come across this post which says:

'Font downloads are disabled in IE, under Internet Options / Security / Custom Level / Font Downloads enable /disable. They might be disabled by your network admin, in which case you would not be able to see nor change this setting.'

Unfortunately, these security settings are disabled for us so I cannot try this.

LucyGwilliamAdmin:
@brockfanning could you take a look at this please?

Also once merged - is there a way for it be added to the 0.7.0 release as it's a bug fix rather than a new feature, or would a new release (0.7.1) have to be made?

LucyGwilliamAdmin:
@brockfanning - thanks for testing.

When I remove the `style="height: 34px;"` attribute from the button, it looks like this on IE:
image
compared to having the style attribute:
image


Agree with you on having `right: 20px` so have changed that.

LucyGwilliamAdmin:
@brockfanning do you think this is ready to be merged in?

brockfanning:
My mistake, I see that the `height: 34px` that's in the CSS now is only applied to the `input` element. Could we add CSS to apply that height to the `button` element? Like maybe:
```
#search-btn {
  height: 34px;
}
```
It's usually a good idea to avoid putting style directly in the HTML, because it can confusing things for developers who are trying to tweak things later.

PR #264, opened 2019-07-25
Force a specific unit if needed when there is no headline

### LINKED ISSUE #262: Still slight issue with default selections
A fix went through so that if no headline data is available, something would still show. However, looking [here](https://sdg.mango-solutions.com/0-7-0/4-1-1/), no data is showing when indicator is first clicked on. Data is available for disaggregation but not under the default units type.

brockfanning:
This should fix #262 

This updates the automatic "minimum field" selection logic, for when an indicator has no headline. It should also look for a unit, and "force" this unit if necessary.

### DIFF
--- a/_includes/assets/js/indicatorModel.js
+++ b/_includes/assets/js/indicatorModel.js
@@ -582,7 +582,8 @@ var indicatorModel = function (options) {
     if((options.initial || options.unitsChangeSeries) && !this.hasHeadline) {
       // if there is no initial data, select some:
 
-      var minimumFieldSelections = {};
+      var minimumFieldSelections = {},
+          forceUnit = false;
       // First, do we have some already pre-configured from data_start_values?
       if (this.startValues) {
         // We need to confirm that these values are valid, and pair them up
@@ -607,6 +608,8 @@ var indicatorModel = function (options) {
         // value in each drop-down, up until there are enough selected to display
         // data on the graph. First we get the number of fields:
         var fieldNames = _.pluck(this.fieldItemStates, 'field');
+        // Manually add "Units" so that we can check for required units.
+        fieldNames.push('Units');
         // We filter our full dataset to only those fields.
         var fieldData = _.map(this.data, function(item) { return _.pick(item, fieldNames); });
         // We then sort the data by each field. We go in reverse order so that the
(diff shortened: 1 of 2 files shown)


LucyGwilliamAdmin:
Something to do with [this](https://github.com/open-sdg/open-sdg/blob/b6ecb4612865515e2b2731db65652ce0ceb3ce9a/_includes/components/fields-template.html):
By default, no 'style="width:XX;"' attribute exists

LucyGwilliamAdmin:
I guess this code needs to be added somewhere:

document.getElementsByClassName("selected")[DEFAULTSELECTIONS].setAttribute("style", "width:XX;")

LucyGwilliamAdmin:
@brockfanning although looks as if this isn't a problem for the original 0.6.0 PR i.e. when a units didn't have to be selected first e.g. [10.6.1](https://sdg.mango-solutions.com/0-7-0/10-6-1/)

brockfanning:
I haven't tested, but just pushed a new commit as a shot in the dark, in case it's as simple as needing to "click" the Unit before clicking the fields.

LucyGwilliamAdmin:
This hasn't seemed to fix it for some reason:
https://sdg.mango-solutions.com/default-selections/4-1-1/

LucyGwilliamAdmin:
Some more info:
If I click another unit it shows that data is selected for 'Subject'

brockfanning:
Ok cool, thanks for the testing. I will dig into this some more.

LucyGwilliamAdmin:
@brockfanning style attribute is actually there but for some reason the width is set to 0:
image

LucyGwilliamAdmin:
@brockfanning do we want to merge this in as it is for now for 0.8.0? It's a step in the right direction compared to what it is doing at the moment i.e. no data showing on the graph?

brockfanning:
I leave that up to you. I unfortunately have to cut back in hours, and so have not come back to this. We'd want to keep the issue open though so we don't forget about the bug.

LucyGwilliamAdmin:
For some reason, this isn't working at all now:
https://sdg.mango-solutions.com/0-8-0/4-1-1/

LucyGwilliamAdmin:
Looks like even though Unit type is being selected automatically, a disaggregation isn't

LucyGwilliamAdmin:
@brockfanning should I remove this fix from the release and changelog?

brockfanning:
No I think it makes sense to leave it in the changelog, for historical purposes. I've re-opened the issue and will try to take a look. Hopefully it's a quick fix.

PR #266, opened 2019-07-27
Use page.url instead of page.permalink to determine active menu item

brockfanning:
Page.url is more reliable, since it is a Jekyll-generated variable that doesn't depend on user configuration.
This also increases consistency in the logic by removing slashes before comparing the current path to the menu item path. This avoid confusing "misses" when a trailing slash is used or not used in the menu config.


brockfanning:
To see the problem, see [here](https://sdg-kazstat.github.io/reporting-status/). By contrast see [here](https://kazstat.github.io/sdg-site-kazstat/reporting-status/) for the fix in action.

PR #277, opened 2019-08-08
Merge latest changes into DAC (no significant overwrites)

Norric1Admin:
The only file being overwritten is search.html, and the edit is identical code moved from line 5 to 2.


LucyGwilliamAdmin:
@Norric1Admin the code was identical but the change of order was important as it affected order of tabbing - the button was coming before the input

PR #287, opened 2019-08-09
Accessibility updates

LucyGwilliamAdmin:
UK site has recently been tested at digital accessibility center. The following fixes were suggested for the site:
* Screen reader should read out if a disaggregation drop-down is expanded or collapsed
* Make obvious what the social media handles in the footer are
* Give embedded features a title
* Add a label to the search bar for screen readers
* Change the tab title when on a goal page so that it's obvious to screen reader users what page they are on
* Split up footer links so that they're read out separately by screen reader
* Fix menu drop down on mobile devices
* Screen reader to read out whether checkbox is checked/unchecked


_includes/footer.html
```diff
@@ -6,10 +6,10 @@
           <ul>
             <li><a href="mailto:{{site.email_contacts.suggestions}}">{{ t.menu.contact_us }}</a></li>
             {% if site.twitter %}
-            <li><a class="twitter-follow-button" href="https://twitter.com/{{ site.twitter }}" data-show-count="false">@{{ site.twitter }}</a></li>
+            <li><a class="twitter-follow-button" href="https://twitter.com/{{ site.twitter }}" data-show-count="false">Twitter: @{{ site.twitter }}</a></li>
```
brockfanning: Do you think "Twitter" needs to be translated in any languages? If not, this is fine. But I suspect it might need to be translated, even if only in Chinese, Arabic, etc. If we want to allow this to be translated, it would need to be a reference to a "translation key", like maybe `{{ t.general.twitter }}`? (Which of course would need to be added to sdg-translations)

_includes/footer.html
```diff
@@ -6,10 +6,10 @@
...
             {% endif %}
             {% if site.facebook %}
-            <li><a href="https://facebook.com/{{ site.facebook }}" target="_blank">@{{ site.facebook }}</a></li>
+            <li><a href="https://facebook.com/{{ site.facebook }}" target="_blank">Facebook: @{{ site.facebook }}</a></li>
```
brockfanning: Same idea here - does "Facebook" need to be translated?

_includes/search.html
```diff
@@ -1,6 +1,6 @@
 <div id="search" class="menu-target">
   <input id="indicator_search" title="{{ t.search.indicator_search }}" placeholder="{{ t.search.indicator_search }}" data-url="{{ site.baseurl }}{{ baseurl_folder }}/indicators.json" data-pageurl="{{ site.baseurl }}{{ baseurl_folder }}/search/?q=" />
   <label for="indicator_search">
-    <button id="search-btn" type="submit"><i class="fa fa-search" aria-hidden="true"></i></button><span>{{ t.search.search }}:</span>
+    <button id="search-btn" type="submit"><i class="fa fa-search" aria-hidden="false" aria-label="Indicator search"></i></button><span>{{ t.search.search }}:</span>
```
brockfanning: Since it's an aria-label I suspect it needs to be translated? Luckily we have this as a translation key already: 
`{{ t.search.indicator_search }}`

_layouts/indicator.html
```diff
@@ -84,9 +84,10 @@ <h4>{{ t.indicator.sub_categories }}</h4>
             <li role="presentation" class="nav-item map" style="display:none">
               <a data-no-disagg="true" class="nav-link" data-toggle="tab" href="#mapview" aria-controls="mapview" role="tab" {% include autotrack.html preset="tab_data_map" category="Tab change" action="Change data view" label="Change to Map tab" %}>{{ t.indicator.map }}</a>
             </li>
-            {%- if meta.embedded_map_html -%}
+            {%- if meta.embedded_feature_url -%}
```
brockfanning: I like the change. Just noting that we should call this out as a breaking change in the release notes.

_layouts/indicator.html
```diff
@@ -84,9 +84,10 @@ <h4>{{ t.indicator.sub_categories }}</h4>
...
             <li role="presentation" class="nav-item embedded-map">
-              <a data-no-disagg="true" class="nav-link" data-toggle="tab" href="#embeddedmapview" aria-controls="embeddedmapview" role="tab" {% include autotrack.html preset="tab_data_embed" category="Tab change" action="Change data view" label="Change to embedded item tab" %}>{{ t.indicator.map }}</a>
+              <a class="nav-link" data-toggle="tab" href="#embeddedmapview" aria-controls="embeddedmapview" role="tab" {% include autotrack.html preset="tab_data_embed" category="Tab change" action="Change data view" label="Change to embedded item tab" %}>{{ meta.embedded_feature_tab_title }}</a>
```
brockfanning: The `data-no-disagg="true" ` needs to be added back here.

_layouts/indicator.html
```diff
@@ -104,9 +105,19 @@ <h4>{{ t.indicator.sub_categories }}</h4>
...
               </div>
             </div>
-            {% if meta.embedded_map_html %}
+            {% if meta.embedded_feature_url %}
             <div role="tabpanel" class="tab-pane" id="embeddedmapview" class="embedded-map">
-              {{ meta.embedded_map_html }}
+              <div role="tabpanel" class="tab-pane" id="embeddedmapview" >
+              <div class="row">
+              <div class="col-xs-12">
+                <h3 tabindex="0">{{ meta.embedded_feature_title }}</h3>
```
brockfanning: Running this through the `t` filter will allow this to support the possibility of translation: `{{ meta.embedded_feature_title | t }}`

_includes/footer.html
LucyGwilliamAdmin: @brockfanning I'm checking translations now - if a translation isn't available for a language should I just add it to sdg-translations as English for those languages that aren't available?

LucyGwilliamAdmin:
I have tested that these are all working as expected on https://sdg.mango-solutions.com/open-sdg-dac/ and everything seems fine.

@brockfanning can you do final review please?

LucyGwilliamAdmin:
@brockfanning do you think this is ready to be merged in?

brockfanning:
Sorry for the delay. I haven't made it through all the changes. Also I've been meaning to test with a multilingual site, because I worry that the wider contrast button may crowd the language toggle on mobile.

LucyGwilliamAdmin:
No problem. Just tested on Kazakh site and text doesn't fit in box on desktop or mobile for contrast button.
Desktop:
image
Mobile:
image

LucyGwilliamAdmin:
@brockfanning maybe one way to overcome this would be to have the button text to change between 'High contrast' and 'Default contrast' instead of 'Enable high contrast' and 'Disable high contrast

LucyGwilliamAdmin:
@brockfanning I've made some changes to the contrast button now:
### UK site (en)
Desktop:
image
Mobile:
image

### Armenia site (ar)
Desktop:
image
Mobile:
image

### Armenia site (ru)
Desktop:
image
Mobile:
image

### Kazakhstan site (kz)
Desktop:
image
Mobile:
image

Do you think it works like this?

LucyGwilliamAdmin:
One problem I'm struggling with is awkward sized values. For example, for UK site, default contrast looks like this at it's very smallest (279px):
image
But the very smallest the window will go without moving the button in high contrast mode is 425px:
image
Is there a way I can force the words to be on separate lines?

LucyGwilliamAdmin:
I've managed to fix this now so mobile looks like this now:
image

brockfanning:
Sorry to be a roadblock, but I worry that some countries may not like the more verbose high-contrast button, though I do see the benefit for those with screen readers. I know a lot of work has already been done, but just to be sure - was there no way to satisfy the requirement using "aria" attributes instead of changing the button text? I feel like, if the high-contrast change is necessary, that it should be optional.

Also just a tip, since I know a lot of improvements are in this PR - in general it is most efficient for each pull request to be as small and targeted change as possible. Bundling a lot of changes into one PR has some significant downsides:
1. More time-consuming to test
2. Which means it will take longer to get approved
3. Which means it is more likely to end up with conflicts when other PRs get merged

LucyGwilliamAdmin:
@brockfanning thanks so much for the feedback! 

I have split this PR up into indivdual PRs for each new feature. 

I will do some work to make the contrast button change optional and will then submit separate PR for that. 

Closing this now

PR #292, opened 2019-08-12
Change contrast button

LucyGwilliamAdmin:
Test branch: https://sdg.mango-solutions.com/high-contrast/

### DIFF
--- a/_includes/assets/js/accessibility.js
+++ b/_includes/assets/js/accessibility.js
@@ -84,8 +84,9 @@ function getContrastToggleLabel(identifier){
 function imageFix(contrast) {
   if (contrast == 'high') {
     _.each($('img:not([src*=high-contrast])'), function(goalImage){
+      if ($(goalImage).attr('src').slice(0, 35) != "https://platform-cdn.sharethis.com/") {
       $(goalImage).attr('src', $(goalImage).attr('src').replace('img/', 'img/high-contrast/'));
-    })
+      }})
   } else {
     // Remove high-contrast
     _.each($('img[src*=high-contrast]'), function(goalImage){
--- a/_includes/assets/js/accessibility.js
+++ b/_includes/assets/js/accessibility.js
@@ -62,11 +62,11 @@ _.each(contrastIdentifiers, function(contrast) {
     var gaAttributes = opensdg.autotrack('switch_contrast', 'Accessibility', 'Change contrast setting', contrast);
     $('.contrast-switcher').append($('<li />').attr({
       'class': 'nav-link contrast contrast-' + contrast
-    }).html($('<a />').attr(gaAttributes).attr({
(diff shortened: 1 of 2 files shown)


LucyGwilliamAdmin:
@brockfanning could you take a look at this please? for some reason text isn't appearing in the contrast button as it should?

brockfanning:
@LucyGwilliamAdmin I think the problem may be in how you are translating the text. For javascript files, the mechanism for translation is different. Instead of using curly braces and the `t` Jekyll object, you need to use the `translations` Javascript object (with no curly braces).

For example instead of `{{ t.header.disable_high_contrast }}` you would use `translations.header.disable_high_contrast`.

One requirement, though, is that you need to first include the `multilingual-js.html` Jekyll include, like so:
`{% include multilingual-js.html key="header" %}`

This brings the "header" translations into the `translations` Javascript object. More info on this approach is [here](https://open-sdg.readthedocs.io/en/latest/translation/#translation-in-javascript).

LucyGwilliamAdmin:
Thanks @brockfanning would `{% include multilingual-js.html key="header" %}` go into accessibility.js?

LucyGwilliamAdmin:
And do you still need to do this if the translation is sitting in the country's repo e.g. in sdg-indicators/data/translations/en/header.yml?

brockfanning:
The `{% include multilingual-js.html key="header" %}` needs to go either in a Jekyll include or layout. Since you need this in the header, it might make sense to put it in [this file](https://github.com/open-sdg/open-sdg/blob/master/_includes/header.html).

By the way, we would not want to add any "translation keys" to Open SDG that are not already in SDG Translations. In other words we'd want to put `header.enable_high_contrast` and `header.disable_high_contrast` in SDG Translations, before we can rely on them in Open SDG. It's fine to expect countries to translate their own custom text, but we should try to make sure that any hardcoded text is taken care of in SDG Translations, if possible.

LucyGwilliamAdmin:
@brockfanning thanks, what do you normally do for translations? Do you ask someone who speaks the language to translate or could I use Google Translate?

brockfanning:
My hope with sdg-translations was that we could cultivate a "network" of translators to help with that. I haven't done a great job there, so I think that Google Translate is probably the only way to keep it moving, ideally followed by feedback from native speakers or translators.

LucyGwilliamAdmin:
@brockfanning do you have any idea what might be causing this error?

LucyGwilliamAdmin:
Just realised I'm overwriting header.html - will check if works now

LucyGwilliamAdmin:
Ok, so this is working here:
https://sdg.mango-solutions.com/high-contrast1/

However not great when window is made smaller and disappears on mobile, causing problems with mobile menu bar

LucyGwilliamAdmin:
Closing PR as created new cleaner branch

Issue #295, opened 2019-08-14
Customise data notice which US site has


LucyGwilliamAdmin:
Can we do this:
![image](https://user-images.githubusercontent.com/43404358/63017019-0bd9cc80-be8d-11e9-8320-824f8b7ffc83.png) (https://sdg.data.gov/17-3-2/)

But customise it e.g. have fields in metadata which provides text and colour?

brockfanning:
The US site uses a [custom data file](https://github.com/GSA/sdg-indicators-usa/blob/develop/data/sdg_data_notices.csv) to manage those data notices. The "notice" column corresponds to a translation key in their [fork of sdg-translations](https://github.com/GSA/sdg-translations/blob/develop/translations/en/data_notice.yml). The template that renders it is [here](https://github.com/GSA/sdg-indicators-usa/blob/develop/_includes/components/indicator-content.html).

Controlling it in the metadata would also be a fine approach. 

The only worry I would have with giving control of the colour to data managers, is that some colours may result in insufficient contrast.

LucyGwilliamAdmin:
@brockfanning on the test branch I have used the warning, danger and success styles that already exist so when high contrast mode is activated the colour of the box and the text changes to the correct colour. Maybe we could keep those and then add three more neutral colours (so we aren't using RAG) and then the data managers have a range of colours to choose from, which will also be accessible in both default-contrast and high-contrast?

brockfanning:
I'll go ahead and close this, since the feature made it in.

PR #296, opened 2019-08-14
Add data notice above catch-all

LucyGwilliamAdmin:
Feature for #295

### DIFF
--- a/_includes/components/data-notice.html
+++ b/_includes/components/data-notice.html
@@ -1,0 +1,9 @@
+{% if meta.data_notice_heading %}
+  <div class="alert {{ meta.data_notice_colour }}">
+      <i class="fa fa-exclamation-circle" aria-hidden="true"></i>
+      <div class="data-notice">
+        <h4>{{ meta.data_notice_heading }}</h4>
+        <p>{{ meta.data_notice_text }}</p>
+      </div>
+  </div>
+{% endif %}
--- a/_includes/components/data-notice.html
+++ b/_includes/components/data-notice.html
@@ -1,5 +1,5 @@
 {% if meta.data_notice_heading %}
-  <div class="alert {{ meta.data_notice_colour }}">
+  <div class="alert {{ meta.data_notice_class }}">
       <i class="fa fa-exclamation-circle" aria-hidden="true"></i>
       <div class="data-notice">
         <h4>{{ meta.data_notice_heading | t }}</h4>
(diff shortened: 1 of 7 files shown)


LucyGwilliamAdmin:
@brockfanning made those changing. Can be seen working [here](https://sdg.mango-solutions.com/data-notice-test1/1-1-1/)

LucyGwilliamAdmin:
There's 3 fields which can be used in the metadata file to configure this box:
* data_notice_heading
* data_notice_text
* data_notice_colour - there's 3 set options for this: [success](https://sdg.mango-solutions.com/data-notice-test1/1-1-1/), [danger](https://sdg.mango-solutions.com/data-notice-test1/1-2-1/) and [warning](https://sdg.mango-solutions.com/data-notice-test1/1-2-2/)

I did start taking a look at adding some set neutral options following Matt's feedback but didn't want to do too much on it until we've decided if and how we're going to use them.

brockfanning:
Looking good. A few more comments:
1. Not a huge deal, but since the options are success/warning/danger, maybe a more accurate field name would be something like "data_notice_level", or "data_notice_type", or if you want to get totally technical, "data_notice_class".
2. We should add some documentation of this feature in [this file](https://github.com/open-sdg/open-sdg/blob/master/docs/metadata-format.md).
3. This might be a good chance to try out adding an automated test. If you're interested, here would be the steps:
    1. Add the required metadata to a particular indicator, such as 1.2.1 in [this file in the testing data repo](https://github.com/open-sdg/open-sdg-data-testing/blob/develop/meta/1-2-1.md). For example you might give it a `data_notice_heading` of "My data heading".
    2. Add a new file [here](https://github.com/open-sdg/open-sdg/tree/master/tests/features) called something like `DataNotice.feature`
    3. Use another of the .feature files as a guide. The "scenario" might be something like:
        ```
        Scenario: Data notices appear on indicator pages
            Given I am on "/1-2-1"
            Then I should see "My data heading"
        ```
    Thereafter, we could rest assured that the data notice functionality would never get broken. :)

brockfanning:
@LucyGwilliamAdmin A note about custom selectors in tests. You could add the following to [this file](https://github.com/open-sdg/open-sdg/blob/master/tests/features/support/mink.js):

```
"success data notice": "div.alert.success",
```

And then I believe your step in the test would be the following:

```
Then I should see 1 "success data notice" element
```

brockfanning:
Also generally each "Scenario" has a different name, depending on what is being tested. In this case, actually, I think you could put all the steps into a single scenario, like maybe:

```
  Scenario: Data notices appear on indicator pages
    Given I am on "/1-2-1"
    Then I should see "My data notice heading"
    And I should see "My data notice text"
    And I should see 1 "success data notice" element
```

PR #311, opened 2019-08-18
get list of sources for graph footer

### LINKED ISSUE #204: Only 1 source displaying in graph/table footer
No matter how many sources an indicator has, only the first source name shows in the graph/table footer. For example, in the image below the indicator data comes from two diff sources (Office for National Statistics and World Bank), however only Office for National Statistics is shown in the footer. 

image

Is there anyway that this could be the set of all source organisations from each of the sources? 

https://github.com/open-sdg/open-sdg/blob/master/_layouts/indicator.html#L46

Tried to make a start [here](https://github.com/open-sdg/open-sdg/blob/multiple_sources/_layouts/indicator.html) but no idea how to put a set into HTML

LucyGwilliamAdmin:
Fixes #204 

Although not sure how to code it with translation (| t) filter. Have tried a couple of different things but haven't had any luck

https://sdg.mango-solutions.com/multiple-sources-1

### DIFF
--- a/_layouts/indicator.html
+++ b/_layouts/indicator.html
@@ -47,7 +47,8 @@
     {% assign src_active = "source_active_" | append: i %}
     {% if meta[src_active] == true %}
       {% assign src_org = "source_organisation_" | append: i %} 
-      {% assign sources_list = sources_list | push: meta[src_org]  %}
+      {% assign src_org_translated = meta[src_org] | t %}
+      {% assign sources_list = sources_list | push: src_org_translated  %}
     {% endif %}
   {% endfor %}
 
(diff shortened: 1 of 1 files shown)


brockfanning:
Looks good, just need to add the `t` filter. In cases like this, each individual `source_organisation_*` needs to be translated separately. So you probably need to put each one into a variable so you can run it through `t` before "pushing" it onto the list. Eg, instead of this:

`{% assign sources_list = sources_list | push: meta[src_org]  %}`

...you might try something like this:

```
{% assign src_org_translated = meta[src_org] | t %}
{% assign sources_list = sources_list | push: src_org_translated  %}
```

To be clear, the reason that you can't do this:

`{% assign sources_list = sources_list | push: meta[src_org] | t  %}`

...is that (I think) Jekyll will try to run the entire `sources_list` through `t`, and `t` only works for individual strings.

LucyGwilliamAdmin:
@brockfanning great - thanks for your help!

PR #315, opened 2019-08-20
Change chart attributes from canvas to html

### LINKED ISSUE #73: Download chart as image

LucyGwilliamAdmin:
Fixes #73 
Can be seen on this [feature branch](https://sdg.mango-solutions.com/html-chart)

### DIFF
--- a/_includes/components/charts/chart.html
+++ b/_includes/components/charts/chart.html
@@ -5,7 +5,7 @@
     </div>
 
     {% assign graph_template = 'components/charts/' | append: include.graph_type | append: '.html' %}
-    <div id="selectionsChart" class="chart-text">
+    <div id="selectionsChart">
       <div id="chart-canvas">
         <h3 class="chart-title">{{ meta.graph_title | t }}</h3>
         <div id="chart" class="plot-container">
--- a/_includes/components/charts/chart.html
+++ b/_includes/components/charts/chart.html
@@ -13,8 +13,8 @@
         </div>
         <div id="plotLegend"></div>
       </div>
-      <button id="btnSave" title={{ translations.indicator.download_chart_image_title }} class='btn btn-primary btn-download'>
-        {{ translations.indicator.download_chart_image }}
+      <button id="btnSave" title={{ t.indicator.download_chart_image_title }} class='btn btn-primary btn-download'>
+        {{ t.indicator.download_chart_image }}
       </button>
     </div>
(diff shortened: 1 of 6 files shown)


brockfanning:
Looks good, but we'll need to get `indicator.download_chart_image_title` added to sdg-translations.
Also when working in a Jekyll file, it would be `t` and not `translations`. So it would be `t.indicator.download_chart_image_title` instead of `translations.indicator.download_chart_image_title`. This is admittedly confusing with the javascript's `translations`. Maybe we should think about making them the same.

LucyGwilliamAdmin:
@brockfanning thanks. I've just changed translations to t and I think Ive made a PR to set-translations for both download_chart_image and download_chart_image_title

_includes/components/charts/chart.html
```diff
@@ -5,11 +5,18 @@
...
-        {% include {{graph_template}} %}
+    <div id="selectionsChart">
+      <div id="chart-canvas">
+        <h3 class="chart-title">{{ meta.graph_title | t }}</h3>
+        <div id="chart" class="plot-container">
+          {% include {{graph_template}} %}
+        </div>
+        <div id="plotLegend"></div>
+      </div>
+      <button id="btnSave" title={{ t.indicator.download_chart_image_title }} class='btn btn-primary btn-download'>
```
LucyGwilliamAdmin: @brockfanning does `title={{ t.indicator.download_chart_image_title }}` need to be `title="{{ t.indicator.download_chart_image_title }}"` asking since I just tested this [here](https://sdg.mango-solutions.com/html-chart/1-1-1/) and the title seems to be just 'Download' instead of 'Download chart as image' which is what is in the translations file. Tested for armenia locally and same there (just took first word):

image

brockfanning: Yes that's a good point - HTML attributes always need to contained in quotes.

LucyGwilliamAdmin:
@brockfanning ok great, button title is working as it should now. Do you think this is ready?

PR #324, opened 2019-08-21
Fix mobile menu

LucyGwilliamAdmin:
When the screen readers is focused on the Menu link on mobile, the reader reads it as "Menu toggle. Click to expand or collapse the menu".

The link now has aria-expanded and aria-haspopup attributes that help the reader know this is an expandable menu

When the user clicks on the Menu, the focus switches automatically to the first element in the menu ("Goals"). This will make it easier to navigate the menu from a screen reader.

### DIFF
--- a/_includes/assets/js/menu.js
+++ b/_includes/assets/js/menu.js
@@ -10,7 +10,7 @@ $(function() {
   
   this.topLevelMenuToggle = document.querySelector("#menuToggle");
   
-  topLevelMenuToggle.addEventListener("click", function(){
+  this.topLevelMenuToggle.addEventListener("click", function(){
     setTopLevelMenuAccessibilityActions();
   });
   function setTopLevelMenuAccessibilityActions(){
--- a/_includes/assets/js/menu.js
+++ b/_includes/assets/js/menu.js
@@ -7,6 +7,34 @@ $(function() {
     $('.top-level li').removeClass('active');
     $('.top-level span').removeClass('open');
   };  
+  
+  this.topLevelMenuToggle = document.querySelector("#menuToggle");
+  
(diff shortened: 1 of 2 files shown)


brockfanning:
The code looks great, very clear. I'm just wondering if we should add this javascript into the existing [menu.js file](https://github.com/open-sdg/open-sdg/blob/master/_includes/assets/js/menu.js) instead of using inline javascript. For developers it's nice for things to be all in one place, and generally it's a better practice to have javascript in separate files, as opposed to inline javascript.

LucyGwilliamAdmin:
Thanks @brockfanning just moved it

LucyGwilliamAdmin:
Although getting circle error now

brockfanning:
I think it might be a javascript error. Would you be able to test locally and look in the console, or put this up on a testing server?

LucyGwilliamAdmin:
Built this here: https://sdg.mango-solutions.com/mobile-menu-fix/
Having this error:
image

brockfanning:
Hmm, yes I think the "this" is a little confusing. Can I suggest this? Instead of
```
this.topLevelMenuToggle = document.querySelector("#menuToggle");
```
...try
```
var topLevelMenuToggle = document.querySelector("#menuToggle");
```

This should let it work the way it was before (without the `this.`).

LucyGwilliamAdmin:
@brockfanning I'm testing 0.8.0 release on UK site and I'm getting this error:
image
then when I click on it, this appears:
image
Do you know what might be causing this?

LucyGwilliamAdmin:
This returns `null` in the console:
image

LucyGwilliamAdmin:
@brockfanning nevermind, we're overwriting header.html

PR #325, opened 2019-08-21
Add optional contrast button

LucyGwilliamAdmin:
Using the line `contrast_type: long` in their config file, countries can choose whether to use the small 'A' contrast buttons or the longer 'Default contrast'/'High contrast' button

Here's a feature branch where `contrast _type: long`:
https://sdg.mango-solutions.com/optional-contrast/

### DIFF
--- a/_includes/assets/js/accessibility.js
+++ b/_includes/assets/js/accessibility.js
@@ -70,7 +70,7 @@ var accessibilitySwitcher = function() {
       'href': 'javascript:void(0)',
       'title': 'Set to ' + contrast + ' contrast',
       'data-contrast': contrast,
-    }).text(getContrastToggleLabel(contrast)).click(function() {
+    }).html(getContrastToggleLabel(contrast).replace(" ", "<br/>")).click(function() {
       setActiveContrast($(this).data('contrast'));
       imageFix(contrast);
     })));
--- a/_includes/assets/js/accessibility.js
+++ b/_includes/assets/js/accessibility.js
@@ -3,7 +3,7 @@ var accessibilitySwitcher = function() {
   var contrastIdentifiers = ['default', 'high'];
 
   function setActiveContrast(contrast) {
-    var contrastType = "{{ site.contrast_type }}
+    var contrastType = "{{ site.contrast_type }}"
     _.each(contrastIdentifiers, function(id) {
       $('body').removeClass('contrast-' + id);
     });
--- a/_includes/assets/js/accessibility.js
+++ b/_includes/assets/js/accessibility.js
@@ -3,9 +3,13 @@ var accessibilitySwitcher = function() {
   var contrastIdentifiers = ['default', 'high'];
 
   function setActiveContrast(contrast) {
+    var contrastType = "{{ site.contrast_type }}
     _.each(contrastIdentifiers, function(id) {
       $('body').removeClass('contrast-' + id);
     });
+    if(contrastType === "long"){
+	    $("body").addClass("long");
+    }
     $('body').addClass('contrast-' + contrast);
 
     createCookie("contrast", contrast, 365);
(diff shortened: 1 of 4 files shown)


brockfanning:
Sorry, forgot to mention a couple things about the configuration document:
1. So far I've tried to keep them alphabetical (though I'm not sure how useful that is)
2. The basic template is something like this:

```
### name_of_setting

Description of setting

(three backticks)nohighlight
name_of_setting: most common value of setting
(three backtickets)
```

(replacing "(three backticks)" with three actual backticks...)

brockfanning:
@LucyGwilliamAdmin This might still be in progress, but in case it helps, when I test with `contrast_type: long` their appears to be a problem with width. I think maybe it's from [this line](https://github.com/open-sdg/open-sdg/blob/master/assets/css/default.scss#L268)?

brockfanning:
@LucyGwilliamAdmin This might still be in progress, but in case it helps, when I test with `contrast_type: long` their appears to be a problem with width. I think maybe it's from [this line](https://github.com/open-sdg/open-sdg/blob/master/assets/css/default.scss#L268)?

brockfanning:
@LucyGwilliamAdmin This might still be in progress, but in case it helps, when I test with `contrast_type: long` their appears to be a problem with width. I think maybe it's from [this line](https://github.com/open-sdg/open-sdg/blob/master/assets/css/default.scss#L268)?

brockfanning:
@LucyGwilliamAdmin This might still be in progress, but in case it helps, when I test with `contrast_type: long` their appears to be a problem with width. I think maybe it's from [this line](https://github.com/open-sdg/open-sdg/blob/master/assets/css/default.scss#L268)?

LucyGwilliamAdmin:
No longer in progress so thanks for pointing that out. If `contrast_type: long` then the styling should come from the bottom of the css file, from [this line](https://github.com/open-sdg/open-sdg/blob/optional-contrast-final/assets/css/default.scss#L2050) onwards.

Where are you testing to? Just so I can see what the button looks like

brockfanning:
Sorry to spam before, think I lost connection...

Yes I'm seeing the same problem on your test branch. I was testing locally with the Kazakhstan site.

LucyGwilliamAdmin:
Can you explain what the issue is please? I don't think I'm seeing it.

LucyGwilliamAdmin:
I've just opened it in IE. Is this the same as what you're getting?
image
In Chrome it looks like this:
image

brockfanning:
Yes, I'm also seeing that problem in Firefox. Sorry for not doing a screenshot. So sounds like it's fine in Chrome but too narrow in IE and Firefox.

LucyGwilliamAdmin:
Also a problem on Edge. For some reason two classes are being applied:
image

image
Whereas we would just want the second one to be applied if `contrast_type: long` which is working on Chrome

LucyGwilliamAdmin:
@brockfanning do you know what might be causing this?

LucyGwilliamAdmin:
Ok looks as if `fit-content` is only really supported by Chrome so think that's the issue.

@brockfanning @SavvasStephanides what would you normally use instead?

brockfanning:
I'm not sure, but maybe `width: auto`? Or maybe we could change things so that the `width: 25px` is only applied when the `contrast_type` is _not_ "long"?

LucyGwilliamAdmin:
I've managed to fix desktop with `width: auto`, doesn't work with mobile if I use `display: table-caption` which forces the text to wrap on white space. One way around it I guess it to replace ` ` in the text with `<br/>` . Is there a way to do this with code? - don't think it should be in translations file,

LucyGwilliamAdmin:
Tried this:
image
but got this:
image
When I look at the text in the console, instead of `High</br>contrast`, it is `Default&lt;br/&gt;contrast`

LucyGwilliamAdmin:
@brockfanning I've managed to fit the width problem. Looks like this now:
https://sdg.mango-solutions.com/optional-contrast/

PR #338, opened 2019-08-28
Use timeouts to hack together working disagg/unit forcing.

### LINKED ISSUE #262: Still slight issue with default selections

brockfanning:
This needs testing, but in theory fixes #262 

I'm not crazy about this fix, but it's quick.

### DIFF
--- a/_includes/assets/js/indicatorView.js
+++ b/_includes/assets/js/indicatorView.js
@@ -91,6 +91,7 @@ var indicatorView = function (model, options) {
           $('#fields .variable-options input[type="checkbox"]')
             .filter('[data-field="' + fieldToSelect + '"]')
             .filter('[value="' + fieldValue + '"]')
+            .filter(':not(:checked)')
             .first()
             .click();
         }


LucyGwilliamAdmin:
@brockfanning I could have a go at creating the tests for practice if you can look over them afterwards?

brockfanning:
@LucyGwilliamAdmin That would be awesome. Generally my approach for these is:

1. Fabricate some data and/or metadata that will trigger the behavior you want to test, and pick a random indicator in [open-sdg-data-testing](https://github.com/open-sdg/open-sdg-data-testing) to use. **Make sure not to use an indicator that's already being used for another test.**
2. Write the test itself, targetting that particular indicator. Eg: `Given I am on "/1-5-1"`
3. Push up a PR and confirm that the test passes
4. Make another commit that points the test to a different, unrelated indicator. Eg: `Given I am on "/1-6-1"`
5. Confirm that the test now fails
6. Point back to the original indicator

Note that you can also run tests locally by going to the project root in the terminal and running: `make test`. 

I believe I've got the issues worked out for this one, so you're welcome to start on the test.

LucyGwilliamAdmin:
@brockfanning I've tried to make a start on the test but I've got a little stuck...
So far I've got this
`Feature: Force disagg/units`

`As a site visitor`
`I want all charts to show data by default`
`So that I don't think that there's a problem with the chart`
  
`Scenario: Data shows on chart even if headline data isn't available`
`Given I am on "/1-4-2"`
`And I wait 0.5 seconds` - would this be 3.5 seconds or 0.5 seconds since a lot of the tests say 3 seconds plus the 0.5 seconds for this new feature?
`Then I should see 1 (graph line) element` - since the line is within the canvas I can't actually isolate an element for this?

brockfanning:
@LucyGwilliamAdmin Regarding the "I wait" step, I'd just do 1 second to start, like the other tests. Hopefully that works, but it can be increased if the test is failing.

Regarding the "graph line" element - I have never bothered with trying to test for graph line elements. It would actually be a very high level of effort, I think, because "canvas" is involved. So I have just tested for the "chart legend item" element, [like here](https://github.com/open-sdg/open-sdg/blob/master/tests/features/Headline.feature#L15). I'd recommend you do the same here.

LucyGwilliamAdmin:
I've uploaded the feature for the test but the travis check has failed. When I go on to open-sdg-site-testing the feature also isn't working there (even though) it was working on the real life data on UK test site when I tested it.

brockfanning:
Hmm, I'm not sure why the test for 1-4-1 passed, because it doesn't appear to be working: https://open-sdg.github.io/open-sdg-site-testing/1-4-1/

I'll take a closer look.

LucyGwilliamAdmin:
Yes, looks like even though an option has been automatically selected from the parent (Group 1), an option isn't being automatically selected from the child (Group 2)

brockfanning:
Yes I think that there are still cases that don't work. I made some progress fleshing out the tests, but got sidetracked. I'll try to push up where I'm at today.

brockfanning:
@LucyGwilliamAdmin Could you give this another review and/or manual test? I've fixed one bug (I think) and done a revamp of the Headline.feature tests. Hopefully they're more comprehensive now, and also include a lot of definitions/explanation, since this functionality is something of a mind-bender. What do you think?

PR #344, opened 2019-09-07
Support flattened global translations

brockfanning:
The sdg-translations project now has "flattened" global translation keys. For example, instead of this:
```
1:
    title: Foo
2:
    title: Bar
```
...now has this:
```
1-title: Foo
2-title: Bar
```

As a consequence several parts of the templates in Open SDG need to be adjusted. This is an attempt at that, as well as a general cleanup/deprecation of some indicator variables.

### DIFF
--- a/_includes/components/breadcrumb.html
+++ b/_includes/components/breadcrumb.html
@@ -1,7 +1,7 @@
 <ol class="breadcrumb">
   <li><a href="{{ site.baseurl }}{{ baseurl_folder }}/">{{ t.general.home }}</a></li>
-  {% if goal_href %}
-      <li><a href="{{ goal_href }}" title="{{ goal_title }}">{{ t.general.goal }} {{ goal_number }}</a></li>
+  {% if goal_uri %}
+      <li><a href="{{ goal_uri }}" title="{{ goal_title }}">{{ t.general.goal }} {{ goal_number }}</a></li>
       <li class="active">{{ t.general.indicator }} {{ page.indicator }}</li>
   {% else %}
       <li class="active">{{ t.general.goal }} {{ goal_number }}</li>
--- a/_includes/indicator-variables.html
+++ b/_includes/indicator-variables.html
@@ -12,10 +12,11 @@
 {% assign headline = site.data.headline[indicator_id] %}
 
 {%- assign goal_number = meta.sdg_goal | downcase -%}
-{%- assign goal = site.data.translations[default_language].global_goals[goal_number] -%}
-{%- assign translated_goal = t.global_goals[goal_number] -%}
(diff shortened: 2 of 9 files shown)


LucyGwilliamAdmin:
Sorry, we've just realised we're having this again:
image

Is this something to do with the new translation changes?

brockfanning:
Hmm, let me take a look. More soon.

LucyGwilliamAdmin:
Upgraded but still having issue

brockfanning:
Actually it looks like this is happening on the [staging site](https://onsdigital.github.io/sdg-indicators/1-1-1/) already, so I don't think it's related to this PR. Do you have any idea when that started? Or did it never get resolved? I remember we were looking into that recently.

brockfanning:
@LucyGwilliamAdmin Think I see the problem. The next Open SDG upgrade (0.9.0) will need to be simultaneous with an SDG Translations upgrade (0.8.0, already released). So try changing [this line](https://github.com/ONSdigital/sdg-indicators/blob/flattened-translations/_config.yml#L33) to version 0.8.0.

LucyGwilliamAdmin:
So I have upgraded to 0.8.0 and that fixed the indicator title but now the goal title is gone from the ribbon on the goal page:
image

brockfanning:
@LucyGwilliamAdmin Sorry for the confusion - I meant to upgrade to sdg-translations 0.8.0 in your "flattened-translations" branch only - not in "develop". Any site using sdg-translations 0.8.0 will also need to be using open-sdg 0.9.0 (which we haven't released yet) and vice versa.

Hmm, given this issue maybe we should go ahead and release 0.9.0 of open-sdg?

LucyGwilliamAdmin:
@brockfanning sorry I thought you meant it would fix the issue on the staging/live site (which it did).

I think that might be best.

LucyGwilliamAdmin:
Should I finish the set chart axis feature first?

LucyGwilliamAdmin:
@brockfanning the tab title is also missing e.g. 
image
but I guess that is also something to do with the upgrade of sdg-translations.

If we don't want to do a new release just to fix UK site, I can downgrade sdg-translations to 0.7.0 - we will then be back with the issue of having duplicate indicator codes in the banner on indicator pages. No idea where that came from - as we haven't made any commits that would have messed with that.

brockfanning:
I think I know what's going on - for now you could try downgrading back to sdg-translations 0.7.0, but also make this change: in Gemfile, change `gem "jekyll-open-sdg-plugins", "~> 0.0.13"` to `gem "jekyll-open-sdg-plugins", "0.0.13"`. I think that will resolve the double-id problem.

LucyGwilliamAdmin:
This has fixed it - thanks!

PR #348, opened 2019-09-09
Add chart type that allows y-axis limits to be set

LucyGwilliamAdmin:


LucyGwilliamAdmin:
@brockfanning does this need a test? I just started a PR on open-sdg-data-testing but then I realised I can't really make a feature as again it uses canvas.

SavvasStephanides:
Do you have a test branch for this?

LucyGwilliamAdmin:
@SavvasStephanides yes, see [here](https://sdg.mango-solutions.com/scaled-graph/16-6-2/)

brockfanning:
Also forgot to answer your question about tests - yes it's going to really hard (if not impossible) to test the existence of particular text inside a canvas. However you can at least add a test to ensure that the chart appears at all. That's better than nothing, I think.

LucyGwilliamAdmin:
Am I using the wrong code here for the if statement or the `meta.-` here? no idea why it's passing and not showing on the feature branch

brockfanning:
Yes, remember that `meta` is a Jekyll (ie, "Liquid") variable. But you're using it as if it were a Javascript variable. So I think it should be something like:
```
{% if meta.graph_min_value and meta.graph_max_value %}
<script>
...javascript code...
</script>
{% endif %}
```

LucyGwilliamAdmin:
Still no luck unfortunately

SavvasStephanides:
@LucyGwilliamAdmin 

1. In JS, the whole statement needs to be in a bracket (eg. `if(statement1 && statement2){}`)

2. What are `graph_min_value` and `graph_max_value` supposed to return? I'm a bit confused. Be very careful with equalities in JS it's very confusing. From the name it seems to be returning a numeric value which when added as part of an `if` statement, will return `true` solely because you've defined it.

LucyGwilliamAdmin:
@SavvasStephanides ah ok
I've moved the if statement into liquid but you're right `graph_min_value` and `graph_max_value` are numbers. For the case on the test branch they are 1 and 10 respectively.

LucyGwilliamAdmin:
Is there a better way for me to check if they exist in the metadata file?

SavvasStephanides:
> I've moved the if statement into liquid but you're right `graph_min_value` and `graph_max_value` are numbers. For the case on the test branch they are 1 and 10 respectively.

Then both will return `true` in an `if` statement as long as they're defined and non-zero. Not sure if that's the case with Liquid / Jekyll. Might be worth reading up on ["Truthy" and "Falsy"](https://www.sitepoint.com/javascript-truthy-falsy/) in JS for future reference

LucyGwilliamAdmin:
so to check if they exist in JS the code would be..
`<script>`
`if({{meta.graph_min_value}} && {{meta.graph_max_value}}){}`
`</script>`

SavvasStephanides:
I'm not sure about this since you're mixing JS and Jekyll syntax

LucyGwilliamAdmin:
do you know how I would do it in just JS?

SavvasStephanides:
Seems like including Jekyll syntax is the only way, not sure if @brockfanning knows of a better way. Give it a go and let us know!

LucyGwilliamAdmin:
I have tried loads different things and don't seem to be having any luck even though all checks are passing

brockfanning:
Yes, there's definitely no way to consult the `meta` variable without using Liquid. I'd be curious what specifically is happening. Is the `chartConfigAlter` function not appearing at all? Or is the `chartConfigAlter` function appearing but not having any effect? Or is there a Javascript error happening? Something else? It would be worth checking the source of the HTML page and the developer console for these answers.

LucyGwilliamAdmin:
No errors and the function isn't appearing in the source at all

SavvasStephanides:
> No errors and the function isn't appearing in the source at all

Could be an issue with your indicators branch having the wrong `open-sdg` version

LucyGwilliamAdmin:
Oh dear, I'm so mad at myself

LucyGwilliamAdmin:
works :) I'm going to close this PR and make a new branch though as I've made way too many commits trying to figure the problem out. Will also make the other changes @brockfanning suggested

LucyGwilliamAdmin:
@brockfanning @SavvasStephanides should this script go in to the includes/assets/js folder so that it can be reused by both line charts and bar charts or should it sit in the chart html files?

LucyGwilliamAdmin:
I am, however, getting this error on indicators which don't have `graph_min_value` and `graph_max_value in their indicators:
image
image

LucyGwilliamAdmin:
Also isn't working for bar charts:
https://sdg.mango-solutions.com/scaled-graph/5-5-1/

LucyGwilliamAdmin:
If changed the order around e.g. if ... <script> instead of <script> if... and error is gone and bar chart is working now.

brockfanning:
There's some room for preferences here. As for the conditional - personally I would prefer the conditional be in Liquid, so that the script tag doesn't even appear if the metadata fields aren't there. As for whether to separate out that code so that it can be re-used for both bar and line - I like that idea in general. However I also wouldn't object if you just copy/paste the same snippet into both line.html and bar.html. It's a tough decision sometimes about whether it's better to avoid code duplication by abstracting things, or whether it's better to keep things simple and just copy/paste some code.

PR #351, opened 2019-09-11
Allow min and max value be set on y axis

LucyGwilliamAdmin:
For #348 

This PR will allow data providers/editors to set a min and max value on the y-axis using new metadata fields `graph_min_value` and `graph_max_value`

Test branch:
https://sdg.mango-solutions.com/set-axis-test
Bar chart test:
https://sdg.mango-solutions.com/set-axis-test/5-5-1/
Line graph test:
https://sdg.mango-solutions.com/set-axis-test/16-6-2

### DIFF
--- a/_includes/components/charts/bar.html
+++ b/_includes/components/charts/bar.html
@@ -4,3 +4,24 @@ apart from providing the canvas tag. For details on the Chart.js configuration
 being used, see `createPlot()` in `indicatorView.js`.
 {% endcomment %}
 <canvas></canvas>
+{% if meta.graph_min_value and meta.graph_max_value %}
+  <script>
+    opensdg.chartConfigAlter = function(config) {
+      var overrides = {
+        options: {
+          scales: {
+            yAxes: [{
+              ticks: {
+                min: {{ meta.graph_min_value }},
+                max: {{ meta.graph_max_value }}
+                }
+              }]
+            }
+          }
+        }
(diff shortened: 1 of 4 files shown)


LucyGwilliamAdmin:
@brockfanning so I have added a feature and a test on open-sdg-data-testing - just wondering if it's worth me changing the indicator code in the feature to check that it fails as it shouldn't as the chart should show regardless?

SavvasStephanides:
Lucy, can you make sure that it still works when `meta.graph_min_value` is set to 0 (or max, but less likely scenario)? In JS `graph_min_value` would return `false` if it's in an `if` statement and set to 0. Hopefully Jekyll is different in that respect but worth checking nonetheless.

LucyGwilliamAdmin:
Example for when min is 0:
https://sdg.mango-solutions.com/set-axis-test/1-1-1/

Example for when min is negative (-5):
https://sdg.mango-solutions.com/set-axis-test/8-2-1/

SavvasStephanides:
Great. Thanks!

brockfanning:
Already been merged, but just a few more nitpicks:
1. The indentation is still not quite right, starting with the closing brace beneath "max".
2. It might be better if the name of the test scenario was more descriptive for this particular case, like maybe "Indicators with a min/max value display data in a chart"
3. The steps beneath a scenario are generally indented 2 spaces

PR #358, opened 2019-09-13
Bugfix - syntax problem in goal-by-target layout

brockfanning:
This was a regression bug caused by the last release (0.9.0). We should probably write a test for this, to avoid this happening again.


brockfanning:
@LucyGwilliamAdmin Just FYI, this layout ("goal-by-target") is currently broken, so I would not spend any time testing this - ie, it can only get better. ;)

brockfanning:
Just a note that I added a test for this, and some associated plumbing. I'll go ahead and merge since the tests are passing.

PR #360, opened 2019-09-14
Functionality and docs for a 'languages_public' mapping

### LINKED ISSUE #349: Let language codes differ between code and URL
Right now the language codes in the code (rather, the folder names for the translation files) must be the same as the language code in URLs. In other words, if the URL is going to be `http://my-platform.org/foo` then the language code (folder name for the translation files) _must_ be `foo`.

However it may be necessary in some cases for language codes in the code (the folder names) to be different from what appears in the URL. This is because the language codes in the code may need to adhere to a global standards, while the language codes in the URLs may be whatever the site admins prefer.

For example, the standard language code for Armenian is "hy", but the site admins prefer the URLs use "am".

Is it possible to implement some sort of mapping to easily support this? Or is this more trouble than it's worth?

brockfanning:
Fixes #349 

For more details see the issue #349


LucyGwilliamAdmin:
@brockfanning sorry I haven't reviewed this yet as have been out of office until today. What do you think is the best way to test this?

brockfanning:
No problem @LucyGwilliamAdmin. There's a test included in the PR which is passing, along with all the existing tests.

If you want to do a manual test, I guess the most critical thing is to make sure that:
1. Existing single-language sites still work
2. Existing multi-language sites still work

After that of course you could test the actual behavior being added here. Though I consider that optional, since it's new functionality. FYI, this is something that we'll likely use (eventually) for both Armenia and Kazakhstan, since they are both using non-standard language codes. So the functionality will eventually get a real-world test.

brockfanning:
@LucyGwilliamAdmin @SavvasStephanides Just a ping about this one - it's a pretty specific use-case, but this may be needed soon for Armenia and/or Kazakhstan. Let me know if I can clarify anything.

brockfanning:
@LucyGwilliamAdmin No worries, thanks for testing!

PR #364, opened 2019-09-18
Footerfield for Copyright

### LINKED ISSUE #329: Footer field for "copyright"
Either adding a footer field for a copyright entry or making the footer fields generally configurable.

The second approach could be even better, as countries could complement what ever they want.

MoBosse:
Fixes  #329
gives the possibility to create a field for copyright information in the metadata


brockfanning:
@MoBosse Looks great. Just two recommendations:

1. To allow for translation of the field, I would suggest instead of `data-copyright="{{ meta.copyright }}` that you have `data-copyright="{{ meta.copyright | t }}`.
2. We should mention this field in the documentation, perhaps in [this section](https://github.com/open-sdg/open-sdg/blob/master/docs/metadata-format.md#national-metadata).

PR #365, opened 2019-09-19
Update versions in the gemspec file.

brockfanning:


brockfanning:
Our gemspec file was getting a little out of date. Updating it in anticipation of release 0.9.2.

PR #368, opened 2019-09-20
Configurable URLs for the edit buttons

### LINKED ISSUE #367: Configurable URLs for editing data/metadata
Right now the URLs for editing data/metadata are hardcoded to use Prose.io. Can we make them configurable? In our docs we talk about the possibility of alternatives to Github, but as long as we're hardcoding Prose.io then we're essentially requiring the use of Github.

brockfanning:
Fixes #367

### DIFF
--- a/_includes/components/edit-buttons.html
+++ b/_includes/components/edit-buttons.html
@@ -1,0 +1,13 @@
+{% assign indicator_id_slug = meta.indicator | slugify %}
+
+{% if site.data_edit_url %}
+<a class="btn btn-primary" href="{{ site.data_edit_url | replace: '[id]', indicator_id_slug }}">{{ t.indicator.edit_data }}</a>
+{% else %}
+<a class="btn btn-primary" href="http://prose.io/#{{ site.org_name }}/{{ site.repo_name }}/edit/{{ site.branch }}/{{ site.data_dir_for_edits | default: 'data' }}/indicator_{{ meta.indicator | slugify }}.csv">{{ t.indicator.edit_data }}</a>
+{% endif %}
+
+{% if site.metadata_edit_url %}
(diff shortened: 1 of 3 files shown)


brockfanning:
This can be demoed [here](https://brockfanning.gitlab.io/open-sdg-site/1-1-1/) by clicking on the "Edit" buttons. The "Edit Data" and "Edit Metadata" buttons should both take you to the file editor on Gitlab.com, instead of the usual Prose.io.

NOTE though - before testing you should sign up for a (free) Gitlab.com account.

_includes/components/edit-buttons.html
```diff
@@ -0,0 +1,13 @@
+{% assign indicator_id_slug = meta.indicator | slugify %}
+
+{% if site.data_edit_url %}
+<a class="btn btn-primary" href="{{ site.data_edit_url | replace: '[id]', indicator_id_slug }}">{{ t.indicator.edit_data }}</a>
+{% else %}
+<a class="btn btn-primary" href="http://prose.io/#{{ site.org_name }}/{{ site.repo_name }}/edit/{{ site.branch }}/{{ site.data_dir_for_edits | default: 'data' }}/indicator_{{ meta.indicator | slugify }}.csv">{{ t.indicator.edit_data }}</a>
```
SavvasStephanides: Although I understand the reasoning behind this, I find it rather strange that although you have a dedicated `yaml` file for URLs like this, you still need to hardcode a default.

brockfanning: @SavvasStephanides Yeah I agree it would be better not to have the default at all. I just included it for backwards compatibility. In general I've found it easier to code in backwards compatibility when possible. This allows us to quietly change the documentation - meaning new users will use the new config, while ensuring that existing users' sites will not break.

PR #370, opened 2019-09-20
Set goal variables before head

### LINKED ISSUE #369: Variable collisions related to goals
I believe a recent update has resulted in some collisions of global variables (always a danger with this kind of architecture). It's manifesting as missing goal icons and banner backgrounds on indicators pages. I will look for a quick fix.

brockfanning:
Fixes #369 

I believe the best fix is to set goal variables before the head.


brockfanning:
Thanks @LucyGwilliamAdmin I merged and pushed a new tag, 0.9.3.

PR #378, opened 2019-10-06
Metadata tabs configuation

### LINKED ISSUE #339: changing metadata fields
@brockfanning 

How do I change the number and caption of the global metadata fields? I only need about five or six metadata fields. Perhaps I also want to change the name "global metadata" to anything else.

The caption I'think i change with adding a metadata_fields.yml in my translations folder, but where do I change the number of the fields?

And how do I hide the National Metadata?

brockfanning:
Fixes #339 

This implements what is discussed in #339 and supports control of what metadata tabs display, as well as the ability to hide empty metadata fields. The PR includes documentation of the new config options `metadata_tabs` and `hide_empty_metadata` and also includes automated tests for them.


brockfanning:
Ah, let me think on that. In the PR the "Sources" tab is still hardcoded in, because it's a bit different than the others. But I agree it would make sense to be configurable just like the others.

brockfanning:
@LucyGwilliamAdmin I've made a change to include "Sources" in the configurable tabs. So now, if you are using these configurable tabs, and you want "Sources", then you've got to have an item in the list with the specific key of "sources". See the new notes in the documentation in the PR.

If you could test this out again that would be great!

LucyGwilliamAdmin:
Ok great - I can turn off sources tab on and off. Do you think there's a way that if there was no source metadata in a metadata file, that that tab could be hidden on just those indicators? I only ask as sometimes there isn't a source (e.g. non-statistical indicator). 

Not urgent, so we could merge this as is for now and then come back to it at a later date, if it's a requires a lot of work.

brockfanning:
@LucyGwilliamAdmin That sounds possible. I like the idea of doing that in a separate issue, so I'll go ahead and merge this.

PR #394, opened 2019-11-15
More consistent template variables without globals

brockfanning:
This is a proposal to use some newly-added variables from the jekyll-open-sdg-plugins project. The main purposes of this are:
1. Remove all "global" variables from the Jekyll templates. By "global" I mean variables that are created in one include file and then used throughout other include/layout files. These are problematic for a couple of reasons:
    * There is the risk that the variables will get changed somewhere, which will cause unexpected results
    * Putting important variables in include files can causes problems if those are overridden and need to change
2. Consistent API for accessing properties of goals/targets/indicators. This fixes the problem we currently have where the names of properties are inconsistent. For example, the indicator id is "indicator", the target id is "target_id", and the goal id is "sdg_goal".

As a bonus, all of the things added here are automatically translated, so they don't need to be run through the `t` filter.

NOTE: This is a large PR that changes almost every file, and has plenty of breaking changes. This is intended to be first steps towards release 1.0.0.

Breaking changes:
* `t` has changed to `page.t`. Eg: `t.general.goal` is now `page.t.general.goal`. This is more verbose, but it lets us avoid using a global variable. If there are strong feelings against this we could easily add back in a global `t` variable.
* The include files `indicator-variables.html`, `goal-variables.html`, and `multilingual.html` have all been removed. These were dedicated to global variables that are no longer needed. This means that global variables like `meta`, `current_language`, etc. are no longer available.
* The convoluted logic for indicator names is gone - instead the metadata *must* contain an `indicator_name` field, and that's what is used. This will cause some unexpected changes for countries that may have been relying on overrides of `indicator_name_national`, or falling back to the global indicator names.
* The "subfolder approach" for translation is no longer supported. This was a rarely (never?) used feature that ended up being inferior to the "translation key" approach, and supporting it has mainly just caused confusion and unnecessarily long documentation.
* The `remotedatabaseurl` setting is no longer supported. Sites should use `remote_data_prefix` instead, if not already.

Dependencies:
* This depends on the latest version of jekyll-open-sdg-plugins, which is 1.0.0-rc6.

Documentation:
* This needs to documentation updates, but we might postpone that until we're about ready to release 1.0.0.


LucyGwilliamAdmin:
@brockfanning is this ready for review?

Not really sure where I would start with testing this..

brockfanning:
@LucyGwilliamAdmin It actually shouldn't be too hard to test, but I'd like to also update the documentation. I'll also add some testing instructions. When it's ready for testing I'll ping you and remove the "WIP".

brockfanning:
@LucyGwilliamAdmin I think this is about ready to test. I did a [quick commit](https://github.com/brockfanning/sdg-indicators-1/commit/5d55e34041727229871f1b923582799c41387a68) to demonstrate a test on the UK site. Basically it just involves changing `t` to `page.t` in any overridden layouts/includes. It's fairly easy for the UK because there are only a couple of overridden files. For a country with a lot of overridden files, it would be more work to upgrade. (This PR changes pretty much every layout/include file, so it's safe to assume that *all* overridden files would need to be checked for changes).

For documentation of the changes, check out the documentation updates in the PR. In particular:
* [Info on working with Jekyll templates](https://github.com/brockfanning/open-sdg/blob/stable-release-1/docs/customisation.md#working-with-jekyll-templates)
* [Updates to metadata documentation](https://github.com/brockfanning/open-sdg/blob/stable-release-1/docs/metadata-format.md)

brockfanning:
@LucyGwilliamAdmin Another note - you might notice that the UK indicator names go back to having the double-indicator-id problem. I put up this PR to demonstrate what would need to be to fix that: https://github.com/ONSdigital/sdg-data/pull/1662

PR #397, opened 2019-11-20
Backwards compatibility after recent jekyll-open-sdg-plugins change

brockfanning:


brockfanning:
@LucyGwilliamAdmin I believe this will fix the recent testing failures. This may seem to complicate the templates, but I believe this will only be temporary, until #394 ultimately simplifies things greatly.

FYI the reason for the recent testing failures was that a recent update to jekyll-open-sdg-plugins changed what `page.indicator` was, and so the usage of `page.indicator` needed to be updated in open-sdg.

PR #401, opened 2019-11-22
Add footer to embed tab

LucyGwilliamAdmin:
Added a section below the embed within the embed tab where info about the embedded feature can go e.g. source, footnote.

[Here](http://sdgdev-813006012.eu-west-1.elb.amazonaws.com/5-4-1-embed-deploy/5-4-1/)'s an example of how it would look.

Also I realise to be in line with #394 `meta` should be `page.indicator` but wasn't sure which PR that should be part of?

### DIFF
--- a/_layouts/indicator.html
+++ b/_layouts/indicator.html
@@ -130,9 +130,9 @@
                 <script type="text/javascript" src="https://pym.nprapps.org/pym.v1.min.js"></script>
                 <script>document.querySelector("li.nav-item.embedded-map").addEventListener("click",function(){ var pymParent = new pym.Parent('embeddedmapframe', '{{ meta.embedded_feature_url }}', {});})</script>
               </div>
-              {% if meta.embed_footer %}
+              {% if meta.embedded_feature_footer %}
                 <div id="embedfooter" class="table-footer-text">
-                  <p>{{ meta.embed_footer | t | markdownify}}</p>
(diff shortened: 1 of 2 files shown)


brockfanning:
@LucyGwilliamAdmin I was thinking we should probably do a 0.10.0 release before merge #394 which is a rather large update. Would you like to get this into 0.10.0? If so, this can ignore all the changes in #394.

brockfanning:
@LucyGwilliamAdmin If there is no `meta.embed_footer` then this will still display an empty div. If that's OK, then this is fine as-is. Normally empty divs are harmless, but they can cause a problem if they happen to have any CSS styling associated with them. If you'd like to avoid that happening, you could put the Liquid conditional `{% if... %}` stuff around the div, instead of inside it. What do you think?

LucyGwilliamAdmin:
@brockfanning yes, would like to have this in 0.10.0 - that makes sense, I've moved the if statement to outside. I've also added the field to the metadata docs.

brockfanning:
@LucyGwilliamAdmin Actually a thought - do you think we ought to call it "embedded_feature_footer" to be consistent with the other ones?

LucyGwilliamAdmin:
@brockfanning good point - will change that now

brockfanning:
Cool, looks ready to me.

PR #403, opened 2019-11-25
Changelog entry for 1.0.0 with upgrade instructions and non-technical updates

brockfanning:


brockfanning:
@AnnCorp I pushed up an attempt at the non-technical version of the change log. It may not be what you had in mind, so feel free to suggest changes. I tried to make it similar to the change log, but geared towards administrators or managers, and omitting some of the types of thiings they wouldn't care about. I also retroactively went back through all the other releases we've had (0.1.0, 0.2.0, etc.). Here's a link: https://github.com/open-sdg/open-sdg/pull/403/files#diff-237a39176c5b9b47b96a4d1a423454d0

brockfanning:
@AnnCorp @LucyGwilliamAdmin @SavvasStephanides Given that we're about ready to (pre)release version 1, I think this PR is ready for review.

docs/changelog.md
```diff
@@ -1,13 +1,105 @@
...
+* High-contrast fixes for chart grids/ticks (#593)
+* Indicator cards styling updates (#594)
+* Only show disaggregations that have data in drop-downs (#600)
+* Move units above subcategories (#603)
+* Hide the sub-categories when there are no disaggregations (#604)
+* Automatic site builds whenever the data changes (#605)
+
+Breaking changes:
+
+This is a major version upgrade and contains some breaking changes. Full technical details are available [here](./upgrading-1-0-0). But here is a brief summary of the breaking changes:
```
SavvasStephanides, line 62: Since we're avoiding "here" in links, maybe we should apply it too as well

docs/changelog.md
```diff
@@ -1,13 +1,105 @@
...
+
+* This upgrade should be accompanied by an upgrade to:
+    * jekyll-open-sdg-plugins 1.0.0
+    * sdg-build 1.0.0
+    * sdg-translations 1.0.0
+* Some global Liquid variables have changed. For example:
+    * `t` has changed to `page.t`
+    * `meta` has changed to `page.indicator`
+    * `current_language` has changed to `page.language`
+    * Details on Liquid variables can be found [here](./customisation.md#working-with-jekyll-templates).
```
SavvasStephanides, line 72: as above

docs/changelog.md
```diff
@@ -1,13 +1,105 @@
 <h1>Change Log</h1>
 
+## 1.0.0
+
+* More consistent template variables without globals (#394)
+* Auto fit search bar around text (#407)
```
SavvasStephanides, line 6: Might need to be a bit more specific:

```diff
-text
+placeholder text
```

docs/changelog.md
```diff
@@ -1,13 +1,105 @@
...
+* Add link to Open SDG to footer (#411)
+* Configurable footer menu links (#414)
+* Escape variables in alt attributes (#415)
+* Use indicator_available instead of graph_title for the indicator page H2 (#422)
+* Reporting status for any metadata field (#424)
+* Switch to Lunr for search functionality (#427)
+* Metadata field translation keys (#428)
+* Show post excerpts on news page (#431)
+* Sass refactor (#439)
```
SavvasStephanides, line 15: "Refactoring of Sass style sheets" might be more descriptive

docs/changelog.md
```diff
@@ -1,13 +1,105 @@
...
+* Ensure that indicator data always has a value (#441)
+* Set background to white for pages and footer (#448)
+* General javascript cleanup (#451)
```
SavvasStephanides, line 18: Maybe "refactoring" might be more suitable?

docs/changelog.md
```diff
@@ -1,13 +1,105 @@
...
+* Add box shadows to goal tiles (#452)
+* Underline menu items when hovered (#454)
+* Move search box above the top navigation (#455)
+* Keep searchbox the same size when in focus (#456)
+* Multilingual news/posts (#457)
+* Make bars striped when necessary (#459)
```
SavvasStephanides, line 24: Could you briefly expand on "when necessary"? Something like "Make the Xth bar on barcharts striped" (or similar, I don't exactly remember the conditions)

docs/changelog.md
brockfanning: Makes sense.

docs/changelog.md
```diff
@@ -1,13 +1,105 @@
...
+* Add an introduction banner to the frontpage (#486, #545, #565)
+* Set aria-label on search button (#487)
+* Bulk download link beneath the homepage grid (#489)
+* Black text for tab links (#490)
+* Set aria-disabled on 'Clear selections' button (#491)
+* Chart refactor for limits and stacked bars (#497)
+* Fit total number of indicators on one line (#502)
+* Also have overall total in box (#505)
+* Remove goal line from indicator banner (#506)
+* Remove a lot of unused images (#509)
```
SavvasStephanides, line 41: I think "Remove unused images" would suffice here.

docs/changelog.md
brockfanning: Makes sense.

docs/changelog.md
brockfanning: Actually as I recall it really was just a light cleanup - not too much refactoring. Mostly removing unnecessary comments, unused variables, and stuff like that.

SavvasStephanides: Fair enough

docs/changelog.md
brockfanning: Sure, will do. I'll try: "Make bars striped after unique colors are exhausted"

brockfanning:
@SavvasStephanides Ok, just pushed up those updates.

PR #411, opened 2019-12-12
Add link to Open SDG to footer

LucyGwilliamAdmin:
image

### DIFF
--- a/_includes/footer.html
+++ b/_includes/footer.html
@@ -4,6 +4,7 @@
       <div class="col-xs-12">
         <div id="footerLinks">
           <ul>
+            <li>Built using <a href="https://open-sdg.readthedocs.io/en/latest/">Open SDG</a></li>
             <li><a href="mailto:{{site.email_contacts.suggestions}}">{{ page.t.menu.contact_us }}</a></li>
             {% if site.twitter %}
             <li><a class="twitter-follow-button" href="https://twitter.com/{{ site.twitter }}" data-show-count="false">{{ page.t.general.twitter }}: @{{ site.twitter }}</a></li>
--- a/_includes/footer.html
+++ b/_includes/footer.html
@@ -4,7 +4,7 @@
       <div class="col-xs-12">
         <div id="footerLinks">
           <ul>
-            <li>Built using <a href="https://open-sdg.readthedocs.io/en/latest/">Open SDG</a></li>
+            <li><a href="https://open-sdg.readthedocs.io/en/latest/">{{ page.t.general.built_using_open_sdg }}</a></li>
             <li><a href="mailto:{{site.email_contacts.suggestions}}">{{ page.t.menu.contact_us }}</a></li>
             {% if site.twitter %}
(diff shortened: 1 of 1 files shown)


brockfanning:
I love the idea. But we may need to think about translations. I wonder - does "Open SDG" need to be translated? Eg, would it be OK for "Open SDG" to appear in the midst of Chinese characters, for example?

LucyGwilliamAdmin:
@brockfanning Thanks for taking a look

Just one example I could find:
https://www.adidas.com.cn/

They seem to leave adidas in footer amongst Chinese characters

LucyGwilliamAdmin:
Also, I'm not sure how this would work with translations, since `Open SDG` is a link which will move to different places of the sentence depending on the language

LucyGwilliamAdmin:
Unless we just made the whole thing a link?

brockfanning:
@LucyGwilliamAdmin Those concerns are part of the reason I was wondering about the Chinese question. In any case, we're going to have make some translations. One option is to leave a "percent variable" in the translation, like "Built with %platform". As an example, the Spanish translation might be "Construido con %platform". Then, in Open SDG, we do something similar to what we do [here](https://github.com/open-sdg/open-sdg/blob/master/_includes/components/dev-disclaimer.html#L4). Like maybe:
```
{{ page.t.general.built_with_platform | replace_first: '%platform', 'Open SDG' }}
```

But if we're going to go forward with #414 then we wouldn't be able to do anything that complicated. In that case, we'd need to translate the whole thing ("Built with Open SDG") together.

The reason I wonder about this is that so far we haven't put anything in SDG Translations that's unquestionably specific to Open SDG. I would not be opposed to that here, but would recommend that we add it as a dedicated "translation group". So its translation key might be "opensdg.built_with_opensdg".

Hope that all makes sense, let me know if not.

SavvasStephanides:
I don't think "OpenSDG" needs to be translated, since it's a "brand name" (not really a brand but you get my point). "Construido con OpenSDG" for example. My only concern is with Arabic (with switching between left to right and right to left writing).

LucyGwilliamAdmin:
@brockfanning 
Think we should translate whole thing together i.e Built with Open SDG as I think we've all agreed to go ahead with #414 

I have created some translations in sdg-translations but currently these are in general. Happy to move them to a new folder as you suggested above but I know we also discussed having a new repo open-sdg-translations so if we're going to go ahead with that, I will hold off.

Agree with @SavvasStephanides - need to look in to how the translation would work with Arabic

brockfanning:
@LucyGwilliamAdmin I'll be working on the open-sdg-translations repo in the next couple of days, and I'll post an update here when I know more.

brockfanning:
@LucyGwilliamAdmin We can plan on using the key as you've already done here. I've done some quick Google Translations in a new repo [here](https://github.com/open-sdg/translations-opensdg). We can plan on making this new repo a requirement for version 1.0.0.

Can you update the PR? After the recent change to the footer there are conflicting changes.

LucyGwilliamAdmin:
@brockfanning sounds good -have updated the PR

Issue #412, opened 2019-12-12
Footer menu configuration


brockfanning:
Right now we've got a configurable top menu, but hardcoded options (like Twitter, Facebook, etc) for the footer menu. I wonder if we could make the footer menu configurable just like the top menu.

brockfanning:
Fixed by #414

PR #414, opened 2019-12-16
Configurable footer menu links

brockfanning:
This is an attempt at #412.

The upside of this approach is that we don't need specific configuration settings for Twitter and Facebook. So along those lines, those configuration items are removed from the documentation. This also makes it easier to add additional footer menu items.

There is a downside - or at least a change of behavior: The Twitter and Facebook links can't include the account names. For example, instead of "Twitter: @MyTwitterAccout" it has to be just "Twitter". This is because the labels of the menu items have to be controlled by a single "translation key".

Finally this also includes a fix for the main menu as well. Apparently there is a problem with the main menu when the site does not have a "baseurl" - this would only have been visible on a production site, so maybe that is why nobody has mentioned it yet. The problem is that if the "path" of the menu item includes a preceding slash, the href of the menu item ends up with 2 slashes, which can break the link completely. So the fix included here is to remove any preceding slashes.

If we decide not to use this PR, we should still implement the fix mentioned above (in a separate PR).

### DIFF
--- a/docs/configuration.md
+++ b/docs/configuration.md
@@ -43,6 +43,26 @@ This **required** setting should be either `staging` or `production`. Certain fe
 environment: staging
 ```
 
+### footer_menu
+
+This **required** setting controls the footer menu for the platform. It should contain a list of menu items, each containing a `path` and a [translation key](https://open-sdg.readthedocs.io/en/latest/translation/).
+
+The following example provides a footer menu matching older versions of Open SDG, which included options for social media and email contacts.
+
+```nohighlight
+footer_menu:
+  - path: mailto:my-email-address@example.com
+    translation_key: menu.contact-us
+  - path: https://twitter.com/MyTwitterAccount
+    translation_key: general.twitter
+  - path: https://facebook.com/MyFacebookAccount
+    translation_key: general.facebook
+  - path: /faq
+    translation_key: menu.faq
+  - path: /cookies
+    translation_key: menu.cookies
+```
+
 ### goal_image_base
 
 This **required** setting controls the base URL for downloading the imagery for the goals (PNG files). The platform will use this as a base, and complete the URLs (behind the scenes) by adding a language and number. For example, if you set this to `https://example.com`, then the platform will try to download the Spanish image for Goal 4 at: `https://example.com/en/4.png`.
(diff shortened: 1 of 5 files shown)


brockfanning:
@LucyGwilliamAdmin @SavvasStephanides I was wondering about your thoughts on this one. Apart from the code review, do you think this is a good/bad idea, given the upsides/downsides mentioned above?

LucyGwilliamAdmin:
I do like the idea of configurable footer links as makes it easier for countries to add their own.

I'm not sure about not having the username in the social media links - could the translation keys be overwritten locally to include a username?

SavvasStephanides:
I like the idea of a configurable footer menu, it keeps it consistent with the configurable header.

> There is a downside - or at least a change of behavior: The Twitter and Facebook links can't include the account names. For example, instead of "Twitter: @MyTwitterAccout" it has to be just "Twitter". This is because the labels of the menu items have to be controlled by a single "translation key".

I'm not sure why this is the case. Is it because of the "@" symbol? Either way, "Twitter" should be more than enough for the link label. It is common practice to not mention the Twitter/Facebook username when linking to these services.

brockfanning:
@LucyGwilliamAdmin @SavvasStephanides It sounds like you both like the idea, so I think the next step would be a manual test and/or review of the code? This does qualify as a breaking change, I think, but I think it would be fine to make part of version 1.0.0, as it's an easy update to make.

To your questions:
* Could the translation keys be overwritten locally to include a username?
    Yes, that would work. They could also use a custom translation key, like `custom.twitter_menu_item` or something.
* Is it because of the "@" symbol?
    No it's because with the configurable menu approach, the "translation_key" of the menu items can either be a full string (like "Foo") or an actual translation key (like "custom.foo"), but it can't be a combination of both. For example, it can't be "Foo: custom.foo". So it would not be possible to have the Twitter account name combined with the translation key "general.twitter", in the menu item. The only way is to translate the whole thing together, like Lucy described.

LucyGwilliamAdmin:
I've actually since seen your comments about `\` being there or not so need to answer on that part. I have since managed to build a feature branch of the master branch so thought I would be able to test this but seem to be getting this error:

https://travis-ci.org/ONSdigital/sdg-indicators/builds/628749910

docs/configuration.md
```diff
@@ -43,6 +43,26 @@ This **required** setting should be either `staging` or `production`. Certain fe
...
+### footer_menu
+
+This **required** setting controls the footer menu for the platform. It should contain a list of menu items, each containing a `path` and a [translation key](https://open-sdg.readthedocs.io/en/latest/translation/).
+
+The following example provides a footer menu matching older versions of Open SDG, which included options for social media and email contacts.
+
+```nohighlight
+footer_menu:
+  - path: mailto:my-email-address@example.com
+    translation_key: menu.contact-us
```
LucyGwilliamAdmin: This should be `translation_key: menu.contact_us`

brockfanning:
@LucyGwilliamAdmin Thanks for catching that - just made the change.

PR #415, opened 2019-12-16
Escape variables in alt attributes

### LINKED ISSUE #413: Escape all content inside HTML attributes
There is some variable content inside HTML attributes that is not being escaped. This has the potential for invalid HTML. An example is [here](https://github.com/open-sdg/open-sdg/blob/master/_layouts/goal.html#L8). We ought to fix that and any others like it.

brockfanning:
Fixes #413 

This may not be comprehensive but does fix some problems.


brockfanning:
@LucyGwilliamAdmin @SavvasStephanides Here's an example of the issue that this should fix: https://app.circleci.com/jobs/github/open-sdg/open-sdg/1190 (see ```htmlproofer 3.15.0 | Error:  invalid attribute name `zéro"'```)

Issue #278, opened 2019-08-09
Text rendering outside metadata table


LucyGwilliamAdmin:
Darri (Iceland): While browsing the website I have noticed a rendering anomaly in some of the metadata tables for SDG indicators (see. e.g. https://sustainabledevelopment-uk.github.io/16-1-1/). In some cases the text in the tables displays outside the table limits.

![image](https://user-images.githubusercontent.com/43404358/62768595-152ef780-ba8f-11e9-8cf5-aae05014bf77.png)

brockfanning:
Confirmed - I see the same thing Chrome (though interestingly Firefox doesn't have the problem).

brockfanning:
Estimate: This should not be that hard, but unfortunately I'm not seeing an obvious cause. I'd estimate about 5-10 hours due to more investigation. This may involve `table-layout: fixed` in some way.

brockfanning:
@LucyGwilliamAdmin After findings in #421 do you think we should close this? Or do we need any documentation added?

LucyGwilliamAdmin:
@brockfanning I liked your suggestion of having a warning in the metadata documentation about really long words e.g. raw URLs but other than that I don't think this requires anything else

PR #421, opened 2019-12-23
Stop text from rendering outside of metadata table

LucyGwilliamAdmin:
Fix for #278 

See metadata table [here](https://sustainabledevelopment-uk.github.io/16-1-1/) (broken) compared with [here](http://sdgdev-813006012.eu-west-1.elb.amazonaws.com/metadata-table/16-1-1/)

### DIFF
--- a/_includes/components/metadata.html
+++ b/_includes/components/metadata.html
@@ -1,4 +1,4 @@
-<table class="table table-hover">
+<table class="table table-hover" id="metadata-table">
   {% for indicator_metadata in site.data.schema %}
     {%- if page.t.metadata_fields[indicator_metadata.name] -%}
       {%- assign metadata_label = page.t.metadata_fields[indicator_metadata.name] -%}
--- a/assets/css/default.scss
+++ b/assets/css/default.scss
@@ -341,6 +341,10 @@ header {
   }
 }
 
+#metadata-table {
+  table-layout: fixed;
+}
+
 @media only screen and (max-width: 768px) {
 
   header .container {


brockfanning:
@LucyGwilliamAdmin The only problem with "table-layout: fixed" is that it makes the field label column wider than it needs to be.

I think I've figured this out though - it's not totally a problem with the code, it's actually the content of this particular indicator. The default table-layout is "auto", which means that the table cell widths will be as wide as the "widest unbreakable content in the cells" (from [here](https://www.w3schools.com/cssref/pr_tab_table-layout.asp)). This means that if there is any really long word, anywhere in any of the cells of the column, the whole column (and all its cells) will be that wide.

In the case of this indicator, it's that long URL in "Other Information". I think if that could be turned into a link (with markdown) it could be shorter, and that would fix the problem.

LucyGwilliamAdmin:
Ahh ok - great thanks!

LucyGwilliamAdmin:
@brockfanning turns out that this by itself doesn't fix it anyway - it seems that coincidentally that the URL for 16.1.1 has been created using markdown but is waiting to be made live so if it hadn't the text would still be going over like [here](http://sdgdev-813006012.eu-west-1.elb.amazonaws.com/metadata-table/3-4-2/).

Agree best way to fix is to just make sure links use markdown. Other than links, there shouldn't be anything else that extends the column right? If so, does anything actually need to be done for this?

brockfanning:
@LucyGwilliamAdmin The only think I can think of would be a note in the documentation warning against really long words, like raw URLs.

PR #422, opened 2020-01-03
Use indicator_available instead of graph_title for the indicator page H2

### LINKED ISSUE #337: H2 heading should depend on indicator_available
We use a metadata field called `graph_title`, for example [here](https://github.com/open-sdg/open-sdg/blob/master/_layouts/indicator.html#L42) and [here](https://github.com/open-sdg/open-sdg/blob/master/_layouts/indicator.html#L59). Is this still needed? Can't it just be the indicator name?

brockfanning:
Fixes #337 

This changes the H2 on the indicator pages to `indicator_available` instead of `graph_title`.

This could be considered a breaking change. The most probable result for most implementations will be that the H2 will disappear, since there is nothing in the `indicator_available` field. On the bright side, this isn't so bad, because that H2 was just duplicating the graph title anyway.

### DIFF
--- a/_layouts/indicator.html
+++ b/_layouts/indicator.html
@@ -39,11 +39,13 @@
     </div>
   </div>
 
-  <div class="row">
+  {% if page.indicator.indicator_available %}
+  <div class="row metadata-available-indicator">
     <div class="col-xs-12">
-      <h2>{{ page.indicator.graph_title }}</h2>
+      <h2>{{ page.indicator.indicator_available }}</h2>
     </div>
   </div>
+  {% endif %}
 
   {% assign sources_list = '' | split: ',' %}
   {% for i in (1..12) %}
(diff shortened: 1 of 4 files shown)


SavvasStephanides:
@brockfanning Can you please explain what `indicator_available` means? From the name it kind of looks like it's a boolean variable.

LucyGwilliamAdmin:
@SavvasStephanides `indicator_available` is a free text field where countries can specify what indicator is available if the data doesn't match exactly what the UN asks for.

brockfanning:
I'm all for a metadata field name revamp, though if we do that I think we may want to look for a standard schema outside of Open SDG to adopt. I believe the SDMX working group is getting close to a standard proposal, so it's probably worth it to wait a bit and see. When we do change fields it could be somewhat disruptive to existing implementations, so we have to be careful.

I'll go ahead and merge this. I'm also tagging this "Change record" to indicate that we need to document this change in a non-technical way.

PR #427, opened 2020-01-11
Switch to Lunr for search functionality

### LINKED ISSUE #226: Search functionality refactor using Lunr.js?
This may be a good opportunity to standardize around a third-party open-source library, remove custom code from the platform, and improve functionality: use [Lunr.js](https://lunrjs.com/) for the search functionality. Is it feasible to refactor the search functionality so that it uses Lunr.js instead of custom javascript?

brockfanning:
Fixes #226 

This is first draft of a sitewide search using Lunr, replacing the indicator-only custom search we already have. Lots of testing would be needed and probably some refactoring. It also needs a general review to decide if it's an improvement, user-experience-wise.

Update: since we decided to pursue this further, I've extended it a bit. It now includes two optional configuration settings: `search_index_extra_fields`, which can be used to add additional indicator metadata fields to the search index, and `search_index_boost`, which can be used to fine-tune the relevance of the search results, by giving a 'boost' to any of the indexed fields.

This depends on the latest version of jekyll-open-sdg-plugins: 1.0.0-rc17


brockfanning:
@LucyGwilliamAdmin @SavvasStephanides @AnnCorp I meant to bring this up on our call. This is a quick try of Lunr, a client-side search solution. Here are some screenshots for review:

A search for "poverty", showing goals and indicators:

lunr-search-example1


A search for "guidance", showing only pages:

lunr-search-example2

brockfanning:
Here's a screenshot of the "did you mean" functionality:

lunr-search-example3

brockfanning:
Still needs some work:
* translations
* support custom "boost" scores
* better automated tests

brockfanning:
This is ready for review now.

_includes/assets/js/search.js
```diff
@@ -1,115 +1,118 @@
...
-    for(var goalLoop = 0; goalLoop < data.length; goalLoop++) {
-      for(var indicatorLoop = 0; indicatorLoop < data[goalLoop].goal.indicators.length; indicatorLoop++) {
-        var currentIndicator = data[goalLoop].goal.indicators[indicatorLoop];
-        currentIndicator.goalId = data[goalLoop].goal.id;
-        currentIndicator.goalTitle = data[goalLoop].goal.title;
-        that.indicatorData.push(currentIndicator);
-      }
+var indicatorSearch = function() {
+  // Helper function to make a search query "fuzzier", using the ~ syntax.
+  var getFuzzierQuery = function(query, fuzziness) {
```
SavvasStephanides: Consider renaming `fuzziness` to `amountOfFuzziness` to make it clearer that's it's a positive integer. Of course "fuzziness" as a concept is a bit confusing at first glance until you read the Lunr docs which makes things pretty clear.

_includes/assets/js/search.js
```diff
@@ -1,115 +1,118 @@
...
+    var opts = {}
+    if (opensdg.searchIndexBoost[field]) {
+      opts['boost'] = intval(opensdg.searchIndexBoost[field])
     }
-  };
+    return opts
+  }
+  var urlParams = new URLSearchParams(window.location.search);
+  var searchTerms = urlParams.get('q');
+  if (searchTerms) {
```
SavvasStephanides: Be careful with using equality like this one in Javascript. `if(searchTerms)` will return false if the query is `undefined` (as you'd expect here) but it will also return false if the search term is a string containing `"0"`. It'd be better if you were more explicit with your expected search term. (Search for "javascript falsy" for more info)

_includes/assets/js/search.js
```diff
@@ -1,115 +1,118 @@
...
-          results.push({
-            goalId: result.goalId,
-            goalTitle: result.goalTitle,
-            indicators: [indicator]
-          });
-        } else {
-          goal.indicators.push(indicator);
+    // Perform the search.
+    var results = searchIndex.search(searchTermsToUse);
+    var didYouMean = false;
```
SavvasStephanides: Rename this variable so that it's clear that `didYouMean` is a boolean value. Not very clear what's the purpose of this variable.

_includes/assets/js/search.js
```diff
@@ -1,115 +1,118 @@
...
+
+    // If we didn't find anything, get progressively "fuzzier" to look for
+    // "did you mean?" options.
+    if (!results.length) {
```
SavvasStephanides: Same as above, regarding falsy equality

_includes/assets/js/search.js
```diff
@@ -1,115 +1,118 @@
...
+      var fuzziness;
+      for (fuzziness = 1; fuzziness < 5; fuzziness++) {
```
SavvasStephanides: If you start your for loop with `for (var fuzziness = 1...`, you don't need to declare `var fuzziness` above.

_includes/assets/js/search.js
```diff
@@ -1,115 +1,118 @@
...
+        var fuzzierQuery = getFuzzierQuery(searchTermsToUse, fuzziness);
+        var alternateResults = searchIndex.search(fuzzierQuery);
+        if (alternateResults.length) {
```
SavvasStephanides: Same as above, regarding falsy equality

brockfanning:
Thanks for the detail @SavvasStephanides! I believe I got everything - how does it look now?

_includes/assets/js/search.js
```diff
@@ -62,19 +38,20 @@ var indicatorSearch = function() {
     }
     // Perform the search.
     var results = searchIndex.search(searchTermsToUse);
-    var didYouMean = false;
+    // Array of alternative search terms to suggest to the user, in the case
+    // where no results were found.
+    var alternateSearchTerms = [];
```
SavvasStephanides: You mean "alternative"? 🙂

brockfanning: That's better yeah - just changed it.

PR #430, opened 2020-01-12
Documentation updates to standardize around translations in data repo

brockfanning:


LucyGwilliamAdmin:
@brockfanning documentation looks good - I know you sent me a document with all of the fields that need to be translated for the Open SDG if adding a new language - is it worth adding a link to this in the documentation?

brockfanning:
@LucyGwilliamAdmin The deprecated sdg-translations project had (has) custom code that generates a CSV file for easier translating. The new replacement repositories (translations-un-sdg and translations-open-sdg) don't have that feature. So for now I've added links to the folders of YAML files that will need to be translated for a new language.

Longer term, I want to add those repositories to a free CAT platform like Transifex or Weblate, which will make it even easier for new languages to be added. But for now I think the links to those folders is sufficient.

PR #431, opened 2020-01-13
Show post excerpts on news page

### LINKED ISSUE #425: Excerpt missing from News page
Even though there is a excerpt field in the markdown files for posts, the excerpt doesn't actually show anywhere

brockfanning:
Fixes #425 

@LucyGwilliamAdmin I haven't tested this - this a quick shot in the dark, in case it works.

### DIFF
--- a/_layouts/news.html
+++ b/_layouts/news.html
@@ -9,12 +9,10 @@ layout: default
       <article>
         <h5><a href="{{site.baseurl}}{{ post.url }}">{{ post.title }}</a></h5>
         <p><time datetime="{{post.date|date:"%F"}}">{{post.date|date:"%b %d, %Y"}}</time></p>
-        {% if post.excerpt %}
         <div class="post-excerpt">
           {{ post.excerpt | markdownify }}
         </div>
         <p>{{post.categories | category_links}}</p>
-        {% endif %}
       </article>
       {% endfor %}
     </div>
--- a/_layouts/news.html
+++ b/_layouts/news.html
@@ -3,13 +3,16 @@ layout: default
 ---
 <div id="main-content" class="container newslist">
   <div class="row">
-    <div class="col-md-9 push-md-3">  
+    <div class="col-md-9 push-md-3">
       {{content}}
       {% for post in site.posts %}
       <article>
         <p>{{post.categories | category_links}}</p>
         <h5><a href="{{site.baseurl}}{{ post.url }}">{{ post.title }}</a></h5>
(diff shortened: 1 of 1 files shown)


LucyGwilliamAdmin:
@brockfanning yes, will test now.

I think another change the news page may benefit from is one we've implemented in the UK:
https://github.com/ONSdigital/sdg-indicators/blob/develop/_layouts/news.html

It just moves the categories from the top of the post on the News page to the bottom. When it was at the top, it looked as if the posts are grouped into categories whereas they're actually in date order.

LucyGwilliamAdmin:
@brockfanning have tested the excerpt part and looks good.
The only thing I can see that we need to document is that if no excerpt is set, the first paragraph (?) is automatically set as excerpt - see this [feature branch](http://sdgdev-813006012.eu-west-1.elb.amazonaws.com/open-sdg-news-page-set-excerpt/updates/) where excerpt was only set for International work.

brockfanning:
@LucyGwilliamAdmin The categories move sounds good - I just added that.

Jekyll actually has its own documentation of the excerpt functionality: https://jekyllrb.com/docs/posts/#post-excerpts

I don't believe we have any post-specific documentation - do you think we should add that?

LucyGwilliamAdmin:
@brockfanning testing the latest changes - do you think it would make sense to have the categories between the date and the excerpt?

This is how it looks at the moment?
image

LucyGwilliamAdmin:
@brockfanning re. documentation - I think we could add something (even if we just give basic overview and simplify what is in the Jekyll guidance)

brockfanning:
Yeah, it's not very clear that those are even categories. It seems like they ought to be links to other listing pages specifically for that category. Or maybe to have "Categories: " as a label next to them? Or maybe just removed altogether?

It also occurs to me that the news/post functionality is not set up to work with multilingual. For example, if a country has a post translated into two languages, I think it would show up twice on the News page. Also, different languages show dates in different ways, so we'd want to figure out how to properly display dates according to the current language.

If you agree, should we turn this into a general news/post revamp? Or make separate tickets for that?

LucyGwilliamAdmin:
@brockfanning I agree that the news/post pages could do with a revamp.
@AnnCorp I know we briefly discussed the news page the other day - what are your thoughts?

brockfanning:
@LucyGwilliamAdmin I think we discussed going ahead with this one. I created some other tickets for followup work:
* #435 
* #434 
* #433

PR #432, opened 2020-01-15
Docs updates regarding reporting status options and data sources.

brockfanning:


brockfanning:
@LucyGwilliamAdmin @SavvasStephanides I don't think we have any documentation about sdg-build and the ways to use it. So I wrote this up pretty quick. Any feedback is welcome.

I also included an unrelated bit about color-coding the reporting status options.

PR #436, opened 2020-01-16
Docs changes

LucyGwilliamAdmin:


brockfanning:
@LucyGwilliamAdmin Looks great. The only thing I would change is to make the URLs in the list of users into actual links. Mkdocs doesn't automatically convert URLs into links, but you can do this easily with angle brackets:
```
|Armenia|<https://armstat.github.io/sdg-site-armenia/>|
|Germany|<https://sustainabledevelopment-germany.github.io/>|
etc...
```

LucyGwilliamAdmin:
@brockfanning great thanks!

PR #439, opened 2020-01-20
Sass refactor

brockfanning:
This PR has 2 goals:
1. Move to a more standard way of having Sass in a Jekyll theme
2. Split up the single Sass file into many smaller files, for easier maintenance and to allow targetted overrides by users

The split-up of the Sass was done fairly quickly, and may have some mistakes in it. Some of the Sass I'm unfamiliar with, so am not sure offhand what part of the platform it's for. However, in theory, even if I put something in the wrong file, everything should still behave as before.

The only danger here is that the order of some CSS may change, which could have a visual effect. So reviewing this should include a visual once-over of the entire platform, if possible.


brockfanning:
@SavvasStephanides A quick way to review this is probably to browse the _sass folder [here](https://github.com/brockfanning/open-sdg/tree/sass-refactor/_sass). The main question I'm wondering about: Does the way I split up the Sass make sense? (base, components, variables, etc.).

@LucyGwilliamAdmin Do you think you could test this branch with the UK site? This could use a visual check of the entire platform. I didn't change any CSS, but definitely **reordered** some. In particular I worry that some of the high-contrast stuff might get messed up.

_sass/components/_breadcrumbs.scss
```diff
@@ -0,0 +1,11 @@
+.breadcrumb {
+  margin-bottom: 15px;
+  a {
+    color: darken(#337ab7, 10%);
+    // text-decoration: underline;
```
SavvasStephanides: You can remove this since it's not in use. If you need to bring it back you can always look back at the git history

_sass/components/_high_contrast.scss
```diff
@@ -0,0 +1,325 @@
+#accessibility-nav {
```
SavvasStephanides: This file can be split up since it seems to include a lot of layout/component styling, which you have already split up for normal contrast in this ticket

_sass/components/_header.scss
```diff
@@ -0,0 +1,297 @@
...
+  height: 34px;
+  }
+
+.table {
+  .table-value {
+    text-align: right;
+  }
+}
+
+@media only screen and (max-width: 768px) {
```
SavvasStephanides: To clean this up even further, you could split your files to reflect different responsive layouts.

_sass/components/_breadcrumbs.scss
brockfanning: Sounds good, I removed it.

_sass/components/_high_contrast.scss
brockfanning: I agree - I split most things out into separate files. It may cause a few specificity problems but we can shuffle the order of the `@include` statements to fix that.

_sass/components/_header.scss
brockfanning: Do you mean to have, for example, a `_header__desktop.scss` file, `_header__mobile.scss`, etc? I agree this would clean things up. But I admit I'm a little hesitant to launch into that. I don't have a good handle on all the media queries, and whether we're consistent with "mobile first" or "desktop first" in general.

SavvasStephanides: Fair enough. We can leave it as is if it's too much work for now. We can back to it on a later date if needed.

SavvasStephanides:
Thanks @brockfanning, the changes look good! Now we're left with checking  if things don't break in the UK site. I'll do a quick check.

SavvasStephanides:
I've done a quick check and it seems fine. Looks OK to me.

PR #447, opened 2020-01-27
Take user straight to indicator page if search by indicator ID

SavvasStephanides:
If a user searches for an indicator by ID (eg. `1.1.1`), the site will take the user straight to the indicator page.

UK feature branch: http://sdgdev-813006012.eu-west-1.elb.amazonaws.com/search_by_indicator/

### DIFF
--- a/_includes/assets/js/search.js
+++ b/_includes/assets/js/search.js
@@ -22,7 +22,7 @@ var indicatorSearch = function(inputElement, indicatorDataStore) {
 
     if(e.keyCode === 13 && searchValue.length) {
       if(searchQueryMatchesIndicatorPattern){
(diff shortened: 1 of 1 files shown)


LucyGwilliamAdmin:
Looks good but think we need an error page when an indicator code doesn't exist.

Also would be good if it included where codes have dashes instead of dots e.g. 1-1-1

brockfanning:
@SavvasStephanides @LucyGwilliamAdmin In the interest of avoiding double-work, could we hold off on this? I thought we'd decided I should continue working on #427, and that changes the same file. We can discuss more next call.

SavvasStephanides:
@brockfanning yes sounds good.

PR #449, opened 2020-01-27
Remove custom CSS stuff from the test config

brockfanning:
This will be needed to get tests to pass again.


brockfanning:
@LucyGwilliamAdmin @SavvasStephanides Please review/approve if you agree - this is need ASAP to get tests to pass. (The reason this is necessary is that I updated the starter repos to match the new sass workflow.)

SavvasStephanides:
Yes I think this is absolutely necessary

PR #450, opened 2020-01-29
Reference the single-repository approach in the quick start

### LINKED ISSUE #442: Document the support for a single repository
We've recently added support for using a single repository, instead of the "site repo" and "data repo" approach. But it's undocumented. We need to decide which of these options we like:
1. Have the quick start detail the double-repository approach (as it is now) but link to a separate documentation page that details the single-repository approach
2. Have the quick start detail the single-repository approach but link to a separate documentation page that details the double-repository approach
3. Have the quick start detail both approaches.

I'm partial to approach-2, since I feel like the quick start that be as quick and simple as possible. But some may consider approach-1 enough of a best practice that it should be encouraged, despite the complexity. I think approach-3 would make the quick start appear too daunting.

For a sneak preview of what the documentation would be like for the single-repository approach, see the [single-repository starter repo here](https://github.com/open-sdg/open-sdg-simple-starter) and go through the README.

brockfanning:
Fixes #442

### DIFF
--- a/docs/faq.md
+++ b/docs/faq.md
@@ -12,7 +12,7 @@ instructions on the quickest way to get a copy of the open-sdg platform up and r
 
 ## How long would it take to set up a copy of Open SDG
 
-With the right skills, it should take less than a day to set up a standard copy, ready for data to be imported. 
+With the right skills, it should take less than a day to set up a standard copy, ready for data to be imported.
 
 ## Can the Open SDG platform be customised?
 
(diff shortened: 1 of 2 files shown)


brockfanning:
Still todo: Add an item in the FAQ about the difference between single and double repositories.

brockfanning:
@AnnCorp I added a FAQ section to this PR. What do you think? Is it too long?

AnnCorp:
@brockfanning I think this text is great thanks - lets just tweak if we get any further questions on this (eg could add country examples if anyone goes down the single repo route)

PR #451, opened 2020-01-29
General javascript cleanup

### LINKED ISSUE #196: Code cleanup and refactor
The codebase could use a cleanup. Here are some examples of things we should try:
* Remove all unused variables, functions, etc
* Remove commented-out code
* Remove unused CSS
* Remove use of "important" in CSS (unless necessary)

brockfanning:
Fixes #196 

We've already done a Sass cleanup. A proper javascript refactor would probably involve splitting things into multiple files, but would likely be a bigger job. This is some "low hanging fruit" in that it just removes unnecessary comments and unused variables.

### DIFF
--- a/_includes/assets/js/indicatorModel.js
+++ b/_includes/assets/js/indicatorModel.js
@@ -351,16 +351,13 @@ var indicatorModel = function (options) {
         unitsChangeSeries: false
       }),
       fields = this.selectedFields,
-      selectedFieldTypes = _.pluck(fields, 'field'),
       datasets = [],
       that = this,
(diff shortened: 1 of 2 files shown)


SavvasStephanides:
Good start Brock. Although yes it will need a lot of refactoring in general. We can discuss this further in the call, but what I'd suggest is following something called "the Boy Scout rule".  In the "clean code" world, it basically means ["leave your code cleaner than you found it"](https://deviq.com/boy-scout-rule/). This way we can make small refactoring changes everytime we work on some code.

PR #459, opened 2020-02-01
Make bars striped when necessary

### LINKED ISSUE #443: Dashed colours not working in bar charts
When bar chart is being used, the legend shows dashed colours when a certain number of disaggregations are selected but the corresponding bars don't show dashed colours e.g. [UK 4.3.1](https://sustainabledevelopment-uk.github.io/4-3-1/)

image

brockfanning:
Fixes #443 

This uses the [patternomaly](https://github.com/ashiguruma/patternomaly) library to set a striped background on bars when necessary. I did not do any browser testing.

### DIFF
--- a/_includes/assets/js/indicatorModel.js
+++ b/_includes/assets/js/indicatorModel.js
@@ -377,15 +377,6 @@ var indicatorModel = function (options) {
           }
         }
       },
-      getBackgroundColor = function(datasetIndex) {
-        return '#' + getColor(datasetIndex);
-      },
-      getBackgroundPattern = function(color) {
-        if (window.pattern && typeof window.pattern.draw === 'function') {
-          return window.pattern.draw('diagonal', color);
-        }
-        return color;
-      },
       getBackground = function(datasetIndex) {
 
         var color = getBackgroundColor(datasetIndex);
@@ -399,6 +390,15 @@ var indicatorModel = function (options) {
           color = getBackgroundPattern(color);
         }
 
+        return color;
+      },
+      getBackgroundColor = function(datasetIndex) {
+        return '#' + getColor(datasetIndex);
+      },
+      getBackgroundPattern = function(color) {
+        if (window.pattern && typeof window.pattern.draw === 'function') {
+          return window.pattern.draw('diagonal', color);
+        }
         return color;
       },
       getBorderDash = function(datasetIndex) {
--- a/_includes/assets/js/indicatorModel.js
+++ b/_includes/assets/js/indicatorModel.js
@@ -377,15 +377,21 @@ var indicatorModel = function (options) {
           }
         }
       },
-      getBackgroundColor = function(datasetIndex) {
+      getBackground = function(datasetIndex) {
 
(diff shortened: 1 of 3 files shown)


_includes/assets/js/indicatorModel.js
```diff
@@ -377,6 +377,16 @@ var indicatorModel = function (options) {
...
         }
       },
+      getBackgroundColor = function(datasetIndex) {
+
+        var color = '#' + getColor(datasetIndex);
+        // offset if there is no headline data:
+        if(!this.hasHeadline) {
+          datasetIndex += 1;
+        }
+        // Either a solid color, or a pattern via the "patternomaly" library.
```
SavvasStephanides: Your code below is quite well self-documented so you don't need this comment (no need to mention the library used).

Tip: Best for your code to be self-documented instead of relying on comments

_includes/assets/js/indicatorModel.js
```diff
@@ -377,6 +377,16 @@ var indicatorModel = function (options) {
           }
         }
       },
+      getBackgroundColor = function(datasetIndex) {
```
SavvasStephanides: Maybe you could change this to `getBackground()` now that it does a bit more than setting a background _color_ (it sets a pattern as well).

Also, (just a suggestion) consider splitting your function into the part of changing color and the part of changing pattern (solid or diagonal):

```javascript
function getBackground(datasetIndex){
    var color = getBackgroundColor(datasetIndex);
    var background = getBackgroundPattern(datasetIndex, color);
    return background;
}
```

_includes/assets/js/indicatorModel.js
brockfanning: Good advice, thanks. I tweaked it to add some safety code. Your comment made me realize my mentioning of the library was really just a useless attempt to document that it relies on a global variable on the "window" object. So I converted it into an explicit check for that variable.

_includes/assets/js/indicatorModel.js
brockfanning: Good ideas, I made these changes.

brockfanning:
Thanks @SavvasStephanides this is ready for review again.

_includes/assets/js/indicatorModel.js
```diff
@@ -377,6 +377,30 @@ var indicatorModel = function (options) {
...
+        return '#' + getColor(datasetIndex);
+      },
+      getBackgroundPattern = function(color) {
+        if (window.pattern && typeof window.pattern.draw === 'function') {
+          return window.pattern.draw('diagonal', color);
+        }
+        return color;
+      },
+      getBackground = function(datasetIndex) {
```
SavvasStephanides: Much better. One last suggestion to move `getBackground` on top of `getBackgroundColor` and `getBackgroundPattern`. Looks great otherwise!

brockfanning: Cool, that's done.

vmartin8888:
it work fine! thanx

PR #462, opened 2020-02-03
Testing revamp: Data in repo, and Github Actions

brockfanning:
This is a change to keep all the test-specific data and metadata in this repository, rather than relying on the separate open-sdg-data-testing repository.

This also switches from CircleCI to Github Actions. The rationale is that 1) it removes one external dependency and 2) it allows Python support without any trouble.


brockfanning:
@LucyGwilliamAdmin @SavvasStephanides This PR removes the dependency on open-sdg-data-testing, and switches to using Github Actions for our automated tests. I don't think it's possible to test this. I think we should merge this (if one of you approves) and then resolve any remaining issues.

brockfanning:
@SavvasStephanides @LucyGwilliamAdmin Disregard - realized I can turn off the approval requirement to handle this. I'm assuming you'll agree this is worthwhile. :)

PR #465, opened 2020-02-04
Adjust homepage whitespace

SavvasStephanides:
[UK Feature branch](http://sdgdev-813006012.eu-west-1.elb.amazonaws.com/opensdg1-whitespace/)

Done some basic whitespacing on the homepage for this change. Let me know if it looks good to you or if you'd like to see some changes. Feedback more than welcome.

### DIFF
--- a/_layouts/frontpage.html
+++ b/_layouts/frontpage.html
@@ -4,10 +4,11 @@
 
     {% assign country_name = site.country.name | t %}
     {% assign heading_default = page.t.frontpage.heading | replace: '%name', country_name %}
-    <h1>{{ site.frontpage_heading | default: heading_default | t }}</h1>
+    <h1 class="spacer medium">{{ site.frontpage_heading | default: heading_default | t }}</h1>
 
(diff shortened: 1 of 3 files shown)


brockfanning:
@SavvasStephanides Is this not possible with CSS alone, rather than adding extra markup?

SavvasStephanides:
The rationale of using spacers is to make the `html` more modular by adding a `<div>` whose sole purpose is to give whitespace rather than alter the existing elements. But you might be right, it's probably unnecessary. I'll change it.

brockfanning:
@SavvasStephanides I've always been of the opinion that visual stuff should be in CSS, and semantic content should be in HTML. I do like modularity, but would prefer modular CSS - like a class that can be added to something to give it margin-top, for example - over the practice of using spacer divs.

SavvasStephanides:
@brockfanning Thanks for the feedback. Removed the empty `<div>`s and added the classes to existing elements instead

brockfanning:
@SavvasStephanides I think the spacer divs got left in there.

PR #468, opened 2020-02-04
Test only the minimum amount of data/metadata

brockfanning:
The purpose of this is only to speed up the automated testing process. A side-benefit is that it also speeds up the "make serve" local development tool. The downside is that only Goals 1-3 are available on the test site, but I don't think that's necessarily a problem. This PR adjusts all the automated tests so that they work with the smaller set of indicators.

I'll see how long this PR takes to test, and edit this comment to report on how much faster it actually is.

EDIT:
* For automated PR tests, this reduces the time from 7 minutes to 5 (saves 2 minutes)
* For local development, this reduces the 'make serve' from 115 seconds to 40 seconds (saves 75 seconds)

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -12,6 +12,9 @@ build: clean
 	# Clone the starter repositories.
 	git clone https://github.com/open-sdg/open-sdg-site-starter.git site-starter
 	git clone https://github.com/open-sdg/open-sdg-data-starter.git data-starter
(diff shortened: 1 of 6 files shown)


LucyGwilliamAdmin:
@brockfanning I was reviewing this and noticed the conflict needed resolving so I did it but forgot to name commit sorry!

PR #474, opened 2020-02-07
Change the styling of the top banner on goal and indicator pages

SavvasStephanides:
This change implements the following changes on the top banner:

* Set the background color to white
* Change the text on the banner to black
* Change the link text to dark grey (`#ddd`)
* Add some whitespace on the banner to give more prominence

[UK feature branch](http://sdgdev-813006012.eu-west-1.elb.amazonaws.com/opensdg1-topbanner/1/)

### Goal page

image

### Indicator page

image

### DIFF
--- a/_sass/base/_typography.scss
+++ b/_sass/base/_typography.scss
@@ -10,7 +10,7 @@ h6 {
 .heading {
   a {
     background: transparent;
-    color: darkgrey;
+    color: #888;
     display: block;
     &:hover {
       text-decoration: none;
--- a/_sass/base/_typography.scss
+++ b/_sass/base/_typography.scss
@@ -19,7 +19,7 @@ h6 {
   h1,
 	h2,
   h3 {
-    color: white;
+    color: black;
   }
   h1 {
     margin-top: 12px;
--- a/_sass/base/_typography.scss
+++ b/_sass/base/_typography.scss
@@ -10,7 +10,7 @@ h6 {
 .heading {
   a {
     background: transparent;
-    color: white;
+    color: darkgrey;
     display: block;
     &:hover {
       text-decoration: none;
--- a/_sass/components/_goal_tiles.scss
+++ b/_sass/components/_goal_tiles.scss
@@ -2,6 +2,8 @@
   $c: nth($goal-colors, $i);
   $dark-c: darken($c, 20%);
   .goal-#{$i} {
+    padding-top: 30px;
+    padding-bottom: 30px;
     &.goal-indicators {
       div {
         span {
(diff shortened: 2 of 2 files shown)


brockfanning:
@LucyGwilliamAdmin Can you review when you have a chance?

LucyGwilliamAdmin:
This looks good - just wondering if it might be work making the goal icons slightly smaller?

SavvasStephanides:
@LucyGwilliamAdmin Any preference as to how small?

phillipgardiner:
Hello. Managed to get myself on here :). Good work on the banner @SavvasStephanides. As discussed on slack yesterday, could we consider 1) Removing the Goal tag line? Does it actually service a purpose? as the breadcrumbs and icon let you know which goal you are situated within. 2) Resize the icon to about 80% rather than 100%. 3) If the above are done, i'm also thinking if its possible to centre align the text with the icon?

brockfanning:
Personally I'd be fine with removing the goal tag line. @AnnCorp @SavvasStephanides @LucyGwilliamAdmin how do you feel about that?

I'd also be fine with limiting the goal images to 80% instead of 100%.

@phillipgardiner Can you clarify about the centre aligning? Do you mean to change the icon images? Or having the goal icon and indicator title be a single centred paragraph?

Last, @SavvasStephanides do you prefer to do these things in separate issues? I was thinking maybe we should merge this and continue separately, but it's up to you.

SavvasStephanides:
Thank you @philipashlock. Good suggestions.

@brockfanning As we discussed on Slack with @philipashlock, I'm all for removing the goal line from the banner. I don't think it serves any purpose.

Let's push this change and break these suggestions into separate issues.

brockfanning:
I made follow-up issues for the smaller icons and removing the tag line. I'll leave the centre-alignment for someone else to create, as I'm not 100% sure I'm picturing the right thing.

phillipgardiner:
@brockfanning I think I mean 'having the goal icon and indicator title be a single centred paragraph' sorry I'm not 100% sure of the correct language to use get my point across. I've very crudely sketched what I mean in the attached document. If I'm honest it's one of those things that until you see it your not sure if it works (especially given the indicator descriptions vary in length). My main thinking was that currently (when only one line) the text looks very bolted to the top (however when there are 3 lines) it looks much better. Potentially making centre align may work better for all?
[Centre Aligning.docx](https://github.com/open-sdg/open-sdg/files/4185615/Centre.Aligning.docx)

phillipgardiner:
@SavvasStephanides I also believe the text in the banner may look better as a 'h1' rather than a 'h2'. (separate note: not sure why 'h1' is smaller than 'h2'??)

brockfanning:
@phillipgardiner Got it, I'll make a follow-up ticket for the centre alignment.

Maybe add your note about h1/h2 to #478 ? Maybe that can include some refactor so that h1 is actually bigger than h2...

Issue #374, opened 2019-09-26
displaying dissaggregated data on the map


christoph4711:
@brockfanning 
Is it possible to display the dissagregated data in the map?

I have data about housing in lower saxony [here](https://christoph4711.github.io/open-sdg-site-starter/4-4-1/)

On the map currently the floor space in squaremeters of the flats is displayed. In the data I have also the amount of 1-, 2-, and 3-room flats. Is it possible to display e.g. the amount of 1-room or 2-room flats in the map based on the selection in the table/graph view?

brockfanning:
@christoph4711 This would be a great feature. It's tricky though, because the disaggregation drop-downs that we use allow more than one disaggregation to be selected, but maps would only be capable of displaying one at a time.

For example, a line graph can easily show "female" and "male" by having 2 lines. But the choropleth maps can only display "female" or "male", but not both.

So for maps, we need a special widget that allows only one selection (like a radio button). So even before we tackle the technical hurdles, there are some user-interface questions to figure out.

christoph4711:
ok. I understand....

christoph4711:
@brockfanning 
@MoBosse 
may be an interesting repository.....

https://github.com/floledermann/mapmap.js

used [here](https://service.destatis.de/DE/karten/migration_integration_regionen.html):

brockfanning:
After doing a revamp of the mapping system in preparation for Open SDG 1.0.0, a simple drop-down select is actually very easy to implement. But I think plenty of UX discussion is in order. Here is a screencast of what it would look like, with the code we already have. (For some reason in the video the drop-down doesn't expand, but you get the idea.)

![captured (8)](https://user-images.githubusercontent.com/1319083/81083641-d5ccf480-8ec2-11ea-9607-ab6cb89a69cf.gif)

@AnnCorp @phillipgardiner @SavvasStephanides @LucyGwilliamAdmin Just tagging for awareness.

brockfanning:
Summary of notes from #769:
1. When there is a column for the region name which "mirrors" the GeoCode column, a random region appears in the disaggregation drop-down. A way to avoid this is to use "exclude_columns" in the data repository map_layer config, but this is not obvious. Can we automate that?
2. In IE, the disaggregation select stops working after the first choice.

brockfanning:
[...]

Issue #197, opened 2019-06-02
Automated quick-start with Docker


brockfanning:
We may be able to make the quick-start easier with more automation. I'm not sure if we could get it to a one-command setup (the dream!) but we should try to get as close to that as possible.

brockfanning:
Related discussion about Docker in #138

brockfanning:
I wonder if a Docker setup could serve two purposes: local development, and also self-hosting in production? Maybe a bit ambitious...

brockfanning:
@SavvasStephanides I tried it out and it seemed to be working great, but I get an error on the site build. It seems like it's having a problem with the bundle install:

```
# Build the Jekyll site.
cd site-starter && bundle install
Traceback (most recent call last):
	2: from /usr/local/bin/bundle:23:in `<main>'
	1: from /usr/lib/ruby/2.5.0/rubygems.rb:308:in `activate_bin_path'
/usr/lib/ruby/2.5.0/rubygems.rb:289:in `find_spec_for_exe': can't find gem bundler (>= 0.a) with executable bundle (Gem::GemNotFoundException)
Makefile:12: recipe for target 'build' failed
make: *** [build] Error 1
Traceback (most recent call last):
	2: from /usr/local/bin/bundle:23:in `<main>'
	1: from /usr/lib/ruby/2.5.0/rubygems.rb:308:in `activate_bin_path'
/usr/lib/ruby/2.5.0/rubygems.rb:289:in `find_spec_for_exe': can't find gem bundler (>= 0.a) with executable bundle (Gem::GemNotFoundException)
Makefile:45: recipe for target 'serve.docker' failed
make: *** [serve.docker] Error 1
```

Not sure if it's relevant but:
```
brockfanning@brockX1:~/repos/github/open-sdg/open-sdg$ docker --version
Docker version 18.09.7, build 2d0083d
```

SavvasStephanides:
@brockfanning Rather strange. Can you try updating Docker in case you have an easier version?

SavvasStephanides:
@brockfanning I made a few updates. The Docker image is now at [Docker Hub](https://hub.docker.com/r/opensdg/make) so no files are required to be stored in the repo. Just checkout `serve_docker` repo and use `make docker.serve` or `make docker` to access the container instance and run anything.

brockfanning:
[...]

PR #475, opened 2020-02-08
GeoJSON revamp: data in repo

### LINKED ISSUE #109: Mapping: Ignore 'Units' and disaggregation columns
Having a 'Units' column in the CSV prevents the map from displaying data. Can we hide the 'Units' column from the mapping script as a short-term solution, and display the first Units option by default?

Alternatively, we could include a 'Units' option in the map, but this implies to users that the disaggregations are available to be used on the map, which they aren't (yet).

**Note**: This issue is a problem with disaggregation in general, as the presence of Geographical data overlapping with another category in the CSV prevents data from displaying on the map. We would like to have disaggregations working on the map in future, but this is likely to be complex.

brockfanning:
Fixes #109 

This depends on https://github.com/open-sdg/sdg-build/pull/99 so a full review needs to be done of that PR as well as this one.

## Breaking changes

This changes the map functionality so that it expects the GeoJSON to be in the data repository. This needs to happen in the data repository via sdg-build. So, if a country upgrades to this version of open-sdg without also upgrading sdg-build versions, their maps will stop functioning.

Along the same lines, the map configuration in _config.yml needs to change.

## Overview of changes

1. The map no longer looks for a single GeoJSON file at a hardcoded location. Instead, it looks for individual GeoJSON files in the data repo that have been pre-built for each indicator.
2. The GeoJSON download button on the maps has been removed, and standard links have been added below the map. Some benefits of this:
    a. They use the same style as other download buttons in the platform
    b. The download is a simple link, so it can be reached by scripts (interoperability)
    c. Instead of confusingly only downloading the visible layer, it now displays all the layers as separate buttons
3. A lot of Javascript code in open-sdg has been removed because it isn't needed
4. Several javascript libraries are no longer being loaded on non-map pages (saves bandwidth)
5. I've removed any reference to the `data_geocode_regex` metadata field. I think this was non-functional anyway, and we would need to reimplement that functionality if it's still wanted.
6. The GeoJSON files now contains all the disaggregation data, though it's currently just using the first one. (After this is merged I'll post a followup PR that displays all the disaggregation, fixing #374.)
7. The options in the map configuration have changed.

## Screenshots

Here is an example of the new download button (the "My English regions" is configurable and just demonstrates that it is translated):
Screenshot from 2020-02-11 09-52-01


brockfanning:
@LucyGwilliamAdmin @SavvasStephanides A lot to go over with this one. We could discuss in the next call. Testing is a bit of a hassle because it depends on an sdg-build change.

christoph4711:
Where I can test the new map functionalities?
And how can I update my starter repo? (Sorry for this question, but I am not familiar with this github stuff)

brockfanning:
@christoph4711 Can you check what version of open-sdg you're using now? It would be in the `remote_theme` line of `_config.yml`. Also can you check the version of jekyll-open-sdg-plugins that you're using (this would be in `Gemfile`). Finally, in the data repository, what version of sdg-build are you using? (this would be in the `scripts/requirements.txt` file in the data repository).

The reason I ask is that in order to test cutting-edge features, you'll need to be using the latest versions of all those things.

Also for your awareness, we are on track to release a stable version 1.0.0 of Open SDG some time in the next month. When we release that, we'll include thorough upgrade instructions. So it may be more efficient to hold off.

In the meantime, I'll try to add some screenshots to this PR so you can see any visual changes.

brockfanning:
One nice thing about this - this makes it almost trivial to add a disaggregation drop-down to the maps. In fact, I'm almost tempted to add it to this PR... Here's an example of what it might look like:
Screenshot from 2020-02-11 09-48-51

christoph4711:
@brockfanning 
This feature would be great! I try to hold off  until 1.0.0. :-)

brockfanning:
Since it was so easy to add, I went ahead and added the disaggregation select to this PR. I've updated the PR description and title.

brockfanning:
@AnnCorp @SavvasStephanides @LucyGwilliamAdmin I've removed the disaggregation select from this PR, and updated the description/title. I can make a follow-up PR after this for continuing with that feature.

Any review/testing is welcome! As a reminder, it depends on another PR in sdg-build.

brockfanning:
@SavvasStephanides @LucyGwilliamAdmin This should be a bit easier to test, now that the sdg-build change has been merged.

brockfanning:
@SavvasStephanides @LucyGwilliamAdmin Just a note that I resolved the merge conflicts on this one.

LucyGwilliamAdmin:
@brockfanning @SavvasStephanides really sorry so I think I've missed something about the make serve stuff?

How do I test this using make serve?

brockfanning:
@LucyGwilliamAdmin In theory, if you have the required software installed (Ruby, Python, Node.js) you can test it by pull my branch and running `make serve` in the root of the repository. If you don't already have my remote set up in Git, the command would be:
```
git remote add brock https://github.com/brockfanning/open-sdg.git
```
Then you could do this:
```
git fetch brock
git checkout geojson-in-data-repo
make serve
```

LucyGwilliamAdmin:
Thanks @brockfanning 

So I've got all the software and this worked right up until the last line (`make serve`), where I got the error:
'make' is not recognized as an internal or external command

brockfanning:
@LucyGwilliamAdmin Ah, I think that it doesn't work in Windows. I guess this is a good reason we need the Docker solution @SavvasStephanides is working on.

LucyGwilliamAdmin:
Ok no worries, thanks for you help though! I think I'll have to leave this one all to you @SavvasStephanides

SavvasStephanides:
@LucyGwilliamAdmin If you can run Docker you can give it a go https://github.com/open-sdg/open-sdg/issues/197#issuecomment-594548506

SavvasStephanides:
@brockfanning Do you have an indicator I can test this on?

brockfanning:
@SavvasStephanides The tests for this use indicator 1.3.1.

SavvasStephanides:
Thanks @brockfanning ! Seems to be working nicely. I'll have a quick check on the code and approve.

_includes/assets/js/plugins/jquery.sdgMap.js
```diff
@@ -2,7 +2,11 @@
  * TODO:
  * Integrate with high-contrast switcher.
  */
-(function($, L, chroma, window, document, undefined) {
+(function($) {
+
+  if (typeof L === 'undefined') {
```
SavvasStephanides: Where does `L` point to here, if you removed it from the function's arguments?

brockfanning: @SavvasStephanides That's a global variable created by Leaflet. It's not there, I assume that Leaflet did not get successfully loaded.

SavvasStephanides: Ah fair enough. Thanks!

PR #481, opened 2020-02-11
Use lunr-languages plugin, or simple fallback search, for non-english languages

### LINKED ISSUE #480: Search not working for some non-English languages
In some non-English languages, performing searches always returns 0 results. At least, I know this happens with Kazakh and Russian.

brockfanning:
Fixes #480 

This works for languages covered by lunr-languages. See here for a list: https://github.com/MihaiValentin/lunr-languages/

This also provides a fallback search functionality for any other languages.

### DIFF
--- a/_layouts/search.html
+++ b/_layouts/search.html
@@ -55,4 +55,8 @@ opensdg.searchIndexExtraFields = [];
 </div>
 
 <script src="https://unpkg.com/lunr/lunr.min.js"></script>
+{% if page.language != 'en' %}
+  <script src="https://cdn.jsdelivr.net/npm/lunr-languages@1.4.0/lunr.stemmer.support.js"></script>
+  <script src="https://cdn.jsdelivr.net/npm/lunr-languages@1.4.0/lunr.{{ page.language }}.js"></script>
+{% endif %}
 {% include footer.html %}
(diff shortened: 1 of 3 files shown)


brockfanning:
Though this doesn't implement all the languages we'd like to support, it is a step in the right direction. I'd like to learn more about the lunr-languages project to see how hard it is to add new languages.

_layouts/search.html
```diff
@@ -55,4 +55,8 @@ <h1>{{ page.t.search.search_results_for }} '<span id="search-term"></span>'</h1>
 </div>
 
 <script src="https://unpkg.com/lunr/lunr.min.js"></script>
+{% if page.language != 'en' %}
+  <script src="https://cdn.jsdelivr.net/npm/lunr-languages@1.4.0/lunr.stemmer.support.js"></script>
+  <script src="https://cdn.jsdelivr.net/npm/lunr-languages@1.4.0/lunr.{{ page.language }}.js"></script>
```
SavvasStephanides: My only concern with this is whether the language codes match the language codes in `lunr.xx.js` files from `lunr.languages`

brockfanning: In that case the script tag will get a 404. There should be no other side effects. Admittedly messy, but at least it's only on the search page.

The bigger problem is that their search likely won't work. I'm not sure yet if there's a way to solve that, other than by implementing a new language in lunr-languages.

brockfanning:
@SavvasStephanides @LucyGwilliamAdmin Hold off on reviewing this, I'll try to include some custom code to allow some sort of basic search functionality for the non-lunr-languages languages.

brockfanning:
@SavvasStephanides @LucyGwilliamAdmin Ok this is ready again for testing. It should now have a working fallback for any languages that don't have lunr support.

PR #485, opened 2020-02-13
Fixed rounding on reporting status page progress bar

lindamejia:


brockfanning:
Awesome, thanks @lindamejia. One request if possible: there are some lines where only indentation changed. It would make review a bit easier if those changes could be left out of the PR, so we can focus on the functional differences.

PR #486, opened 2020-02-13
Add an introduction banner to the frontpage

SavvasStephanides:
This change adds an introductory paragraph to the frontpage, right above the goal tiles.

The banner text is retrieved by translations, so the text for both the title and description must be added to `frontpage.yml` in order for the banner to show up (which also makes it optional).

Example `frontpage.yml`:
```yml
intro_title: 17 Goals to transform our world
intro_description: The Sustainable Development Goals are a call for action by all countries – poor, rich and middle-income – to promote prosperity while protecting the planet. They recognize that ending poverty must go hand-in-hand with strategies that build economic growth and address a range of social needs including education, health, social protection, and job opportunities, while tackling climate change and environmental protection.
```

If either of these is missing in translations, the banner will not be displayed. 

This change also changes the `<country> data for Sustainable Development Goal indicators` heading from `<h1>` to `<h2>`.

[Feature branch](http://sdgdev-813006012.eu-west-1.elb.amazonaws.com/opensdg1-introbanner/)

# Screenshots

### Before

image

### After

image

### Before

image

### After

image

### DIFF
--- a/_layouts/frontpage.html
+++ b/_layouts/frontpage.html
@@ -2,10 +2,12 @@
 {% include header.html %}
 <div id="main-content" class="container goal-tiles" role="main">
 
-    <div class="site-intro">
+    {% if page.t.frontpage.intro_title and page.t.frontpage.intro_description %}
+    <div class="site-intro spacer medium">
         <h1>{{ page.t.frontpage.intro_title }}</h1>
         <div class="description">{{ page.t.frontpage.intro_description }}</div>
     </div>
+    {% endif %}
 
     {% assign country_name = site.country.name | t %}
     {% assign heading_default = page.t.frontpage.heading | replace: '%name', country_name %}
--- a/_layouts/frontpage.html
+++ b/_layouts/frontpage.html
@@ -4,8 +4,8 @@
 
     {% if site.homepage_introduction_banner %}
     <div class="site-intro spacer medium">
-        <h1>{{ site.homepage_introduction_banner.title }}</h1>
-        <div class="description">{{ site.homepage_introduction_banner.description }}</div>
+        <h1>{{ site.homepage_introduction_banner.title | t }}</h1>
+        <div class="description">{{ site.homepage_introduction_banner.description | t }}</div>
     </div>
     {% endif %}
 
--- a/_layouts/frontpage.html
+++ b/_layouts/frontpage.html
@@ -3,8 +3,8 @@
 <div id="main-content" class="container goal-tiles" role="main">
 
     <div class="site-intro">
-        <h1>17 Goals to transform our world</h1>
-        <div class="description">The Sustainable Development Goals are a call for action by all countries – poor, rich and middle-income – to promote prosperity while protecting the planet. They recognize that ending poverty must go hand-in-hand with strategies that build economic growth and address a range of social needs including education, health, social protection, and job opportunities, while tackling climate change and environmental protection.</div>
+        <h1>{{ page.t.frontpage.intro_title }}</h1>
+        <div class="description">{{ page.t.frontpage.intro_description }}</div>
     </div>
 
     {% assign country_name = site.country.name | t %}
--- a/_layouts/frontpage.html
+++ b/_layouts/frontpage.html
@@ -3,10 +3,8 @@
 <div id="main-content" class="container goal-tiles" role="main">
 
     <div class="site-intro">
-        <div class="container">
-            <h1>17 Goals to transform our world</h1>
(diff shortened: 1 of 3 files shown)


brockfanning:
@SavvasStephanides Looks good to me.

I would recommend, though, that we allow countries to control it in the _config.yml file. This would allow us to add documentation of this to [this page](https://open-sdg.readthedocs.io/en/latest/configuration/). As is, I think we'd need to complicate our documentation a bit to explain to users that they need to override a translation file to control it.

So basically instead of checking for `page.t.frontpage.intro_title` and `page.t.frontpage.intro_description` you could check for `site.intro_title` and `site.intro_description`.

And to make it easy for countries out of the box, we could translate those items and add them to [this repo](https://github.com/open-sdg/translations-open-sdg). And then we'd add the following to the site starter repo's [_config.yml](https://github.com/open-sdg/open-sdg-site-starter/blob/develop/_config.yml):

```
intro_title: frontpage.intro_title
intro_description: frontpage.intro_description
```

That way a mono-lingual implementation could control it by simply changing their `_config.yml` file, like so:

```
frontpage_intro_title: My intro title
frontpage_intro_description: My description title
```

And if a country just doesn't want that section to appear, they can simply remove those lines from their `_config.yml`.

SavvasStephanides:
@brockfanning Makes sense. I'll make the changes

SavvasStephanides:
Also if countries need to override their banner text, wouldn't it be better to override their `.yml` files in their translations?

brockfanning:
@SavvasStephanides Yes, I think we should have translations in place, at least before we release version 1.0.0. It would be good to have some final copy before doing all the Google Translations (since that is a tedious chore...) If needed, though, we can put placeholder text in the translations-open-sdg repo for now.

Regarding how countries control the text, my thinking is that having it in the _config.yml gives them the most options. They can either:
1. Leave the _config.yml pointing at `frontpage.intro_title` and `frontpage.intro_description`, and add/update a `frontpage.yml` file in their translations folder
2. If they are mono-lingual, simply put the text directly in _config.yml
3. If they prefer, update their _config.yml to point to any arbitrary translation key, like `my-custom-group.my-site-intro`. (Assumes they have a `my-custom-group.yml` file in their translations folder, with an entry for `my-site-intro`.)

But I think the biggest benefit is just staying consistent and simplifying documentation. So far I don't think we have any features that depend on overriding translations for control, so we haven't had to add any documentation instructing users to do that.

SavvasStephanides:
@brockfanning I finally understood what you mean (I think):

So with this change, the homepage banner text is configured in the `_config.yml` file like so:

```yml
homepage_introduction_banner:
    title: title goes here
    description: description goes here
```

If it's not set up in the config, it will not display.

brockfanning:
@SavvasStephanides It's looking good. Just a little bit of Liquid filtering is needed:

1. Run the title through the `t` filter, to allow it to be translated: `{{ site.homepage_introduction_banner.title | t }}`
2. Run the description through the `t` filter as well, and probably also the `markdownify` filter if you want to allow Markdown in the description: `{{ site.homepage_introduction_banner.description | t | markdownify }}`. Note though that using `markdownify` means it will include paragraphs ([p] tags), so you might need to adjust any top/bottom margins if that causes an issue.

Also just my opinion but I don't think that `homepage_introduction_banner_is_setup` is necessary. You're only using it once, in the very next line, and I think it's more clear to use `{% if site.homepage_introduction_banner %}`.

SavvasStephanides:
Thanks @brockfanning 

> 1. Run the title through the `t` filter, to allow it to be translated: `{{ site.homepage_introduction_banner.title | t }}`

Good idea

> 2. Run the description through the `t` filter as well, and probably also the `markdownify` filter if you want to allow Markdown in the description: `{{ site.homepage_introduction_banner.description | t | markdownify }}`. Note though that using `markdownify` means it will include paragraphs ([p] tags), so you might need to adjust any top/bottom margins if that causes an issue.

Let's keep it basic without any Markdown for now. Let's not overcomplicate things.

> Also just my opinion but I don't think that `homepage_introduction_banner_is_setup` is necessary. You're only using it once, in the very next line, and I think it's more clear to use `{% if site.homepage_introduction_banner %}`.

This was more of a "cleaner code" thing. I think `site.homepage_introduction_banner` doesn't make it obvious straight away that we're checking if that it is set up. I like to have things more explicit.

brockfanning:
@SavvasStephanides Actually `{% if site.xyz %}` is pretty standard in Jekyll for quick boolean checks for site configuration. If you search our codebase for `if site.` you'll find it all over the place. It reads like "if the site has xyz", which I feel is pretty clean already.

SavvasStephanides:
@brockfanning I see that, just wanted to be more explicit, since we're trying to make things cleaner. I'll change it back to keep it more consistent

brockfanning:
@SavvasStephanides Sounds good. Once you add the `t` filters it should be ready to merge, from my perspective.

PR #489, opened 2020-02-14
Bulk download link beneath the homepage grid

### LINKED ISSUE #380: Bulk download indicator data
Covers two key user stories

AS AN international consumer of country level information
I NEED to bulk download data/metadata
SO THAT I can quickly acquire country data to load into my own systems

AS AN involved analyst
I NEED to be able to download all of the data for the indicators without having to go into them individually

Initial criteria (in order of urgency/portance)
- Data available in open data format - CSV
- Download option clearly visible on site
- Initially manual download OK, but consider machine access later
- Add SDMX format for international data exchange
Consider through research what other options analysts want eg Excel format (like ONS website?) or being able to select what download

UNECE developing regional platform and have been reviewing country national reporting platforms and and how easy data can be bulk downloaded. Currently no bulk data download in Open SDG so one is needed.

Note Polish platform seems to have static Excel file download for all data, linked from 18th icon on main page.

brockfanning:
Fixes #380 

This depends on:
* The latest version of jekyll-open-sdg-plugins: 1.0.0.rc22
* The latest translations from [open-sdg/translations-open-sdg](https://github.com/open-sdg/translations-open-sdg)

The file size and last updated date come from the sdg-build library.

## Screenshots:

English:
bulk-download-english

French:
bulk-download-french

Full page:
screenshot-127 0 0 1_4000-2020 02 13-21_06_49

### DIFF
--- a/_includes/components/download-all-data.html
+++ b/_includes/components/download-all-data.html
@@ -1,0 +1,13 @@
+<div class="zip-download-container">
+    <a href="{{ page.remote_data_prefix }}/zip/all_indicators.zip"
+       class="btn btn-primary btn-download" aria-describedby="zip-download-info">
+        {{ page.t.frontpage.download_all }}
+    </a>
+    <div id="zip-download-info">
+        {{ page.t.frontpage.zip_file }} - {{ site.data.zip.size_human }}
+        {% if site.data.zip.timestamp %}
+            <br>
+            {{ page.t.metadata_fields.national_data_update_url }} - {{ site.data.zip.timestamp | t_date: 'standard' }}
+        {% endif %}
+    </div>
+</div>
--- a/_layouts/frontpage.html
+++ b/_layouts/frontpage.html
@@ -35,5 +35,15 @@
         </div>
     {% endif %}
     </div>
+    <div class="zip-download-container">
+        <a href="{{ page.remote_data_prefix | append: '/zip/all_indicators.zip' }}" class="btn btn-primary btn-download">
+            {{ page.t.frontpage.download_all }} - {{ site.data.zip.size_human }}
+        </a>
+        {% if site.data.zip.timestamp %}
+        <div class="zip-date">
+            {{ page.t.metadata_fields.national_data_update_url }} - {{ site.data.zip.timestamp | t_date: 'standard' }}
+        </div>
+        {% endif %}
+    </div>
 </div>
 {% include footer.html %}
--- a/_layouts/frontpage.html
+++ b/_layouts/frontpage.html
@@ -35,15 +35,6 @@
         </div>
     {% endif %}
     </div>
-    <div class="zip-download-container">
-        <a href="{{ page.remote_data_prefix | append: '/zip/all_indicators.zip' }}" class="btn btn-primary btn-download">
-            {{ page.t.frontpage.download_all }} - {{ site.data.zip.size_human }}
-        </a>
-        {% if site.data.zip.timestamp %}
-        <div class="zip-date">
-            {{ page.t.metadata_fields.national_data_update_url }} - {{ site.data.zip.timestamp | t_date: 'standard' }}
-        </div>
-        {% endif %}
-    </div>
+    {% include components/download-all-data.html %}
 </div>
 {% include footer.html %}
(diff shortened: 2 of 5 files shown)


brockfanning:
@phillipgardiner @AnnCorp @LucyGwilliamAdmin @SavvasStephanides Here's another try at the bulk download button, with file size and date.

phillipgardiner:
Hey @brockfanning , thanks for doing this. I wonder if it might be worth considering moving the file type and size data out of the button? This may stop some countries from having really long buttons. I've mocked up a little example, but it may be worth playing around with placement of the text (e.g. all on one line, next to button etc to see what works best. Thoughts? 
[All data.pdf](https://github.com/open-sdg/open-sdg/files/4203948/All.data.pdf)

brockfanning:
@phillipgardiner That works for me, no objections here. Ideally we can decide first on the language used, and then play around with placement and design. The language part involves a lot of clicking/copying/pasting in Google Translate, so it's best to nail that down first.

@SavvasStephanides @AnnCorp @LucyGwilliamAdmin Are you OK with this language?

- "Download all SDG data"
- ".ZIP file" (plus the file size, which won't be translated)
- "Data last updated"

I only suggest "Data last updated" instead of "Last updated" because we already have "Data last updated" translated.

LucyGwilliamAdmin:
I'm happy with those :)

AnnCorp:
@brockfanning I'm happy with that proposal too though am wondering could we have just 'Download all data' and remove the 'SDG' bit?  One less thing to translate (or are there commonly used corresponding acronyms in the languages we have?) and as it's in an SDG platform maybe don't need that bit?

SavvasStephanides:
I agree with @AnnCorp , I think "Download all data" is better. 

Would you consider adding a small [download icon from FontAwesome](https://fontawesome.com/icons/download?style=solid) on the button?

Last point: I'm wondering how adding the "data last updated" bit on the bottom would work for accessibility, ie. they'll have to lose focus of the button in order to access this information. Let me know your thoughts

brockfanning:
That's a good point about the accessibility. We'll need to make sure that the labels have the appropriate properties, like maybe aria-describedby.

SavvasStephanides:
@brockfanning Maybe a `aria-label` would work as well.

brockfanning:
Just pushed an update to this PR. It depends on https://github.com/open-sdg/sdg-build/pull/103 so tests will fail until that is merged. Here's a screenshot:

Screenshot from 2020-02-17 19-34-48

AnnCorp:
Looks good @brockfanning . One thing have just noticed from comparing is that the other download buttons in current site use title case eg 'Download Chart Image' 'Download Source CSV' rather than sentence case like in the screen shot above which would make the other buttons like 'Download chart image'. Checked the ONS website for reference and sentence case is used there. Think we need to be consistent and go one way or the other.  I think I prefer the sentence case Brock has used above, but have you got any thoughts @phillipgardiner  based on GDS guidance?

phillipgardiner:
@AnnCorp @brockfanning GDS guidance is to use sentence case (e.g. Download all data)

brockfanning:
@SavvasStephanides I forgot to reply regarding the icon: I think that would look cool, personally. We could try that here, though to be consistent maybe we should do that in a separate ticket and apply it to all download buttons. (Maybe make them all sentence cased at the same time?)

brockfanning:
Also a note - I removed the "." from ".ZIP" only because Google Translate was doing some weird things with it (putting it in front of the translated word for "file", in some languages).

SavvasStephanides:
@brockfanning I've done a quick check and this is what I get. Is this an issue with the translations not being accessed?

image

brockfanning:
@SavvasStephanides Yes, that looks like you're probably using an older tag of a translation repository from before I added the translations for this feature. I'd recommend, when testing, to just point to the translation repositories without specifying a tag. But the easiest way to test this is to pull my brockfanning/bulk-download-link-2 branch and do `make serve`.

brockfanning:
@SavvasStephanides I added a note in the description about depending on the latest translations (sorry I didn't have that before).

SavvasStephanides:
@brockfanning Thanks! Still facing a few issues with `make`, which I'll try and solve later. 

What's the URL for the new translations to point to?

Issue #488, opened 2020-02-13
Indicator page accessibility issues


brockfanning:
Here are the 3 things that pa11y is complaining about, on our indicator pages:

```
 • This element has insufficient contrast at this conformance level. Expected a contrast
   ratio of at least 4.5:1, but text in this element has a contrast ratio of 2.79:1.
   Recommendation:  change text colour to #565656.

   (#clear)

   <button id="clear" class="disabled">Clear selections <i class="fa f...</button>
```
The above seems valid - the "Clear selections" button in the left sidebar does not have enough contrast when disabled. Of course, it's meant to be disabled, so maybe we can add an aria property to show that it's disabled?

```
 • This table appears to be used for layout, but contains a caption element. Layout tables
   must not contain captions.

   (#datatables > table)

   <table class="table-responsive table table-hover dataTable no-footer" role="grid">
   <caption>Proportion of p...</table>
```

The above seems clearly invalid. That table is a regular table, and not being used for layout. So we just need to tell pa11y not to complain about that.

```
 • This element has insufficient contrast at this conformance level. Expected a contrast
   ratio of at least 4.5:1, but text in this element has a contrast ratio of 3.93:1.
   Recommendation:  change background to #fefefe.

   (#indicator-main > section > ul > li:nth-child(2) > a)

   <a class="nav-link" data-toggle="tab" href="#global" aria-controls="global" role="tab"
   data-on="click" data-event-category="Tab change" data-event-action="Change metadata view"
   data-event-label="Change to My global metadata title tab">My global metad...

 • This element has insufficient contrast at this conformance level. Expected a contrast
   ratio of at least 4.5:1, but text in this element has a contrast ratio of 3.93:1.
   Recommendation:  change background to #fefefe.

   (#indicator-main > section > ul > li:nth-child(3) > a)

   <a class="nav-link" data-toggle="tab" href="#sources" aria-controls="sources" role="tab"
   data-on="click" data-event-category="Tab change" data-event-action="Change metadata view"
   data-event-label="Change to My sources title tab">My sources title</a>

 • This element has insufficient contrast at this conformance level. Expected a contrast
   ratio of at least 4.5:1, but text in this element has a contrast ratio of 3.93:1.
   Recommendation:  change background to #fefefe.

   (#indicator-main > section > ul > li:nth-child(4) > a)

   <a class="nav-link" data-toggle="tab" href="#edit" aria-controls="edit" role="tab">Edit</a>

[...]

PR #492, opened 2020-02-14
Look for translated builds when rendering the no-javascript headline

brockfanning:
We have a data table that is rendered in HTML, but hidden with javascript. However, it wasn't compatible with the "translated builds" approach, where the data repo has separate builds for each language. This fixes that.

Side benefit: this fixes the pa11y complaint about layout tables mentioned in #488

### DIFF
--- a/_includes/components/headline.html
+++ b/_includes/components/headline.html
@@ -1,4 +1,7 @@
 {% assign sdg_headline_data = site.data.headlines[page.indicator.slug] %}
+{% if site.data[page.language].headlines[page.indicator.slug] %}
+  {% assign sdg_headline_data = site.data[page.language].headlines[page.indicator.slug] %}
+{% endif %}
 
 <div id="headlineTable">
 


_includes/components/headline.html
```diff
@@ -1,4 +1,7 @@
 {% assign sdg_headline_data = site.data.headlines[page.indicator.slug] %}
+{% if site.data[page.language].headlines[page.indicator.slug] %}
+  {% assign sdg_headline_data = site.data[page.language].headlines[page.indicator.slug] %}
```
SavvasStephanides: I'm curious if you can assign `sdg_headline_data` before your `if` statement here.

brockfanning: @SavvasStephanides I don't understand the question - maybe you could give an example?

SavvasStephanides: Sorry Brock, I had to have a closer look to see that the two `sdg_headline_data` declarations are different and not duplicate (line1 and line3).

PR #496, opened 2020-02-16
WIP: Min max values by unit

### LINKED ISSUE #361: min_ /max_value with multiple units
I just tried the min_ /max_values (very cool) and noticed that they are fixed when changing the unit see [here](https://g205sdgs.github.io/sdg-indicators/12-2-1/).
Is there a possibility to define min_/ max_values, which are connected to the respective unit?

brockfanning:
Fixes #361 

Depends on https://github.com/open-sdg/sdg-build/pull/106

* The changes the `chart_max_value` and `chart_min_value` metadata fields into a `chart_limits` field. So this is a breaking change.
* This includes a moderate refactor of our global `opensdg` javascript object, wrapping it in a function to better control its interface. This change also has some breaking changes:
    * This changes the way the remoteDataBaseUrl is retrieved
    * This changes the way custom data rounding functions are added
    * This changes the way chart config alteration functions are added


brockfanning:
This is really more of a chart revamp/refactor. I'll close this and open a new one.

Issue #164, opened 2019-05-10
Pie chart


brockfanning:
Is it possible to include pie chart and/or stacked bar visualizations? These would need to fundamentally involve disaggregation, and would only work with percentage-based datasets. But it should be possible, right?

LucyGwilliamAdmin:
So could these only be used for indicators that have one disaggregation?

LucyGwilliamAdmin:
Maybe something like this:
![Uploading piechart1.png…]()
![Uploading piechart2.png…]()

Not only does the dataset have to be percentage-based but it has to be the case where the disaggregation is split e.g. not like [1.2.1](https://sustainabledevelopment-uk.github.io/1-2-1/)

brockfanning:
Linking to similar discussion: https://github.com/dawncomer/open-sdg-site-starter/issues/16

brockfanning:
To carry over an idea from the other thread: there could be a metadata field `stacked_disaggregation` which could be used like:
```
stacked_disaggregation: Gender
```
The idea is that this would communicate to the platform that this disaggregation should get "stacked" whereas all other disaggregations are displayed separately. So for a bar chart using the example above, all "Gender" disaggregations would be stacked in one bar, whereas all other disaggregations ("Age", "Region", etc.) would be in their own bars. I guess it would make sense to start with all of the "Gender" disaggregations pre-selected (which can already be done with `data_start_values`).

It's kind of a mind-bender to think about how it would all fit together. But that might be a simple place to start.

brockfanning:
It would be nice to get this into 1.0.0, since it may carry with it some structural changes.

brockfanning:
Stacked bars can be tested in #497, though it depends on another PR in sdg-build.

brockfanning:
We have implemented an approach to stack bar charts, so I'm going to change this issue to just pie charts.

PR #497, opened 2020-02-17
Chart refactor for limits and stacked bars

### LINKED ISSUE #361: min_ /max_value with multiple units

brockfanning:
This fixes #361 and part of #164 (the stacked bar chart), and includes a few breaking changes:

1. Remove support for `graph_min_value` and `graph_max_value` (replace them with `graph_limits`)
2. Change the way that `data_start_values` needs to be formatted (list of objects instead of list of strings). This was necessary to allow multiple start values for the same disaggregation (eg, starting with both "Female" and "Male" selected under "Sex").
3. Change the way the chart alterations callbacks are added (`opensdg.chartConfigAlter(callback)` instead of `opensdg.chartConfigAlter=callback`). This was necessary to allow multiple alterations of the chart config.

This also adds a new metadata field called `graph_units_without_headline`, which can be used to hide the headline for the specified units (which will almost always be the case with stacked bars, most likely).

To test, you can do a `make serve` and check out indicator 2-3-1 (stacked bars) and indicator 2-3-2 (min/max values).

Here's a screenshot of the stacked bars:
stacked_bar_5_1_2


brockfanning:
@SavvasStephanides @LucyGwilliamAdmin The sdg-build change has been merged, so this is ready for testing. The easiest way to do a quick test is from my note above: To test, you can do a `make serve` and check out indicator 2-3-1 (stacked bars) and indicator 2-3-2 (min/max values).

PR #498, opened 2020-02-18
Add flag next in header next to logo

SavvasStephanides:
This adds a flag next to the SDG logo.

Screenshot:

image

### DIFF
--- a/_sass/layouts/_header.scss
+++ b/_sass/layouts/_header.scss
@@ -27,7 +27,7 @@ header {
 .navbar-brand {
   display: block;
   width: 75%;
-  max-width: 300px;
+  max-width: 360px;
   img {
     width: 100%;
 
--- a/_sass/layouts/_header.scss
+++ b/_sass/layouts/_header.scss
@@ -30,6 +30,12 @@ header {
   max-width: 300px;
   img {
     width: 100%;
+
+    padding-right: 60px;
+    background: url('../img/header_flag.png');
+    background-size: 57px;
+    background-repeat: no-repeat;
+    background-position: right center;
   }
 }
 


brockfanning:
@SavvasStephanides
1. I thought the flag was going to be on the left - am I remembering your mockup wrong?
2. Regardless it should probably have some margin between the two logo images.
3. Isn't it going to be tough to make it accessible using a CSS background image? I would guess that using an [img] tag would be simpler to make accessible (by giving it an alt property).
4. Shouldn't this PR also include the image file itself? Like some generic placeholder flag?
5. Do we want to allow for countries to turn off the flag with a config setting? I think we discussed it but can't remember the consensus.

SavvasStephanides:
@brockfanning 

> 1. I thought the flag was going to be on the left - am I remembering your mockup wrong?

The mockup points the flag to the right

image

> 2. Regardless it should probably have some margin between the two logo images.

I'll check that

> 3. Isn't it going to be tough to make it accessible using a CSS background image? I would guess that using an [img] tag would be simpler to make accessible (by giving it an alt property).

As far as accessibility is concerned, I think it might be better to have it as an extension of the logo. For keyboard users, adding a separate image would lead to an unnecessary extra keypress/swipe and might potentially cause confusion (two images going to the same place, two logos).

> 4. Shouldn't this PR also include the image file itself? Like some generic placeholder flag?

I don't think a placeholder image would be good. It should be more of an opt-in thing where they can add their own flags when they want. 

> 5. Do we want to allow for countries to turn off the flag with a config setting? I think we discussed it but can't remember the consensus.

Yes, I forgot about that, although not sure how to set this up in the `scss` file. Do you have any idea how to do this? (Possibly pass it as a parameter somehow)

SavvasStephanides:
Decided to close this. We're leaving it up to individual countries to update their logo

PR #499, opened 2020-02-18
Make the accessibility tests automatic

brockfanning:
Now that we're passing the current accessibility tests, we can make them automatic.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -53,8 +53,8 @@ test.features:
 	# Cucumber.
 	cd tests && npx cucumber-js
 
-test.accessibility: test.before test.prep test.serve
+test.accessibility:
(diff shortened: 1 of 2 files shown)


SavvasStephanides:
I think it might be good to keep the accessibility tests in the README in case anyone wants to manually run an accessibility test. Thoughts?

brockfanning:
@SavvasStephanides This changes lumps the accessibility test in with the others, so just running `make test` will run accessibility tests.

brockfanning:
@SavvasStephanides Think I misunderstood - that does seem useful to allow for just running one type of test. I'll make a change and try it in this PR, one sec.

Issue #478, opened 2020-02-11
Vertically center H1 in indicator page banners


brockfanning:
This is a follow-up from discussion in #474

brockfanning:
In addition to the markup change, @phillipgardiner suggests making the banner text vertically centered.

Tech note - what that require "flex" display maybe? I feel like vertical alignment is always tricky.

phillipgardiner:
I also believe the text in the banner may look better as a 'h1' rather than a 'h2'. (separate note: not sure why 'h1' is smaller than 'h2'??)

brockfanning:
Renaming this, since the other part has already been done.

phillipgardiner:
No longer required due to changes to the indicator banner made within 'Adjust heading structure across site #651'

Issue #477, opened 2020-02-11
Smaller goal icons in goal/indicator page banners


brockfanning:
This is a follow-up on the discussion in #474 

I believe the consensus was to limit the goal icons in goal page banners and indicator page banners to 80% instead of 100%. But @SavvasStephanides @LucyGwilliamAdmin @phillipgardiner correct me if I'm wrong.

phillipgardiner:
80% :)

brockfanning:
A note that this has been postponed because the grid system causes unwanted whitespace between the icon and the H1.

phillipgardiner:
No longer required due to changes to the indicator banner made within 'Adjust heading structure across site #651'

PR #500, opened 2020-02-18
Remove goal line from indicator banner

SavvasStephanides:
Screenshot:

image

### DIFF
--- a/_layouts/indicator.html
+++ b/_layouts/indicator.html
@@ -14,11 +14,6 @@
         </a>
       </div>
       <div class="col-xs-8 col-md-9 col-lg-10">
-        <h1>
-          <a href="{{ page.goal.url }}">
-            <span class="hidden-sm hidden-md hidden-lg">{{ page.t.general.goal }} {{ page.goal.number }}: </span>{{ page.goal.name }}
-          </a>
-        </h1>
         <h2>{{ page.t.general.indicator }} {{ page.indicator.number }}: {{ page.indicator.name }}</h2>
       </div>
     </div>
--- a/_sass/base/_typography.scss
+++ b/_sass/base/_typography.scss
@@ -63,7 +63,7 @@ h6 {
   }
 
   img {
-    max-width: 100%;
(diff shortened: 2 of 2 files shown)


brockfanning:
@SavvasStephanides Looks like this addresses both #478 and #477. I'm wondering if it should be split up, because of some technical issues. More below:
* In #478 there was also a recommendation to vertically center the indicator name, and to change it to an H1 (which would make sense semantically).
* With #477 (smaller goal icon) it looks like there is going to be a conflict with the Bootstrap grid system, right? I see you shrunk the icon, but it left white space between the icon and the indicator name, I'm guessing because of the grid system. My gut reaction is that if the grid system is going to cause a problem here, we might consider **not** shrinking that image. Do you think there is an _easy_ way to get rid of that left-over white space between the icon and the indicator name?
* Also a nitpick: the number that was discussed in #477 was 80%.

SavvasStephanides:
Thanks for the feedback @brockfanning 

> @SavvasStephanides Looks like this addresses both #478 and #477. I'm wondering if it should be split up, because of some technical issues. More below:
> 
> * In #478 there was also a recommendation to vertically center the indicator name, and to change it to an H1 (which would make sense semantically).

Yes, makes sense

> * With #477 (smaller goal icon) it looks like there is going to be a conflict with the Bootstrap grid system, right? I see you shrunk the icon, but it left white space between the icon and the indicator name, I'm guessing because of the grid system. My gut reaction is that if the grid system is going to cause a problem here, we might consider **not** shrinking that image. Do you think there is an _easy_ way to get rid of that left-over white space between the icon and the indicator name?

I'm not sure, but I don't mind the bit of whitespace it leaves behind. What do you think?

> * Also a nitpick: the number that was discussed in #477 was 80%.

I'll change that.

brockfanning:
@SavvasStephanides I guess I could grow accustomed to that extra white-space. I'm probably not the best judge. Maybe we should make a new screenshot after #506 gets merged and continue to discuss this in #477.

PR #502, opened 2020-02-19
Fit total number of indicators on one line

LucyGwilliamAdmin:


brockfanning:
@LucyGwilliamAdmin Small thing, but could you tweak the indentation so it matches the other code?

LucyGwilliamAdmin:
@brockfanning I'm not sure what's happening as when I go to edit it the indents look like they are correct but then when I save they are not

brockfanning:
@LucyGwilliamAdmin I think there is a TAB character in there somewhere. Try removing all the whitespace and only using spaces.

LucyGwilliamAdmin:
Ah yes, that fixed it - thanks!

PR #505, opened 2020-02-20
Also have overall total in box

LucyGwilliamAdmin:


SavvasStephanides:
Is it best if leave styling changes outside the expert review for later?

brockfanning:
Sure, any style change can technically wait for a minor/hotfix release after 1.0.0.

I do agree with this change though, since right now it just looks like one sentence "Overall Reporting Status 244 Indicators".

Technical note - I wonder if it would still work using `.status-goal` instead of `h2, h3`, since I believe they both have that class.

SavvasStephanides:
Nitpicking here, (@phillipgardiner might be more suitable to help with this) but maybe the overall count could be slightly larger since it's part of the title?

image

phillipgardiner:
There are so many different options! However I believe this page probably has more wholesale changes planned for the future. Therefore I would just go with something that is consistent for now.

SavvasStephanides:
Thanks @phillipgardiner.

PR #506, opened 2020-02-20
Remove goal line from indicator banner

SavvasStephanides:
Removes the goal line from indicator page and changes the indicator title to `<h1>`

**Note**: I tried to align the title vertically, but haven't been able to in a way that would fit both short and long titles. Let's leave it at that at the moment and revisit later.

## Screenshot:

image

### DIFF
--- a/_layouts/indicator.html
+++ b/_layouts/indicator.html
@@ -14,12 +14,7 @@
         </a>
       </div>
       <div class="col-xs-8 col-md-9 col-lg-10">
-        <h1>
-          <a href="{{ page.goal.url }}">
-            <span class="hidden-sm hidden-md hidden-lg">{{ page.t.general.goal }} {{ page.goal.number }}: </span>{{ page.goal.name }}
-          </a>
-        </h1>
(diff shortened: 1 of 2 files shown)


brockfanning:
@SavvasStephanides Leaving the vertical centering for later sounds good to me.

One quick note on the screenshot - it seems like it needs some more line-height, doesn't it? To me the line-height doesn't match the font size.

SavvasStephanides:
> One quick note on the screenshot - it seems like it needs some more line-height, doesn't it? To me the line-height doesn't match the font size.

You're right

SavvasStephanides:
## Updated screenshot

image

brockfanning:
Cool, I like it. I'm ready to approve, but I'll let @phillipgardiner have a say.

phillipgardiner:
Looks good. My only comment is how come the indicator number is repeated? @SavvasStephanides @brockfanning

SavvasStephanides:
@phillipgardiner I think that's my mistake (it's UK site issue). That should be fixed when we update the UK site to 1.0 (I hope)

PR #508, opened 2020-02-20
Allow translation of indicator content

brockfanning:
This is to support the translation of the indicator content (in the metadata files, the stuff below the `---`).

This would allow something like this:

```
my_field: my value
my_other_field: my other value
---
indicator_descriptions.1-1-1
```

Then if there was a translation file called "indicator_descriptions.yml" containing an item for "1-1-1", then that translation would get used for the indicator content.

This is a quick fix to help with the fact that the "subfolder" approach to translation was removed recently. We should follow up with:
1. Tests and documentation for this functionality
2. Revisiting whether the subfolder approach could be supported

### DIFF
--- a/_includes/components/indicator-content.html
+++ b/_includes/components/indicator-content.html
@@ -1,3 +1,3 @@
 <div id="page-content">
-  {{ page.indicator.page_content | markdownify }}
+  {{ page.indicator.page_content | strip | t | markdownify }}
 </div>


brockfanning:
@LucyGwilliamAdmin Here's a quick PR to maybe give you a workaround for translating the indicator content.

LucyGwilliamAdmin:
@brockfanning is this still required now the sub-folder approach has been brought back?

brockfanning:
@LucyGwilliamAdmin Not strictly required, but for consistency I think it would still be a good thing to do.

LucyGwilliamAdmin:
@brockfanning so just tested this on [my test site](https://lucygwilliamadmin.github.io/sdg-site-quick-start/en/1-1-1/) and didn't seem to work

Might be an issue with [my yml file](https://github.com/LucyGwilliamAdmin/sdg-data-quick-start/blob/develop/translations/en/indicator_descriptions.yml) as think I've not done it properly due to the markdown and several lines

brockfanning:
Maybe try indenting the extra lines using 2 spaces each?

LucyGwilliamAdmin:
@brockfanning just tried indenting all the lines but it didn't work

LucyGwilliamAdmin:
@brockfanning just did another test of this and have sort of got it working - must have done something wrong before.

LucyGwilliamAdmin:
Just need to be more careful with the markdown

LucyGwilliamAdmin:
E.g. see indicator content [here](https://raw.githubusercontent.com/ONSdigital/sdg-data/develop/meta/3-4-2.md) compared with [here](https://raw.githubusercontent.com/LucyGwilliamAdmin/sdg-data-quick-start/develop/translations/ru/indicator_descriptions.yml) for same results ([here](https://sustainabledevelopment-uk.github.io/3-4-2/) and [here](https://lucygwilliamadmin.github.io/sdg-site-quick-start/1-1-1/)) and it's still not quite right (some extra indents in the section under the bullet points

PR #509, opened 2020-02-21
Remove a lot of unused images

brockfanning:
Most of the images we have under `assets/img` are not actually being used anywhere. As far as I can tell, all of these images can be removed without any consequences. Some testing would be in order. Also, maybe there are historical reasons for these images being there (which I'm not aware of).


brockfanning:
Added back one image (the SDG wheel) that I see now will be used in another PR: #486

PR #514, opened 2020-02-25
Updating styling of reporting status progress bars

SavvasStephanides:
### Completed / not started:

image

### Completed / in progress:

image

Contrast check for graphical objects:

image

### Completed / not applicable

image

*I'm only slightly concerned about text in front of the not-applicable pattern*


SavvasStephanides:
Hold off on reviewing for now. Just noticed that the styling affects the goal page.

brockfanning:
Meanwhile, the screenshots look good to me.

Since you have a concern about accessibility, it might be a good idea to add a "notapplicable" indicator to the tests, and make sure that it gets tested. To do this, you would add the metadata file in [this folder](https://github.com/open-sdg/open-sdg/tree/master/tests/data/meta), and then add the indicator's URL to [this config file](https://github.com/open-sdg/open-sdg/blob/master/tests/.pa11yci).

SavvasStephanides:
@brockfanning I think this is ready for review. We should handle high contrast in another ticket since it's a more general change.

brockfanning:
The github diff seems to have gotten confused. Could you list the files I should look at? Or maybe copy the files to a new branch off master and open a new PR. I’m not sure why this happens in GitHub sometimes.

SavvasStephanides:
@brockfanning I think that was my attempt at resolving merge conflicts. The files changed here are:

* `_reporting_status.scss`
* `_mixins.scss`
* `1-2-2.md` (new file)
* `.pa11yci` (although it seems to have disappeared)

brockfanning:
@SavvasStephanides Ok, I checked locally and it does look like things got messed up somehow, Git-wise. I would recommend creating a new branch off the latest master branch, and copying the changes files there, to create a new PR.

Also yeah, there is no more .pa11yci file, as of the merging of #523 this morning. Now there are three pa11yci files, in the `tests/accessibility` folder.

One thought occurred to me though: do we actually need to test the indicator page here? This change would only appear on the reporting status page and the goal 1 page, right? And we're already testing both of those pages. So maybe it's not necessary to make any pa11yci config changes -- just making the 'notapplicable' metadata change is enough. I could be missing something though.

SavvasStephanides:
You're right. I think I'll close this and push a new request.

PR #522, opened 2020-02-26
Fix for the edge case of local data with untranslated builds

brockfanning:
This resolves an issue with the edge case of a site using local data (instead of remote data) and untranslated builds. The problem was the the javascript variable `remoteDataBaseUrl` was just `/`, and this was getting concatenated with `/comb/etc...` to make: `//comb/etc...`. Browsers were interpreting that as `https://comb/etc...` was was causing those requests to fail.

### DIFF
--- a/_includes/scripts.html
+++ b/_includes/scripts.html
@@ -33,8 +33,13 @@ $(function() {
           $(obj).append($('<img />').attr('src', $(obj).data('img')));
       });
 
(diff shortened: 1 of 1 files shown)


brockfanning:
@SavvasStephanides This is probably a rare edge case, but because it's a show-stopping bug, and fairly easily fixed, I think we should resolve this before 1.0.0. Do you mind taking a look?

PR #523, opened 2020-02-27
Accessibility tests for mobile and high-contrast

brockfanning:
This adds accessibility testing for mobile and high-contrast. To get it to pass, it also includes a couple of high-contrast fixes (zip file download info, and indicator page headline legend).


brockfanning:
@LucyGwilliamAdmin @SavvasStephanides As mentioned in Slack this is not hard to do with our pa11y-ci setup. Since we want to be fully tested, accessibility-wise, before launching 1.0.0, I added this to the 1.0.0 launch project.

PR #524, opened 2020-02-29
Documentation and tests for the 'tags' metadata field

### LINKED ISSUE #495: Documentation and tests for tags functionality
Right now the 'tags' (metadata field) functionality is undocumented, and we don't have any tests for it. It shouldn't be hard to add both, and it should be done before 1.0.0 is released.

brockfanning:
Fixes #495


LucyGwilliamAdmin:
@brockfanning on this (as I guess it would also apply to the multiple graph titles functionality), how would this work for metadata csv input?

@brockfanning @AnnCorp Also do we need to also add documentation for csv input? Happy to put something together

brockfanning:
@LucyGwilliamAdmin At the moment, any metadata fields that needs some sort of "data structure" like this will not be possible with the CSV input. So that would mean that these types of fields still require YAML. This is also the case with the new `graph_limits` and `graph_units_without_headline` fields.

PR #525, opened 2020-02-29
Documentation of the chart types

brockfanning:
I don't believe we have any documentation of the chart types for data providers. Here is a quick page for describing the chart types.


brockfanning:
@LucyGwilliamAdmin Actually hold off on this - I'm going to add something about stacked bar charts, now that that got merged.

LucyGwilliamAdmin:
Ok, great - maybe a separate ticket but I'm also thinking this is something that could be added to the features page. 

Just a quick explanation that different graph types are available and then some links to the diff chart types on some of the diff platforms? Again happy to add this.

brockfanning:
@LucyGwilliamAdmin Just added info to the features section, plus info about stacked bars.

LucyGwilliamAdmin:
@brockfanning just trying to use the guidance to create a stacked bar chart on a feature branch and have a quick question: does graph_type need to be stacked_bar or does it remain as bar and then having the `graph_stacked_disaggregation` is what turns it into a stacked bar chart?

LucyGwilliamAdmin:
@brockfanning I can't seem to get this to work:
http://sdgdev-813006012.eu-west-1.elb.amazonaws.com/multi-chart-title-deploy/4-3-1/

I think I've changed the metadata file correctly:
https://github.com/ONSdigital/sdg-data/blob/multi-chart-titles/meta/4-3-1.md

brockfanning:
@LucyGwilliamAdmin To answer your first question - "stacked_bar" is a separate value for "graph_type". Eg, `graph_type: stacked_bar`. Does that make a difference for your other question?

brockfanning:
@LucyGwilliamAdmin I realized I left that part out of the documentation - I've just made an update to this branch.

LucyGwilliamAdmin:
@brockfanning just changed the graph type but still isn't working - Sex is still showing as disaggregation in sidebar

http://sdgdev-813006012.eu-west-1.elb.amazonaws.com/multi-chart-title-deploy/4-3-1/

brockfanning:
@LucyGwilliamAdmin I don't think you've set `graph_units_without_headline` and `data_start_values`, unless I'm missing it.

LucyGwilliamAdmin:
@brockfanning I did set those originally, but I'm just playing around with different combinations to see if I could get it to work. Also I was unsure of when to use `graph_units_without_headline` - is it when you have more than one unit type? and do you need to use it if your data has no headline data?

brockfanning:
@LucyGwilliamAdmin Good feedback, thanks. I just added some clarification. Does this help?

brockfanning:
@LucyGwilliamAdmin Another question - your note about `bar` vs `stacked_bar` has me wondering: do you think it would be better if there was only `bar`, and the `graph_stacked_disaggregation` metadata field is what makes them appear "stacked"? That is technically easy to do if you think it's preferable that way.

LucyGwilliamAdmin:
@brockfanning I've got a little closer to getting this working but still not quite:
http://sdgdev-813006012.eu-west-1.elb.amazonaws.com/multi-chart-title-deploy/4-3-1/

I guess we could just have `graph_stacked_disaggregation` - do you think it would be more confusing?

LucyGwilliamAdmin:
@brockfanning do you know what may be causing the issues I'm having here:
http://sdgdev-813006012.eu-west-1.elb.amazonaws.com/multi-chart-title-deploy/4-3-1/

Also do you think `graph_stacked_disaggregation` drop down should be removed from side bar?

LucyGwilliamAdmin:
Heres's the metadata file:
https://github.com/ONSdigital/sdg-data/blob/multi-chart-titles/meta/4-3-1.md

brockfanning:
@LucyGwilliamAdmin I'm getting that feeling that my implementation of stacked bars still needs work. It never got a real test until now, and I think you're catching its limitations.

Do you have any object to removing it from the code for now? Or at least, removing it from the documentation?

LucyGwilliamAdmin:
@brockfanning no I agree - we could remove it from documentation until we're happy with it

brockfanning:
@LucyGwilliamAdmin In taking one last look at the code, I believe I figured out the problem. I've pushed another commit to this branch in case it resolves the problem. If not, I'll remove the documentation about it and will keep working on it.

The things I changed in this latest commit:
1. No more "stacked_bar" type. Now it's just "bar". (I think this is probably simpler.)
2. No more "graph_units_without_headline" field. I believe this was unnecessary.

LucyGwilliamAdmin:
@brockfanning 

Ok so I'm getting [this](http://sdgdev-813006012.eu-west-1.elb.amazonaws.com/stacked-bar/4-3-1/) now (after manually selecting Male and Female, even though I'm setting them as starting values):

image

I don't think this is what I would expect to see.

I'm not sure if this is the correct type of data this would be used for so maybe not the best indicator to test on?

brockfanning:
@LucyGwilliamAdmin Cool. I will look into the starting values problem. (It should be starting with Female/Male selected, so something is going wrong there.)

As for the stacked bars, that's actually what I would expect, given the way it's coded. So, from a technical perspective that part appears to be working right.

But if the visualisation is not what you would want, then that is a bigger problem. What do you think would be the ideal visualisation in this case?

LucyGwilliamAdmin:
If this is how you are expecting it to look then I'm fine with it.

I think I'm just struggling to visualise it as don't think it's the right type of data a stacked bar chart would use.

For example, I think it would expect the height of the stacked bar to be the same height as the overall disaggregation, if that makes sense?

brockfanning:
I agree, and in fact that's how this initial feature request was described - all the bars being identical in height, but varying in the height of the parts of the bars (if that makes sense). See the second screenshot here: https://github.com/dawncomer/open-sdg-site-starter/issues/16

My approach with implementing it in Open SDG has been to focus on getting the basic functionality working, and hoping that the data can be controlled in a way to get that result. I just tweaked the test data in this branch, and it gives me these stacked bars in 2.3.1, which look a little better:

Screenshot from 2020-03-20 12-43-59

This makes me realize we may want to support graph_type per unit of measurement (similar to how to support graph_title per unit of measurement maybe?). Because the stacked bars may make sense for percentages, but definitely not for totals.

By the way, locally I'm not able to recreate the data_start_values problem. I'm trying with your branches, but I had to make a couple of changes:
* change the graph_type of 4.3.1 to "bar"
* point remote_theme to brockfanning/open-sdg@chart-documentation

With those changes I see this when I load 4.3.1:

Screenshot from 2020-03-20 12-16-11

PR #526, opened 2020-02-29
More detailed documentation on 'reporting_status_extra_fields'

### LINKED ISSUE #501: Improve documentation of reporting status extra fields
We could use more details on how to use reporting status extra fields. What we have now is: https://open-sdg.readthedocs.io/en/latest/reporting-status/

To start, I don't think we ever did this suggestion by @LucyGwilliamAdmin: 
https://github.com/open-sdg/open-sdg/pull/424#issuecomment-574067065

brockfanning:
Fixes #501


LucyGwilliamAdmin:
@brockfanning I'm just thinking is this different if you are using output instead of the new data config file?

and now I'm also thinking do we need some sort of documentation around this? Just because I think you said the new data config file can't be used with SDMX data input?

brockfanning:
@LucyGwilliamAdmin Let me think on a way for that data config file to support SDMX, or other inputs. Otherwise yes we could add some language here about how it would be different without the data config file.

brockfanning:
In YAML, these 2 are identical ways to represent lists:

```
reporting_status_extra_fields: ['un_custodian_agency']
```

```
reporting_status_extra_fields:
  - un_custodian_agency
```

But the only way to have an empty list is `[]`.

PR #537, opened 2020-03-04
Update configuration document with homepage banner

SavvasStephanides:


brockfanning:
@SavvasStephanides Looks good to me, but I meant to bring this up on the call: We actually have existing config options called `frontpage_heading` and `frontpage_instructions` (see [here](https://open-sdg.readthedocs.io/en/latest/configuration/#frontpage_heading)). For consistency, do you think we should change this new feature to `frontpage_introduction_banner` instead of `homepage_introduction_banner`?

SavvasStephanides:
@brockfanning Yes, I agree. That would keep things consistent.

SavvasStephanides:
@brockfanning I've changed this here first. It should be a small change to do to set the actual settings

PR #538, opened 2020-03-05
Chart title per unit of measurement

### LINKED ISSUE #531: Multiple graph titles - one per unit of measurement
In many cases I've seen the `graph_title` used to describe only one of multiple available units of measurement. Would it be helpful if the `graph_title` field could have multiple values - one for each unit of measurement?

brockfanning:
Fixes #531 

This is intended to have full backwards compatibility: it adds an optional alternative field called `graph_titles` (plural) with a more complex structure to allow a separate graph title per unit of measurement. Docs and tests are included.

While I was looking at the docs I also noticed some Markdown issues, so threw some unrelated fixes.


brockfanning:
@LucyGwilliamAdmin Would you be able to test this out using a feature branch maybe? Here is the structure of the `graph_title` field that you can try:

```
graph_title:
  - unit: Percent
    title: My graph title for percentages
  - unit: Total
    title: My graph title for totals
```

This would need to be done with YAML (as opposed to CSV or Excel). Also, the `title` can be a translation key.

LucyGwilliamAdmin:
@brockfanning just testing this on a feature branch of UK site and get [this error](https://travis-ci.org/github/ONSdigital/sdg-data/builds/661134084):
image

brockfanning:
@LucyGwilliamAdmin Ah, I forgot about validation and schema. This is going to need a tweak to sdg-build, and I've opened a PR about it: https://github.com/open-sdg/sdg-build/pull/120/files

brockfanning:
@LucyGwilliamAdmin On further thought I think this should be its own separate metadata field, like `graph_titles` (plural). Hopefully we can continue to support `graph_title` but allow `graph_titles` to override it. I'll add to this PR.

brockfanning:
@LucyGwilliamAdmin This is ready to review again, but you should use `graph_titles` instead of `graph_title`. You can actually use both, but `graph_titles` should override `graph_title`. I've updated the PR description and earlier comments.

PR #546, opened 2020-03-11
jQuery 3.4.1

### LINKED ISSUE #539: Update to jQuery 3.x
We're currently using jQuery 1.12.4 which is no longer receiving security patches. We should update to 3.x series which is the only series receiving security patches.

brockfanning:
Fixes #539


brockfanning:
@SavvasStephanides This is passing tests, so that's a good sign. Can you take a look?

brockfanning:
Git issues... I'll try this again.

PR #547, opened 2020-03-11
jQuery 3.4.1

### LINKED ISSUE #539: Update to jQuery 3.x

brockfanning:
Fixes #539


brockfanning:
@LucyGwilliamAdmin @SavvasStephanides This passes tests and appears fine in local manual testing. I'd say it's fine to do this before version 1. Can you review/approve if you agree?

SavvasStephanides:
Normally I wouldn't recommend upgrading libraries right before a major release unless absolutely necessary, but if you're 100% certain this doesn't break anything, then I'm fine with it.

PR #553, opened 2020-03-17
allow the use of user-defined colors

### LINKED ISSUE #326: More colors for bar charts
If there are more than 6 time series, the colours are repeated in the bar chart.

grafik

Is it possible to provide more colours?

MoBosse:
This should allow to define a set of colors in the site-repositories config-file to be used in the charts.
Fixes #326

### DIFF
--- a/_includes/assets/js/indicatorModel.js
+++ b/_includes/assets/js/indicatorModel.js
@@ -192,10 +192,7 @@ var indicatorModel = function (options) {
 
   var headlineColor = '777777';
   
-  // use custom colors if defined
-  //this.colorSet = {{ site.graph_color_set | jsonify }};
-  //his.numberOfColors = {{ site.graph_color_number | jsonify }};
-  //this.customColors = {{ site.graph_color_list | jsonify }};
+  // use custom colors
   var colors = opensdg.chartColors(this.indicatorId);
   
   // allow headline + (2 x others)
--- a/_includes/assets/js/indicatorModel.js
+++ b/_includes/assets/js/indicatorModel.js
@@ -193,10 +193,10 @@ var indicatorModel = function (options) {
   var headlineColor = '777777';
   
   // use custom colors if defined
-  this.colorSet = {{ site.graph_color_set | jsonify }};
-  this.numberOfColors = {{ site.graph_color_number | jsonify }};
-  this.customColors = {{ site.graph_color_list | jsonify }};
-  var colors = opensdg.chartColors(this.indicatorId, this.colorSet, this.numberOfColors, this.customColors);
+  //this.colorSet = {{ site.graph_color_set | jsonify }};
+  //his.numberOfColors = {{ site.graph_color_number | jsonify }};
+  //this.customColors = {{ site.graph_color_list | jsonify }};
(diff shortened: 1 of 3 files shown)


_includes/assets/js/indicatorModel.js
```diff
@@ -191,8 +191,13 @@ var indicatorModel = function (options) {
...
 
   var headlineColor = '777777';
-  var colors = ['7e984f', '8d73ca', 'aaa533', 'c65b8a', '4aac8d', 'c95f44'];
-
+  
+  // use custom colors if defined
+  this.colorSet = {{ site.graph_color_set | jsonify }};
+  this.numberOfColors = {{ site.graph_color_number | jsonify }};
+  this.customColors = {{ site.graph_color_list | jsonify }};
+  var colors = opensdg.chartColors(this.indicatorId, this.colorSet, this.numberOfColors, this.customColors);
```
brockfanning: I believe you can use `this.shortIndicatorId` instead of `this.indicatorId` and you won't have to worry about the "indicator_" being in the front of it. Not a huge deal, just FYI.

_includes/assets/js/indicatorModel.js
```diff
@@ -191,8 +191,13 @@ var indicatorModel = function (options) {
   }());
 
   var headlineColor = '777777';
-  var colors = ['7e984f', '8d73ca', 'aaa533', 'c65b8a', '4aac8d', 'c95f44'];
-
+  
+  // use custom colors if defined
+  this.colorSet = {{ site.graph_color_set | jsonify }};
+  this.numberOfColors = {{ site.graph_color_number | jsonify }};
+  this.customColors = {{ site.graph_color_list | jsonify }};
```
brockfanning: Since `colorSet`, `numberOfColors`, and `customColors` are not needed anywhere else in this script, what about moving them to `chartColors.js`?

MoBosse: Good point but how do I call the values from `config.yml` from `chartColors.js`? 
`this.colorSet = {{ site.graph_color_set | jsonify }}; ` seems not to work here.

MoBosse: @brockfanning ^^^^

brockfanning: @MoBosse I don't think there's anything special about indicatorModel.js, so if it works there then it should also work in chartColors.js. Maybe something like this at the top of the function?
```
var colorSet = {{ site.graph_color_set | jsonify }},
    numberOfColors = {{ site.graph_color_number | jsonify }},
    customColorList = {{ site.graph_color_list | jsonify }};
```

MoBosse: @brockfanning thats pretty much what I tried [here](https://github.com/Nachhaltige-Entwicklung-Deutschland/open-sdg-site-starter/blob/develop/assets/js/chartColors.js#L3) but it keeps giving me a console-message that says 

_SyntaxError: expected property name, got '{'_

MoBosse: I tested this in a Version of chartColors.js that is saved in the assets folder of the site-starter repo. In the includes folder it seems to work. Sorry for the confusion.

PR #557, opened 2020-03-18
docs for config options regarding customized colors

MoBosse:


brockfanning:
@MoBosse Thanks very much! I'm wondering if maybe we should add a disclaimer under graph_color_set:

```
> **NOTE**: Whatever color scheme you choose here, please ensure that all colors satisfy
> the accessibility (minimum contrast) standards in your region. These colors will need to
> be visible on white and black backgrounds.
```

brockfanning:
@AnnCorp Does the language above seem OK to you, as a disclaimer about color contrast?

brockfanning:
I'll add that disclaimer in a separate PR.

Thanks @MoBosse!

PR #560, opened 2020-03-19
Create uk.md

AnnCorp:
UK case study initial version.


LucyGwilliamAdmin:
@AnnCorp actually just looking at the .md file, I think where a new line is wanted, an extra line needs to be added in between

AnnCorp:
@LucyGwilliamAdmin yes, was debating whether to add more spacing or not - lots of the others block all the text. Think might merge and tweak later if (when) do next edit.

PR #563, opened 2020-03-19
Color contrast disclaimer

brockfanning:


brockfanning:
@AnnCorp This is a disclaimer I thought would be good to add to the documentation, under the recently-added color customisation functionality. What do you think?

brockfanning:
@AnnCorp Here's that disclaimer I mentioned in the call. Can you review and approve when you have time?

PR #565, opened 2020-03-23
Make the introduction banner work in high contrast

SavvasStephanides:
This sets the introduction banner text to white while in high contrast, fixing the bad contrast issue

### Screenshot:

Before:

image

After:

image


SavvasStephanides:
@brockfanning Good point!

PR #573, opened 2020-03-24
Make indicator links in goal page appear clickable

SavvasStephanides:
This change makes it clearer that the indicator cards on the goal page are actually clickable. This also sets the goal page layout to `goal-by-target`.

#### Default contrast

image

#### High contrast

image

#### Default contrast

image


brockfanning:
@SavvasStephanides Just so you know, in the automated tests the normal goal pages use the "goal" layout, but I've got a custom goal page using the goal-by-target layout. That custom goal page can be reached at /testing-goal-by-target/. (See [here](https://github.com/open-sdg/open-sdg/blob/master/tests/site/_goals/goal-by-target.md)). So if you want to make sure goal-by-target gets tested, we could add that path to the pa11y configurations.

Also a nitpick, instead of "blue" I would recommend #337ab7 - it's what we're using for link color, I believe it comes from Bootstrap.

brockfanning:
Also if I'm interpreting the test failure right, it doesn't like the tags' yellow background with the blue text? If it still fails with #337ab7 we might have to keep the tags' text black, or make the tags' yellow background fainter.

SavvasStephanides:
> if you want to make sure goal-by-target gets tested, we could add that path to the pa11y configurations.

Good idea.

> Also if I'm interpreting the test failure right, it doesn't like the tags' yellow background with the blue text? If it still fails with #337ab7 we might have to keep the tags' text black, or make the tags' yellow background fainter.

Possibly a darker colour will fix this. Maybe keep it black as it was?

SavvasStephanides:
A bit strange that the contrast tests are still failing on high contrast. Pretty sure everything is black on yellow or vice versa.

image

brockfanning:
@SavvasStephanides Check the Pa11y output - it says the errors on on the goal 1 page in the high contrast tests. I'd recommend doing a `make serve` and looking at goal 1 in high contrast.

SavvasStephanides:
@brockfanning That's what I did. That's where the screenshot above comes from.

EDIT: Turns out, the problem is in the `goal` layout of the goal page 🤦‍♂️

image

SavvasStephanides:
@brockfanning Tests pass now. Ready for re-review

brockfanning:
@SavvasStephanides A couple of pieces of feedback:
1. In high-contrast mode, the indicator status ("Reported online", etc.) is blue, but I think it should be black
2. Regarding the tags - I don't think this PR should change anything tag-related. I believe the expert-review feedback here was just about the indicator titles, so tags should stay black, on both normal and high contrast.

SavvasStephanides:
Thanks for the feedback @brockfanning 

> 1. In high-contrast mode, the indicator status ("Reported online", etc.) is blue, but I think it should be black

Good point

> 2. Regarding the tags - I don't think this PR should change anything tag-related. I believe the expert-review feedback here was just about the indicator titles, so tags should stay black, on both normal and high contrast.

That was the original styling but the tests kept failing and recommending I use the blue colour it is now. I'll revert back and see if it does it again.

brockfanning:
@SavvasStephanides Should I review again or is it still in progress?

SavvasStephanides:
@brockfanning This is ready for review.

SavvasStephanides:
@brockfanning Last change was just resolving merge conflicts

brockfanning:
@SavvasStephanides I'm still seeing blue text on the tags.

SavvasStephanides:
@brockfanning Maybe it reverted back when I tried to resolve the merge conflict. I'll have a look

brockfanning:
@SavvasStephanides To be more clear - I'm seeing blue text on the tags in normal contrast mode.

SavvasStephanides:
@brockfanning Ready for re-review. Should turn the tag colour to black.

PR #587, opened 2020-03-26
Remove table-hover class and styling

### LINKED ISSUE #580: There is no need to have the row turning grey when the mouse hovers over it in the indicator page.
Stop the table row from turning grey when the mouse hovers over it.

image

This also affects the high contrast version:

image

brockfanning:
Fixes #580


brockfanning:
@SavvasStephanides No problem, that's a much simpler change. PR updated.

Issue #222, opened 2019-07-01
'Contacts' tab


LucyGwilliamAdmin:
Suggestion from Kyrgyzstan:
Tab where we could display telephone numbers/email addresses

brockfanning:
I would be interested to hear more details about the proposal. If by "tab" it is meant, another item in the navigation menu for a contact page, then this can be easily done by creating a page, and then updating the [menu config](https://github.com/open-sdg/open-sdg-site-starter/blob/develop/_config.yml#L83). But if it's something else then I would like to hear more about it.

LucyGwilliamAdmin:
@brockfanning this is exactly what I meant - thanks for providing the info to add this

PR #593, opened 2020-03-27
High-contrast fixes for chart grids/ticks

### LINKED ISSUE #579: In high contrast version, the graph text is grey and hard to see.
Change axis text to white in high contrast

image

brockfanning:
Fixes #579 

This sets the following colors:

Normal contrast:
* Grid lines: `#ddd`
* Axis labels: `#000`

expert-review--chart-contrast--normal

High contrast:
* Grid lines: `#222`
* Axis labels: `#fff`

expert-review--chart-contrast--high

Let me know if I should tweak those colors.

Side notes:
1. This fixes a problem where we were setting the swatch for headlines to white, but it needs to stay grey.
2. This fixes a problem where the swatch hover color was bad in high-contrast mode
3. There are a few random white-space/indentation fixes in accessibility.js.
4. There is still an issue with the headline color, which right now is hardcoded to 777777 (see [here](https://github.com/open-sdg/open-sdg/blob/master/_includes/assets/js/indicatorModel.js#L192)). Really that should be white/black depending on the contrast mode, probably. But at least it appears to be meeting the accessibility minimums. I didn't change that here because it is a difficult problem would take some time to figure out.

### DIFF
--- a/_includes/assets/js/accessibility.js
+++ b/_includes/assets/js/accessibility.js
@@ -73,46 +73,55 @@ var accessibilitySwitcher = function() {
     }).html(getContrastToggleLabel(contrast).replace(" ", "<br/>")).click(function() {
       setActiveContrast($(this).data('contrast'));
       imageFix(contrast);
+      broadcastContrastChange(contrast, this);
     })));
   });
-  
-function getContrastToggleLabel(identifier){
-  var contrastType = "{{ site.contrast_type }}"
-  if(contrastType === "long") {
-    if(identifier === "default"){	
-      return translations.header.default_contrast; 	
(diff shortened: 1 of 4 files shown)


SavvasStephanides:
@brockfanning

> But at least it appears to be meeting the accessibility minimums.

I'm wondering whether pa11y can actually test the graphs since they are generated with Javascript

Looks great!

brockfanning:
Yes I believe pa11y uses Puppeteer or PhatomJS, or something similar, so it definitely runs javascript. In fact it has a screenshot feature where you can print screenshots, and see exactly what it's seeing.

PR #596, opened 2020-04-01
Update index.md

AnnCorp:
Highlighting more general pages from the 'home' page and shortening and amending some of the content.


AnnCorp:
@brockfanning Wasn't sure if should change the link to SDG Translations? Also maybe I should add in a full list of languages already available?

brockfanning:
@AnnCorp I liked what you have here - no need to change anything language-related as far as I'm concerned.

PR #597, opened 2020-04-01
Update faq.md

AnnCorp:
First pass through the FAQ page.


AnnCorp:
@brockfanning am wondering if should shorten the single vs double repo section and store that elsewhere but will do that in a separate edit.

PR #600, opened 2020-04-01
Only show items that can be selected in drop down lists

LucyGwilliamAdmin:
@AnnCorp @brockfanning @SavvasStephanides 

Sorry, just realised we haven't added this too project - but think we said we'd put it in to v1?

Before: https://sustainabledevelopment-uk.github.io/3-9-1/
After: http://sdgdev-813006012.eu-west-1.elb.amazonaws.com/multi-chart-title-deploy/3-9-1/

### DIFF
--- a/_sass/layouts/_indicator.scss
+++ b/_sass/layouts/_indicator.scss
@@ -231,6 +231,7 @@
           }
 
           &[data-has-data="false"] {
+            display: none;
             cursor: default;
             &:hover {
               background: white;


SavvasStephanides:
I remember discussing this, but I don't think we decided on whether to include this on v1.0

brockfanning:
It looks good to me. I couldn't think of any reason not to include it in 1.0.0.

PR #603, opened 2020-04-01
Move units above subcategories and promote to H4 instead of H5

### LINKED ISSUE #601: Move "Units" above "Sub categories"

brockfanning:
Fixes #601 

This is a simple change, but I also changed the "Units type" header from an H5 to an H4, because otherwise it would be smaller than "Sub-categories" (which is an H4). Here's a screenshot:

units-above-subcategories

### DIFF
--- a/_includes/components/units-template.html
+++ b/_includes/components/units-template.html
@@ -1,6 +1,6 @@
 <script type="text/template" id="units_template">
   <% if(units.length) { %>
-    <h5>{{ page.t.indicator.units_type }}</h5>
+    <h4>{{ page.t.indicator.units_type }}</h4>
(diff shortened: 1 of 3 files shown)


brockfanning:
@phillipgardiner Here's a change moving the Units above the Sub-categories.

phillipgardiner:
Seems logical given the current limitations.

SavvasStephanides:
Yes it does look more reasonable this way.

What do you think about leaving some more whitespace here?

image

brockfanning:
@SavvasStephanides Makes sense. Here is 20px of bottom margin:

units-bottom-margin

I'll push up the change too.

PR #604, opened 2020-04-01
Hide the sub-categories when there are no disaggregations

### LINKED ISSUE #602: Refresh sub-categories when Unit changes

brockfanning:
Fixes #602


brockfanning:
@SavvasStephanides @LucyGwilliamAdmin Here is a change to hide the sub-categories when necessary. This should resolve the problem on https://sdg-kyrgyzstan.github.io/open-sdg-site-starter/en/7-3-1-2/

LucyGwilliamAdmin:
Just had quick look at how Kytgyzstan's 7-3-1-2 would look:
https://sdg-kyrgyzstan.github.io/open-sdg-site-starter/7-3-1-2/

Do we still want "Click on the legend to remove individual lines from the chart." and "Clear selections" to show when they are no disaggregations?

brockfanning:
@LucyGwilliamAdmin Good points, no I don't think we need those to display. I've pushed up another commit to remove those. Also now this is purely a CSS fix.

PR #605, opened 2020-04-02
Add instructions for triggering site build

LucyGwilliamAdmin:
open-sdg-data-starter PR: https://github.com/open-sdg/open-sdg-data-starter/pull/29

### DIFF
--- a/docs/quick-start.md
+++ b/docs/quick-start.md
@@ -21,7 +21,7 @@ This document will go over the quickest way to get this platform up and running.
 1. As before, choose a name. This one should refer to "data" instead of "site" (eg, `sdg-data-australia`). As before, leave "Public" selected and click "Create repository from template".
     * Bookmark the created repository -- this is your "__data repository__".
 
-## Creating an access token
+## Creating access tokens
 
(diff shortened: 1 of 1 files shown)


brockfanning:
@LucyGwilliamAdmin It's very cool how simple it is, thanks! I just have a couple comments:
1. Are we sure that the additional token is needed? I'm just curious if they could use the same token they already created.
2. One concern is the effect this has on the initial quick start. Since in the quick start we recommend configuring the data repo first, I worry that this might cause a failure, since the site repo is not configured yet. Maybe it's not a problem though. Best case, that particular step fails, but it doesn't affect anything. Worst case, that particular step fails and it prevents the data repo from deploying. Do you have any ideas?

LucyGwilliamAdmin:
@brockfanning You're right - can use same token. Not sure why but I thought a separate one would need to be created.

I'll do check of quick test tomorrow

LucyGwilliamAdmin:
@brockfanning just checked the quick start issue - having the extra bit in the deploy-to-staging data repo workflow doesn't seem to cause an issue:

https://github.com/LucyGwilliamAdmin/qs-data/pull/2

PR #617, opened 2020-04-08
Highlight the high-contrast goal icons on hover

brockfanning:
captured (1)


brockfanning:
@phillipgardiner I don't think we can change the image color, but how about this highlight approach?

phillipgardiner:
@brockfanning Looks good to me.  Just need something that provides the user with visual feedback :)

PR #618, opened 2020-04-08
Font size 14px for long contrast switcher

brockfanning:
In case you want to see different sizes, here's an animation starting at 16px (the current size) and going  down to 15px, 14px, 13px, and 12px:

captured (2)

### DIFF
--- a/_sass/components/_high_contrast.scss
+++ b/_sass/components/_high_contrast.scss
@@ -137,6 +137,7 @@ body.long {
     a {
       margin: auto;
       text-decoration: none;
+      font-size: 14px;
     }
   }
 


brockfanning:
@phillipgardiner Here's a try at this.

phillipgardiner:
I think 14 anything smaller and it starts to get harder to read. But i'm open to opinions.

SavvasStephanides:
14px is a good size to go with although maybe it could go as low as 12px while still being intelligible (see below). We could maybe consider a label change instead so it isn't that long?

image

brockfanning:
@SavvasStephanides See that animation above - Phil recommended no lower than 14 based on that.

Regarding the label wording - I think that might be for a longer internal UK discussion - as I recall that came out of a UK accessibility audit.

phillipgardiner:
@SavvasStephanides @LucyGwilliamAdmin do you know the reason why we use the term 'high contrast' but Open SDG has the 'A' icons. The icons are more aesthetically pleasing but there must be a reason why we broke away from that?

brockfanning:
Here are the pull-requests where we discussed that feature - maybe there are some clues:
* #287 
* #325

LucyGwilliamAdmin:
@phillipgardiner I think it was advised by DAC as couldn't really tell what A A meant

SavvasStephanides:
@phillipgardiner DAC gave us a report which specified that it's better to use full text instead of "A". See issue here https://github.com/ONSdigital/sdg-indicators/issues/3087

The choice of words was completely ours, DAC's instruction was just "be more descriptive"

phillipgardiner:
@SavvasStephanides @LucyGwilliamAdmin Thanks, good to know. I'd say the current wording is pretty descriptive so leave as is for now. So just needs adjusting to 14px.

PR #620, opened 2020-04-08
Disclaimer redesign

brockfanning:
screenshot-127 0 0 1_4000-2020 04 08-14_36_56

screenshot-127 0 0 1_4000-2020 04 08-14_34_11

### DIFF
--- a/_includes/components/dev-disclaimer.html
+++ b/_includes/components/dev-disclaimer.html
@@ -1,6 +1,7 @@
 <div class="container">
   <div class="disclaimer-alert" role="alert">
-    <strong class="phase-tag">{{ page.t.header.alpha }}</strong>
-    {{ page.t.header.disclaimer | replace_first: '%email_address', site.email_contacts.suggestions }}
+    <strong class="phase-tag">{{ site.disclaimer.phase | default: page.t.header.alpha | t }}</strong>
+    {% assign disclaimer_message_default = page.t.header.disclaimer | replace_first: '%email_address', site.email_contacts.suggestions %}
+    {{ site.disclaimer.message | default: disclaimer_message_default | t }}
   </div>
 </div>
--- a/_includes/components/dev-disclaimer.html
+++ b/_includes/components/dev-disclaimer.html
@@ -1,5 +1,5 @@
 <div class="container">
-  <div class="alert alert-danger" role="alert">
+  <div class="disclaimer-alert" role="alert">
     <strong class="phase-tag">{{ page.t.header.alpha }}</strong>
     {{ page.t.header.disclaimer | replace_first: '%email_address', site.email_contacts.suggestions }}
   </div>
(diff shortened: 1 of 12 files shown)


brockfanning:
@phillipgardiner I've got screenshots above, for review. What do you think?

brockfanning:
@phillipgardiner Not sure which I like better, but just adding a screenshot here where the line doesn't go all the way across:

screenshot-127 0 0 1_4000-2020 04 08-15_57_24

phillipgardiner:
@brockfanning Thanks. I think it actually looks better with the line going all the way across. This is slightly different to what GDS propose, but in this case I think it is OK to diverse. So lets go with that. High contrast looks good too.
Could we please stick to the wording though: This is a development website – your [feedback](url) will help us to improve it. @AnnCorp could you advise will this release still be classed as a 'development website'?

AnnCorp:
@brockfanning @phillipgardiner agree re line going all way across and sticking with lets stick with previous wording for now please (including development website bit).

brockfanning:
@phillipgardiner @AnnCorp Ok, sounds good. I didn't change any wording, though -- what you're seeing in the screenshots is actually the [default from sdg-translations](https://github.com/open-sdg/sdg-translations/blob/master/translations/en/header.yml#L4). The UK site has [overridden that wording](https://github.com/ONSdigital/sdg-indicators/blob/develop/data/translations/en/header.yml#L1).

@SavvasStephanides This is ready for review.

phillipgardiner:
@brockfanning must be my confusion between the two versions again! The UK wording is what is recommended by GDS hence we have it there. I suppose it doesn't matter too much what everyone else puts as it can be overwritten and edited anyway?

brockfanning:
Yes, it is sort-of configurable, in that countries can override the translation file, like the UK has done.

SavvasStephanides:
Looks great. Should we leave the "ALPHA" label as blue on high contrast?

image

phillipgardiner:
@SavvasStephanides I believe we are probably OK too leave it. I think this would fall under the graphical object or large text category which seems to pass AA. Also the actual 'Alpha' text is readable. it would only be the blue on black that may fail. - Happy to be corrected though.

SavvasStephanides:
@phillipgardiner Thanks. I thought we'd set the colours in line with the rest of the high contrast colour scheme.

image

brockfanning:
I'm fine either way. I guess if we changed it, it would be black text on white background? That would look like this:

screenshot-127 0 0 1_4000-2020 04 09-11_42_11

brockfanning:
Ha, great minds... :)

brockfanning:
Ok, I've updated the PR with that change. I've also added a bit of padding beneath the "Alpha" - I'm not sure why but previously we had more padding on top than on bottom. I've also added a proper configuration option, so that it's easier for countries to control the text. Last I've added documentation and automated tests of the feature.

Here's how it looks in high-contrast now:

screenshot-127 0 0 1_4000-2020 04 09-16_33_31

@SavvasStephanides It's ready for review again.

phillipgardiner:
Brilliant looks great.

Issue #623, opened 2020-04-09
Reduce over-reliance on remote sources when running locally


SavvasStephanides:
Current when testing changes locally for SDG repos (including OpenSDG, UK platform etc) we seem to rely heavily on remote sources. Usually it's not that much of a problem, but if something happens, like for example the Internet connection is lost, or Github Pages goes down ([as it happened recently](https://www.githubstatus.com/incidents/f3xzj5fgfq5b)), development is pretty much crippled. 

Obviously this is a huge change and not something to happen soon. This is issue is mostly for discussion. I think it'd improve the developer experience immensely, not just for Open SDG, but for individual sites which use Open SDG globally.

Testing Open SDG relies on these Github repos, which are cloned for each test, or serve:

* `open-sdg/open-sdg-site-starter`
* `open-sdg/open-sdg-data-starter`

After that, the starter repos and UK site rely on remote sources:

* Remote data source - _this is a quite a bit of an issue since when testing locally (`http://localhost`), as the data cannot be displayed due to Github's CORS policy. The page is stuck on a loading animation_
* Translations from `sdg-translations`
* Images for goal tiles from `sdg-translations`
* Quite a few Javascript libraries that are currently loaded from remote CDNs (Bootstrap, jQuery, Leaflet, Font Awesome...)

It would be nice if we could reduce this over-reliance on remote sources as much as we can (I understand that some need to stay remote). I'm open to suggestions on how to do achieve this but here are a few of mine:

* Clone, build and run everything locally for testing. This could either mean building each and every repository on its own port (or directory) manually, or create a project which does this for you using `docker-compose`. Automatically set all files to point to the correct links when testing locally and deploying.
* Create a guide on how to prepare everything locally and set all links to point to localhost.
* Add JS libraries as part of the repos. Usually JS libraries come with a "minified" version, which is a lot smaller than the normal version.

brockfanning:
I agree this is a really good idea, even if it is done one bit at a time. I've been in spots where I wanted to do some development while on a hotspot, but I knew that just one `make serve` would max out my cell's data plan... Some thoughts on each item:

1. Local cloning of starter repos: I'm sure we can remove the need for the `make serve` to clone the starter repositories. I can look into that.
2.  Data source: I have been using the npm package "live-server" to do this. This is a bit manual, unfortunately, but is working in a pinch. The steps are:
    * Build the data locally (eg, `python scripts/build_data.py`)
    * Go into the built folder (eg, `cd _site`)
[...]

PR #624, opened 2020-04-10
Less downloading of files in local builds

brockfanning:
This addresses a few of the things mentioned in #623 
* Use local files instead of cloning the data/site starter repositories
* Cache a copy of sdg-translations, and use it for translations and goal icons
* Also removes a reference to a remote geojson file


SavvasStephanides:
Hi @brockfanning. Thanks for responding to my issue so quickly. My question is, will there be any difference in what commands should be used to serve a site is it the same as the README file?

brockfanning:
@SavvasStephanides Everything should be the same, yes. The fact that tests still pass is a good sign.

PR #639, opened 2020-04-16
Make the configuration documentation alphabetical

brockfanning:
This PR includes:
1. Make the configuration settings listed in alphabetical order
2. Instead of separate sections for optional/required, indicate optional/required at the beginning of descriptions.

### DIFF
--- a/docs/configuration.md
+++ b/docs/configuration.md
@@ -8,7 +8,7 @@ _Note about "strings": Many of the settings detailed here contain human-readable
 
 ### analytics
 
-This optional setting can contain another (indented) setting, `ga_prod`, which should be a [Google Analytics tracking ID](https://support.google.com/analytics/answer/1008080?hl=en#GAID). If these settings are used, usage statistics will be sent to Google Analytics. For more information about this, see the [analytics](analytics.md) page.
+_Optional_: This setting can contain another (indented) setting, `ga_prod`, which should be a [Google Analytics tracking ID](https://support.google.com/analytics/answer/1008080?hl=en#GAID). If these settings are used, usage statistics will be sent to Google Analytics. For more information about this, see the [analytics](analytics.md) page.
 
 ```nohighlight
 analytics:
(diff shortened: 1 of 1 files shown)


SavvasStephanides:
Consideration: should these configurations really be laid out alphabetically, or should they be in the order of most important/popular?

brockfanning:
@SavvasStephanides My opinion is that, as a reference document, the list needs to be geared towards looking things up. I use it myself quite a bit, and always wish that it was alphabetical so that I could scan right to what I'm looking for, instead of having to scan the whole list each time.

SavvasStephanides:
@brockfanning Cool. Fair enough then!

LucyGwilliamAdmin:
@brockfanning I'm just looking at this now - just have a question about one of the fields `graph_color_number` - does this correspond to the maximum number of lines allowed on a graph before the warning displays? i.e:
image

LucyGwilliamAdmin:
@brockfanning also a little confused about `non_global_metadata` - couldn't the same result be achieved by using the `metadata_tabs` field?

brockfanning:
> @brockfanning I'm just looking at this now - just have a question about one of the fields `graph_color_number` - does this correspond to the maximum number of lines allowed on a graph before the warning displays? i.e:

Yes, that's my understanding.

> @brockfanning also a little confused about `non_global_metadata` - couldn't the same result be achieved by using the `metadata_tabs` field?

Yes I think that's true. I guess we implemented `non_global_metadata` first and then later implemented `metadata_tabs`.

LucyGwilliamAdmin:
@brockfanning thanks for the clarification.
re. `graph_color_number` do you know what the default maximum number is - just thinking we could include it here.

Could `non_global_metadata` be removed or do we need to keep it for backwards compatability?

brockfanning:
@LucyGwilliamAdmin Maybe we leave it in but note that it's deprecated? I just pushed up a commit with some clarifications on `non_global_metadata` and `graph_color_number`. What do you think?

PR #642, opened 2020-04-16
Use 'indicator_number' instead of 'indicator'

brockfanning:


brockfanning:
@LucyGwilliamAdmin This PR is mainly just evidence that "indicator_number" works the same as "indicator". We have backwards-compatibility code in place to make sure that "indicator" still works, but "indicator_number" is what we would prefer that countries use.

LucyGwilliamAdmin:
Ah, ok - looks good. The only thing I can see is that the prose file should be changed in the starter repos

brockfanning:
@LucyGwilliamAdmin Good catch - I just updated those files.

PR #646, opened 2020-04-18
Indicator page refactor with more components

brockfanning:
This PR doesn't change any functionality, but splits up the indicator page into a lot of components, for the purposes of:
1. Easier development
2. Allow users to override only particular parts of the indicator page

The specific changes are:
1. Turn `indicator.html` into a very simple layout that just includes lots of other files
2. Create an `_includes/components/indicator` folder to store all indicator-related include files
3. Move a lot of existing include files into that folder. Specifically:
    * `components/indicator-content.html` -> `components/indicator/indicator-content.html`
    * `components/edit-buttons.html` -> `components/indicator/edit-buttons.html`
    * `components/fields-template.html` -> `components/indicator/fields-template.html`
    * `components/headline.html` -> `components/indicator/headline.html`
    * `components/metadata.html` -> `components/indicator/metadata.html`
    * `components/data-notice.html` -> `components/indicator/data-notice.html`
    * `components/sources.html` -> `components/indicator/sources.html`
    * `components/units-template.html` -> `components/indicator/units-template.html`
4. Create some new include files with the markup removed from `indicator.html`. Specifically:
    * `components/indicator/breadcrumbs.html`
    * `components/indicator/data-main.html`
    * `components/indicator/data-sidebar.html`
    * `components/indicator/data-section.html`
    * `components/indicator/embedded-feature.html`
    * `components/indicator/header.html`
    * `components/indicator/indicator-available.html`
    * `components/indicator/metadata-section.html`


brockfanning:
@SavvasStephanides I know it's a bit late to propose this, but what do you think about merge this in before 1.0.0? This started as an attempt to refactor the indicatorModel.js and indicatorView.js files, but I fell down a rabbit-hole and ended up refactoring the indicator.html layout instead. The reason I'm interested in getting this into 1.0.0 is that I think it would be a bit too much of a breaking change to go into a minor release (eg, 1.1.0), so it would probably have to wait until 2.0.0 if we don't do it now.

brockfanning:
Just to explain the subsequent commits: I decided to revert a few unnecessary changes, so that the PR is limited to splitting up the indicator.html layout, and shuffling around the related include files.

SavvasStephanides:
Hi @brockfanning. Can you explain further what these "breaking changes" are? Is it a matter of countries might have overridden that particular file?

Either way, may I propose this be a minor release on its own? I think v. 1.0.0 has quite a few breaking changes as is. Maybe a `1.0.1` release which only has your refactoring changes?

From a quick look, it looks like it still works as normal and the tests pass, which is positive.

brockfanning:
@SavvasStephanides Yes that is the only breaking change - if a country has overridden that layout, or any of the include files that I moved, then they would experience this as a breaking change.

The general rule with semantic versioning is that:
1. A hotfix upgrade (eg, 1.0.0 to 1.0.1) should never have any breaking changes. In other words, countries should be able to do hotfix upgrades with 100% confidence and zero testing.
2. A minor upgrade (eg, 1.0.0 to 1.1.0) should only have breaking changes unless absolutely necessary, and if so, only with clear and easy upgrade instructions. Countries should be able perform minor upgrades without any extra work, but testing is recommended.
3. A major upgrade (eg, 1.0.0 to 2.0.0) can be reasonably expected to have breaking changes. Countries should expect to do some extra work to perform major upgrades.

So this 1.0.0 release (which is a "major" upgrade) is an appropriate time for any breaking changes - this is the time when that would be expected. If we introduce breaking changes in 1.1.0 (a "minor" upgrade), that's not the end of the world, but it is a practice that we would like to avoid. And we definitely wouldn't want to do breaking changes in 1.0.1 ("hotfix" upgrade).

SavvasStephanides:
@brockfanning I see your point and I agree.

I am still worried that we will be adding to the big pile of breaking changes in `1.0`. But it does make sense that we put in a major release and `1.0` seems to be our best bet for this.

brockfanning:
@SavvasStephanides On the bright side, we've covered all the bases with this statement here: https://open-sdg.readthedocs.io/en/latest/upgrades/upgrading-1-0-0/#custom-includes-and-layouts

My one regret is thinking of this after we tagged 1.0.0-beta last week. But we knew there might be additional changes after the tag, so I'm OK with merging this and going to 1.0.0-beta2.

SavvasStephanides:
@brockfanning I had a look a the repositories for every country mentioned in the docs as case studies and only [Germany](https://github.com/G205SDGs/sdg-indicators/blob/develop/_layouts/indicator.html) seems to have overridden the `indicator.html` page. The rest don't seem to have overridden the file:

[Armenia](https://github.com/armstat/sdg-site-armenia/tree/develop/_layouts) - [Los Angeles](https://github.com/CityOfLosAngeles/open-sdg-site-starter/tree/S3Dev/_layouts) - [Kyrgyzstan](https://github.com/sdg-kyrgyzstan/open-sdg-site-starter/tree/develop/_layouts) - [Rwanda](https://github.com/sustainabledevelopment-rwanda/sdg-indicators/tree/develop/_layouts) - [UK](https://github.com/ONSdigital/sdg-indicators/tree/develop/_layouts)

Seems it will only be a breaking change for Germany, but I'm sure they'll get it up and running and we can help them should they need it.

brockfanning:
Ok, good to know.

_includes/components/indicator/data-main.html
```diff
@@ -0,0 +1,50 @@
...
+        </li>
+        <li role="presentation" class="nav-item map" style="display:none">
+          <a data-no-disagg="true" class="nav-link" data-toggle="tab" href="#mapview" aria-controls="mapview" role="tab" {% include autotrack.html preset="tab_data_map" category="Tab change" action="Change data view" label="Change to Map tab" %}>{{ page.t.indicator.map }}</a>
+        </li>
+        {%- if page.indicator.embedded_feature_url -%}
+        <li role="presentation" class="nav-item embedded-map">
+          <a data-no-disagg="true" class="nav-link" data-toggle="tab" href="#embeddedmapview" aria-controls="embeddedmapview" role="tab" {% include autotrack.html preset="tab_data_embed" category="Tab change" action="Change data view" label="Change to embedded item tab" %}>{{ page.indicator.embedded_feature_tab_title }}</a>
+        </li>
+        {%- endif -%}
+      </ul>
```
SavvasStephanides, line 25: The tabs could live in their own file such as `data-tabs.html`.

_includes/components/indicator/data-main.html
```diff
@@ -0,0 +1,50 @@
...
+            <img src="{{ site.baseurl }}/assets/img/loading.gif" alt="{{ page.t.indicator.loading_map }}" />
+          </div>
+        </div>
+        {% if page.indicator.embedded_feature_url %}
+        <div role="tabpanel" class="tab-pane" id="embeddedmapview" class="embedded-map">
+          {% include components/indicator/embedded-feature.html %}
+        </div>
+        {% endif %}
+      </div>
+    </div>
```
SavvasStephanides, line 46: Also true for tab panes (and possibly each tab pane separately?)

_includes/components/indicator/metadata-section.html
```diff
@@ -0,0 +1,81 @@
...
+    <a class="nav-link" data-toggle="tab" href="#global" aria-controls="global" role="tab" {% include autotrack.html preset="tab_meta_global" category="Tab change" action="Change metadata view" label="Change to Global Metadata tab" %}>{{ page.t.indicator.global_metadata }}</a>
+  </li>
+  <li role="presentation" class="nav-item">
+    <a class="nav-link" data-toggle="tab" href="#sources" aria-controls="sources" role="tab" {% include autotrack.html preset="tab_meta_sources" category="Tab change" action="Change metadata view" label="Change to Sources tab" %}>{{ page.t.indicator.sources }}</a>
+  </li>
+{% endif %}
+  {% if site.environment == 'staging' %}
+  <li role="presentation" class="nav-item">
+    <a class="nav-link" data-toggle="tab" href="#edit" aria-controls="edit" role="tab">{{ page.t.indicator.edit }}</a>
+  </li>
```
SavvasStephanides, line 26: Same as data tabs. These could live in their own file

brockfanning:
@SavvasStephanides Good idea about splitting out those tabs/panes. This is ready for review again.

brockfanning:
I also noticed that this was causing the metadata section to be full-width, instead of staying in the main column. I actually kind of liked it, but I think it's not the right time for that kind of change, so the above commit puts the metadata back into the main column only.

PR #652, opened 2020-04-26
indicatorModel.js refactor

brockfanning:
This PR is a major refactor (and in some cases re-write) of indicatorModel.js. It's a work in progress, but I'm putting it up since tests are passing. The general goal of this is to make the code easier to understand, so that it becomes easier to fix bugs and add features. A few notes:

1. As much as possible, I tried to move chunks of code into helper functions in a separate file.
2. As much as possible, I tried to use ES5's array methods instead of Underscore.js. (Nothing against Underscore.js, but I thought it might be better to use standard javascript if possible, so more developers can work with it.)
3. One functionality improvement: the "starting values" / "missing headline" functionality, where an indicator starts out with pre-selected values if needed, has been improved - it no longer relies on the View to "click" the starting values; now the View starts out with the values already checked.
4. This included a general renaming of "series" stuff to "field" stuff, to reduce confusion an upcoming series-related feature.

Still to do:
* [x] Still better understanding of the View/Model flows and possibly more refactoring
* [x] Improve names of helper functions as needed and possibly more refactoring
* [x] Possible to split helper functions into separate files?
* [x] A bit more replacement of Underscore code
* [ ] Tons of manual testing
* [x] Documentation and comments as needed


brockfanning:
@SavvasStephanides Let me know if you have any feedback so far, either on the code, the approach, or from manual testing.

_includes/assets/js/indicatorModelHelpers.js
```diff
@@ -0,0 +1,539 @@
...
+      if (!keys.length) {
+        return [];
+      }
+
+      return data[keys[0]].map(function(item, index) {
+        return _.object(keys, keys.map(function(key) {
+          return data[key][index];
+        }));
+      });
+    },
```
SavvasStephanides, line 18: It is unclear what the purpose of this function is from the name and content.

1. Give the function a more meaningful name (`convertXToJson` or `convertJsonToX` depending on purpose) so it knows what it returns and what it expects as input.
2. `!key.length` could be changed to a more explicit `keys.length === 0` to imply "key array is empty".

_includes/assets/js/indicatorModelHelpers.js
```diff
@@ -0,0 +1,539 @@
...
+    extractUnique: function(prop, data) {
+      return _.chain(data).pluck(prop).uniq().filter(function(f) { return f; }).sortBy(function(year) {
```
SavvasStephanides, line 20: `extractUnique` isn't clear what it returns (unique what?). Also better to prefix the function with `get...` since it returns something directly.

_includes/assets/js/indicatorModelHelpers.js
```diff
@@ -0,0 +1,539 @@
...
+    dataHasGeoCodes: function(data) {
+      return this.dataHasColumn(this.GEOCODE_COLUMN, data);
+    },
+    getFirstUnitInData: function(data) {
+      return data.find(function(row) {
+        return row[this.UNIT_COLUMN];
+      }, this)[this.UNIT_COLUMN];
+    },
+    getDataByUnit: function(data, unit) {
+      if (!unit) {
```
SavvasStephanides, line 42: Is `!unit` supposed to check if the `unit` parameter is empty (ie. not provided)? If yes, then the function shouldn't return the `data` if `unit` is not provided, it should throw an error instead since `unit` is essential for the function

_includes/assets/js/indicatorModelHelpers.js
```diff
@@ -0,0 +1,539 @@
...
+      });
+      return Object.keys(valuesByField).map(function(field) {
+        return {
+          field: field,
+          values: valuesByField[field],
+        };
+      });
+    },
+    selectMinimumStartingFields: function(data, fieldNames) {
+      // We filter our full dataset to only those fields.
```
SavvasStephanides, line 94: Give priority in making your code readable instead of writing comments to explain it.

_includes/assets/js/indicatorModelHelpers.js
```diff
@@ -0,0 +1,539 @@
...
+        });
+      });
+      // But actually we want the top-priority sort to be the "size" of the
+      // rows. In other words we want the row with the fewest number of fields.
+      filteredData = filteredData.sort(function(a, b) {
+        return Object.keys(a).length - Object.keys(b).length;
+      });
+      // Convert to an array of objects with 'field' and 'values' keys, omitting
+      // any special columns.
+      var specialColumns = this.nonFieldColumns();
```
SavvasStephanides, line 114: Stick with one naming for these types of columns. "Special columns" or "non field columns"

_includes/assets/js/indicatorModelHelpers.js
```diff
@@ -0,0 +1,539 @@
...
+        this.VALUE_COLUMN,
+        this.UNIT_COLUMN,
+        this.GEOCODE_COLUMN,
+        'Observation status',
+        'Unit multiplier',
+        'Unit measure',
+      ];
+    },
+    getFieldColumnsFromArray: function(arr) {
+      var omit = this.nonFieldColumns();
```
SavvasStephanides, line 136: Rename `omit` to `columnsToOmit` to make it clearer that it's an array of columns that will later be filtered out.

_includes/assets/js/indicatorModelHelpers.js
```diff
@@ -0,0 +1,539 @@
...
+      return arr.filter(function(col) { return !omit.includes(col); });
+    },
+    getFieldColumnsFromData: function(data) {
```
SavvasStephanides, line 139: Unsure what the difference is between "columns" and "field columns"

brockfanning:
Thanks @SavvasStephanides, I'll make those updates.

Regarding that 500+ line wrapper function, that only serves to return an object containing the actual helper functions. This is so that it can be used in indicatorModel.js as sort of a poor-man's "import", in [this line](https://github.com/brockfanning/open-sdg/blob/indicator-mvc-refresh/_includes/assets/js/indicatorModel.js#L3).

I'm also to going to go through the helper functions and identify the ones that can be internal-only. Those don't technically need to be part of the returned object, and also could be easily moved into separate files.

brockfanning:
@SavvasStephanides I'm at another stopping place if you have any more feedback. I also think manual testing is going to be critical for this, because our automated tests don't really cover all the visual stuff that this might break.

brockfanning:
@SavvasStephanides Just a note that I added docblocks to explain the parameters, and tweaked the parameter names to make them more consistent.

_includes/assets/js/indicatorModelChartHelpers.js
```diff
@@ -0,0 +1,175 @@
...
+ */
+function getBackground(datasetIndex, colors) {
+  var color = getBackgroundColor(datasetIndex, colors);
+
+  if (datasetIndex > colors.length) {
+    color = getBackgroundPattern(color);
+  }
+
+  return color;
+}
```
SavvasStephanides, line 112: It is good practice to have the called functions below the function calling them. It's called the Newspaper approach. Look for "newspaper metaphor" [on the Clean Code book](https://www.investigatii.md/uploads/resurse/Clean_Code.pdf). For example:

```js
function getBackground(){
    if(datasetIndex > colors.length){
        var color = getBackgroundPattern();
    else{
        var color = getBackgroundColor();
    }
}

function getBackgroundColor(){
}

function getBackgroundPattern(){
}
```

brockfanning:
@SavvasStephanides I believe I got all those items updated. I tried to do the thing you suggest, of returning the object first and then declaring the functions, but I couldn't get that to work. But instead I've now got all the functions in separate files, so all you really see is the returned object in that main file.

Also I've moved all these files into their own folder.

SavvasStephanides:
@brockfanning That's fine. It was just a small tip to help you in future refactorings.

brockfanning:
@SavvasStephanides @AnnCorp @LucyGwilliamAdmin I believe we're now planning for this to be included in 1.0.0, if testing goes well, and that we should include this in the UK's upcoming rounds of manual testing. Do you prefer that we leave this as a PR, and do the UK testing against my branch? Or should we merge this to master, so the UK testing can be against master?

AnnCorp:
>Or should we merge this to master..
Seems to make sense to me to merge to master if we are planning to include in the release, but what do you think @LucyGwilliamAdmin @SavvasStephanides

LucyGwilliamAdmin:
Agree - think we should merge to master

Issue #651, opened 2020-04-23
Adjust heading structure across site


phillipgardiner:
The hierarchy of headings across the whole site needs to be
considered to ensure visual distinction and consistency

phillipgardiner:
@SavvasStephanides how many H levels do we need? H1, H2, H3, H4????

From what I can work out based on GDS guidelines, https://design-system.service.gov.uk/styles/typography/

I would suggest:
H1 = 36px
H2 = 24px
H3 = 19px
Normal copy text 15px
We may also need to introduce a lead copy text size, something like 17 or 18px for text at the start of the goals pages or the intro text on the homepage.

To be honest need to see it all side by side to be able to be 100% confident.

phillipgardiner:
@phillipgardiner The lowest I've seen is H5 on the Updates page, but I think these should change. It jumps from H2 straight to H5.

phillipgardiner:
@SavvasStephanides let me know if you need any more information. I believe that the H5 on the Updates page would actually be a H3 = 19px in our new structure.

SavvasStephanides:
Hi @phillipgardiner, I need your opinion on something:

On the indicator page the H1 heading is currently set at `24px`

![image](https://user-images.githubusercontent.com/478770/80395023-9262fc80-88aa-11ea-83a5-53af0b25f4d5.png)

If I change that to 36px as you suggested, it looks a little too large (especially is the indicator name is big, which often tends to be):

![image](https://user-images.githubusercontent.com/478770/80395236-d1914d80-88aa-11ea-8f30-dc204dfc21d8.png)

Thoughts?

phillipgardiner:
@SavvasStephanides is the example above using our new font? It may be worth doing that card first so we can see it in context. I see what you mean above, lets discuss in our call later :)

phillipgardiner:
@SavvasStephanides update the sizing in my original comment a bit but still need to discuss

SavvasStephanides:
@phillipgardiner 

> is the example above using our new font? 

The screenshot uses the "Open Sans". Here's what it looks like with Helvetica:

![image](https://user-images.githubusercontent.com/478770/80486458-1cff3680-8953-11ea-8a52-b2ec4d127b81.png)

and Arial:

![image](https://user-images.githubusercontent.com/478770/80486529-3902d800-8953-11ea-950c-2b719f2f076b.png)

phillipgardiner:
[...]

PR #659, opened 2020-05-01
Heading styling / font change

SavvasStephanides:
## Changes

- [x] Heading structure
- [x] Indicator banner changes
- [x] Set font to Helvetica, Arial
- [x] Move goal icon on top at indicator banner for mobile

All these changes are a result of the expert review by @phillipgardiner. 

### Desktop

image

image

image

### Mobile

image

image

image

*The indicator page banner layout change discussed [here](https://github.com/open-sdg/open-sdg/issues/651#issuecomment-622368367) will be done in a different PR*

### DIFF
--- a/_includes/components/indicator/header.html
+++ b/_includes/components/indicator/header.html
@@ -1,15 +1,15 @@
 <div class="heading indicator goal-{{ page.goal.number }}">
   <div class="container">
     <div class="row">
-      <div class="col-xs-4 col-md-3 col-lg-2 goalIcon">
+      <div class="col-xs-4 col-md-3 col-lg-2 goal-icon">
         <a href="{{ page.goal.url }}" title="{{ page.t.indicator.view_indicator_list }}">
           <img src="{{ page.goal.icon }}" alt="{{ page.goal.short | escape }} - {{ page.t.general.goal }} {{ page.goal.number }}" />
         </a>
       </div>
-      <div class="col-xs-8 col-md-9 col-lg-10 indicatorDetails">
+      <div class="col-xs-8 col-md-9 col-lg-10 indicator-details">
         <div>
           <h1>{{ page.t.general.indicator }} {{ page.indicator.number }}</h1>
-          <p class="leadCopy">{{ page.indicator.name }}</p>
+          <p class="lead-copy">{{ page.indicator.name }}</p>
         </div>
       </div>
     </div>
--- a/_includes/components/indicator/header.html
+++ b/_includes/components/indicator/header.html
@@ -7,7 +7,10 @@
         </a>
       </div>
       <div class="col-xs-8 col-md-9 col-lg-10">
-        <h1>{{ page.t.general.indicator }} {{ page.indicator.number }}: {{ page.indicator.name }}</h1>
+        <div aria-labelledby="headerIndicatorNumber headerIndicatorName">
+          <h1 id="headerIndicatorNumber" aria-hidden="true">{{ page.t.general.indicator }} {{ page.indicator.number }}</h1>
+          <p class="leadCopy" id="headerIndicatorName" aria-hidden="true">{{ page.indicator.name }}</p>
+        </div>
       </div>
     </div>
   </div>
--- a/_includes/components/indicator/header.html
+++ b/_includes/components/indicator/header.html
@@ -1,12 +1,12 @@
 <div class="heading indicator goal-{{ page.goal.number }}">
   <div class="container">
     <div class="row">
-      <div class="col-xs-4 col-md-3 col-lg-2">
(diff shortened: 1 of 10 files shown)


_sass/base/_typography.scss
```diff
@@ -1,12 +1,13 @@
-h1,
-h2,
-h3,
-h4,
-h5,
-h6 {
-  font-family: 'Open Sans', sans-serif;
+*{
```
brockfanning, line 1: I believe it's better to use "body" instead of "*", to make it easier to override this rule for specific descendants.

SavvasStephanides: Fair point

SavvasStephanides:
@brockfanning This should be ready for review. It was giving me some weird test failures but now they seem to have disappeared.

_includes/components/indicator/header.html
```diff
@@ -7,7 +7,10 @@
         </a>
       </div>
       <div class="col-xs-8 col-md-9 col-lg-10">
-        <h1>{{ page.t.general.indicator }} {{ page.indicator.number }}: {{ page.indicator.name }}</h1>
+        <div aria-labelledby="headerIndicatorNumber headerIndicatorName">
```
brockfanning, line 10: I'm unsure about about the ARIA usage here. Can you describe what the goal is? My understanding is that `aria-labelledby` should be used for associating widget elements with their labels. But this doesn't seem to fall into that category.

_sass/layouts/_search.scss
```diff
@@ -5,9 +5,9 @@
 }
 
 .layout-search {
-  h1 {
-    font-size: 24px;
-  }
+  // h1 {
+  //   font-size: 24px;
+  // }
```
brockfanning, line 10: A general preference of mine is to avoid leaving in commented code, unless you're sure you want to switch back at some point in the future.

_includes/components/indicator/header.html
SavvasStephanides: This is something I have described [here](https://github.com/open-sdg/open-sdg/issues/651#issuecomment-620657544) (second paragraph onwards).

> My understanding is that aria-labelledby should be used for associating widget elements with their labels.

Not exclusively. You can also use it for giving the content of a tag as a label to another. 

The point of this is for the screen readers to read the title tag ("Indicator X.X.X") and the paragraph below (the indicator full name) as one element. This would make it easier for screen reader users to navigate the indicator banner.

_sass/layouts/_search.scss
SavvasStephanides: Very good point. I'm removing them.

SavvasStephanides:
> Is this PR tied to #651 - ie does it need to reflect all the decisions there?

Yes, hopefully. If there's anything I left out let me know.

brockfanning:
I was just wondering about the change with the goal icon placement on mobile. Though now that I type that out, it does seem like a separate thing.

SavvasStephanides:
Ah fair enough. Yes, there might be a bit more discussion before we implement that change.

_includes/components/indicator/header.html
brockfanning: I'm not sure if it's accomplishing the goal. I tried it out with Orca on Ubuntu - if I do Alt-Shift-H to display a list of headings, it doesn't display this heading at all. If I navigate the headings (using H and Shift-H) it similarly skips that heading. Is it maybe because of the `aria-hidden` attributes?

Not sure Orca is standard enough, but I would assume so. Which screen reader do you use for testing?

SavvasStephanides:
@brockfanning This should be ready for re-review. I have also changed the indicator banner for mobile and it should now look like this:

image

brockfanning:
@SavvasStephanides It looks good to me, but a couple comments:
1. There is still a bit of commented code.
2. So far in this project we've used mostly dashes in our CSS, and not so much camel-casing. For example "search-btn" instead of "searchBtn". If you think we should start switching to camel-casing I don't have a preference, but regardless I think we should aim for a standard. What do you think?

SavvasStephanides:
@brockfanning 
 
> 1. There is still a bit of commented code.

I've spotted [this](https://github.com/open-sdg/open-sdg/blob/headings_styling/_sass/layouts/_reporting_status.scss#L73) which I deleted. Any other instances?

> 2. So far in this project we've used mostly dashes in our CSS, and not so much camel-casing. For example "search-btn" instead of "searchBtn". If you think we should start switching to camel-casing I don't have a preference, but regardless I think we should aim for a standard. What do you think?

Let's stick with dashes (kebab-case). I'll change my files.

SavvasStephanides:
@brockfanning Done the latest changes (removed commented out code, changed CSS class casing, added Open Sans in font family). Ready for re-review

PR #662, opened 2020-05-05
FAQ: examples of GeoJSON/CSV paths

brockfanning:
This adds examples for the CSV and GeoJSON download question. And for consistency in formatting, I moved the examples for the API question to their own line.

### DIFF
--- a/docs/faq.md
+++ b/docs/faq.md
@@ -54,15 +54,27 @@ Open SDG can use an SDMX API endpoint or can use data and metadata in SDMX-ML fo
 
 ## In what formats are data available?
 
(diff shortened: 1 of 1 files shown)


LucyGwilliamAdmin:
@brockfanning I've just approved this but now I'm thinking the paths will change when we upgrade to v1 to have `/en/` in them right?

brockfanning:
@LucyGwilliamAdmin Yes that's true.

One thing we discussed in the last call (I think before you joined) was the fact that these data endpoints are all highly dependant on how the data repo is configured. So we were thinking of add a feature to sdg-build that generates an HTML page listing all the endpoints that are available. This way, we can remove all specific paths from this FAQ documentation, and just say "consult your data repository for specific paths" or something like that.

So it may be most efficient for us to work towards that first. I have been looking into it.

Issue #356, opened 2019-09-13
min max value in maps


christoph4711:
In my map [here](https://christoph4711.github.io/open-sdg-site-starter/1-2-1/) in the map legend the max vaule is displayed with 268505 which is not displayed on the map, because this value is a aggeregation of a couple of cities, what I need to have in the table an the graph view, but can't (yet)  display on the map. In the following the classses displayed on the map are too large...

![grafik](https://user-images.githubusercontent.com/53389268/64840361-7b151e80-d5fb-11e9-9b8c-8998b5fd42d3.png)

It is possible to set a min and max vaule for the map? Or may be the min and max vaule can be caluclated for each year from the data that has a GeoCode, data without GeoCode should be ignored.

brockfanning:
@christoph4711 I'm not sure, but the problem may be related to #109. If so, a quick fix would be to remove the GeoCode from all rows that are maennlich or weiblich. In theory this might fix the problem.

Though of course a better fix would be if we could resolve #109, and an even *better* fix would be if we could come up with a way to work disaggregation into the maps.

brockfanning:
The map functionality had somewhat of a rewrite in version 1.0.0, so I'm curious to know if this is still an issue. Any updates would be appreciated.

LucyGwilliamAdmin:
I didn't create this issue but here are my comments from looking at this UK feature branch:
http://sdgdev-813006012.eu-west-1.elb.amazonaws.com/map-tooltip-fix/3-a-1/

Looks like min and max values on legend aren't min and max of what is currently shown on map e.g.
Table:
![image](https://user-images.githubusercontent.com/43404358/86583489-48b62280-bf7b-11ea-9db6-939ac5b17d76.png)

Map:
![image](https://user-images.githubusercontent.com/43404358/86583529-59669880-bf7b-11ea-8c3f-9d60b7900a9b.png)

Max value is fine (23.3578 (23.4 to 1d.p) is Scotland 2011 value) but 12.9203.. (12.9 to 1d.p) isn't a minimum - min value for 2011 is Northern Ireland (18.9) and min value overall is England 2018: 14.4

However when zoom in to lower level of map, 12.9203.. is min value for South East in 2018

WohnungNDS:
@brockfanning 
@LucyGwilliamAdmin 

[...]

PR #663, opened 2020-05-05
WIP: Map disaggregation select

brockfanning:
This is a work in progress - do not merge. To be discussed in an issue.


WohnungNDS:
@brockfanning it works as you can see [here](https://wohnungnds.github.io/IM_Site/11-1-1/)
There are some Problems with the min/max values. To my mind the min/max values of the durst disaagregation is taken to the legend. Selecting another disaggregation does not change the mkin/max values in the lengend. In the following the values are shown right in the map but not the colors. perhaps you try it with 

"Asylberechtigte und GFK-Flüchtlinge"
min: 155 "Lychow-Dannenberg" in year 2015
max: 47725 "Region Hannover" in year 2019

changing the disaggregation to e.g.
"Duldung"
min 150 "Holzminden" in Year 2016
max 2845 "Region Hannover" in year 2019
refreshes the map data but not the legend

In the following the map colors are like the first disaggregation

perhaps this has been  already discussed in #356

Another point is the sorting of the disaggregation entries as you can see [here](https://wohnungnds.github.io/IM_Site/1-2-3/)

image

Overall it works great testet in chrome, Ipad, Iphone and android

brockfanning:
I'll close this PR because I created a new one: #1498 

I created a separate branch so that I could leave this branch as-is, because I know that it is being used. The new PR linked above is based on the latest code, where is this one is based on an outdated version of Open SDG.

PR #670, opened 2020-05-10
Remove stray closing div

brockfanning:
From a quick glance this appears to be an extra closing div. Manual testing is needed.


brockfanning:
Disregard, I think this is fine as-is.

PR #671, opened 2020-05-10
Enforce black background on high-contrast goal images - to support tr…

brockfanning:
This is to support transparent PNGs. Without this, the goal tile shows the normal contrast background color on hover, if the PNG is transparent.

### DIFF
--- a/_sass/components/_high_contrast.scss
+++ b/_sass/components/_high_contrast.scss
@@ -190,3 +190,14 @@ body.contrast-high #disclaimer {
     color: white;
   }
 }
+
+body.contrast-high {
+  @for $i from 1 through length($goal-colors) {
+    .goal-tiles #goal-#{$i} {
+      background-color: black;
+      &:hover {
+        background-color: black;
+      }
+    }
+  }
+}
--- a/_sass/components/_high_contrast.scss
+++ b/_sass/components/_high_contrast.scss
@@ -192,12 +192,10 @@ body.contrast-high #disclaimer {
 }
 
 body.contrast-high {
-  @for $i from 1 through length($goal-colors) {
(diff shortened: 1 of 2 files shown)


_sass/components/_high_contrast.scss
```diff
@@ -190,3 +190,14 @@ body.contrast-high #disclaimer {
     color: white;
   }
 }
+
+body.contrast-high {
+  @for $i from 1 through length($goal-colors) {
```
SavvasStephanides, line 195: Any reason this should be a loop rather than the elements having a common CSS class name?

brockfanning: @SavvasStephanides I used a loop because it's meant to override the selectors here: https://github.com/open-sdg/open-sdg/blob/master/_sass/components/_goal_tiles.scss

Maybe it's better in this case to just use "important" instead of the loop?

SavvasStephanides: I would avoid adding unnecessary logic into my (S)CSS as much as possible, including `for` loops. The loop in the link you pasted above is necessary because each iteration gives a new colour, which is needed for styling the goal tiles. Your code in this change seems to be better implemented by adding a class which is common in all your goal tiles and implementing the styling change there. The reason I'm saying this is because this change seems to add the same styling update in all goal tiles (unless I'm missing something). Not sure if the `!important` keyword is necessary. Try it and see.

brockfanning: @SavvasStephanides I was assuming that `important` would be necessary to compete with the id selector that the existing loop is doing. And it does look like that's the case. I've pushed up an update using important instead of the loop, and also moving the 'goal-tiles' class to a child element so that it's more accurate.

PR #673, opened 2020-05-11
CircleCI updates in docs

brockfanning:
Some references to CircleCI need to be updated, since we've switched the starter repos to Github Actions.

### DIFF
--- a/docs/deployment.md
+++ b/docs/deployment.md
@@ -8,10 +8,10 @@ That said, the platform does require some server-side technologies to *build* th
 
 ## Automation services
 
-The easiest way to perform these builds is to use a free automation service like TravisCI or CircleCI. Alternatives such as self-hosted Jenkins are possible, but require more maintenance and likely would not be free.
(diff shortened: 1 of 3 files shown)


docs/deployment.md
```diff
@@ -8,10 +8,10 @@ That said, the platform does require some server-side technologies to *build* th
 
 ## Automation services
 
-The easiest way to perform these builds is to use a free automation service like TravisCI or CircleCI. Alternatives such as self-hosted Jenkins are possible, but require more maintenance and likely would not be free.
+The easiest way to perform these builds is to use a free automation service like Github Actions, TravisCI, or CircleCI. Alternatives such as self-hosted Jenkins are possible, but require more maintenance and likely would not be free.
 
 > Note that the "starter" repositories are pre-configured for to handle all of
```
LucyGwilliamAdmin, line 13: Think this should say
`> Note that the "starter" repositories are pre-configured to handle all of`
instead of
> Note that the "starter" repositories are pre-configured **for** to handle all of

PR #674, opened 2020-05-11
Mention Prose.io issues with write permissions

brockfanning:


LucyGwilliamAdmin:
@brockfanning definitely think we need something around this but I want to do some further investigation about what is possible with the different permission types e.g. Admin so we can provide a full view

brockfanning:
I'll go ahead and close this, as I think we're almost ready to move away from Prose.

PR #678, opened 2020-05-16
Series column

brockfanning:
This allows the special column "Series" to display radio buttons in the data sidebar. It can be thought of as a second unit of measurement. It should be usable the same as units in every way. Lots of testing is needed to make sure that all the various scenarios are working.

This depends on the latest code from sdg-build.


LucyGwilliamAdmin:
@brockfanning I think I've found an example of an indicator that has both units and series: [12.3.1 on UK site](https://sustainabledevelopment-uk.github.io/12-3-1/) has 4 unit types:
- Food (Kilogrammes per person)
- Food (Kilotonnes)
- Food and inedible waste (Kilogrammes per person)
- Food and inedible waste (Kilotonnes)

I thought they could be separated up into series and units:
Series:
- Food
- Food and inedible waste
Units:
- Kilogrammes per person
- Kilotonnes

I have tested this out on this [test site](http://sdgdev-813006012.eu-west-1.elb.amazonaws.com/series/12-3-1/) (the time series aren't joined up but this seems to an existing v1 problem as problem also occurs on [v1 test site](http://sdgdev-813006012.eu-west-1.elb.amazonaws.com/opensdg1-data-test/12-3-1/)

LucyGwilliamAdmin:
I also found [8.5.1 on live site](https://sustainabledevelopment-uk.github.io/8-5-1/) which has two units:
1. Annual survey of hours and earnings (ASHE)
2. Annual population survey (APS)

I thought these made more sense as series so here's how that looks on [test site](http://sdgdev-813006012.eu-west-1.elb.amazonaws.com/series/8-5-1/)

For some reason it's behaving differently.

brockfanning:
@LucyGwilliamAdmin That issue with the line not appearing is a really interesting issue. If it's not specific to this branch, then yeah I suspect it's unrelated, and we should open a separate ticket about it.

brockfanning:
@LucyGwilliamAdmin Following up on your question:

> Nazira has also confirmed that she is happy with the feature. She did ask about whether it would be possible to select more than one series at a time - so I guess it would act more like a disaggregation than a Units type but think we could create new issue for that?

Absolutely, and it's an important question to resolve now, before merging this. I believe that, if we want Series to behave like a disaggregation, then this PR can be closed. Hopefully it would work as-is, without any new Series-specific code. It certainly would nice to avoid introducing this complication, if the feature is already supported by our disaggregation code.

One note - with series as a disaggregation, it would not be possible to support series-specific chart titles, if that makes a difference.

LucyGwilliamAdmin:
@brockfanning agreed this would not be needed if Series was to be treated like a disaggregation but I also think there may be different needs from different countries. For example, in the UK we would want it to act like the Units but as Nazira explained, they would want it to act like a disaggregation.

I'm not saying this is what we should do, but just wondering if possible? Could it be configurable i.e. either in config file or individual metadata files, there could be a field e.g. series_type: toggle/dropdown

@AnnCorp @phillipgardiner what are your thoughts?

brockfanning:
Making it configurable is possible, yes, because the javascript can react to particular metadata fields.

Basically I think the options are:

1. Series will always act like a disaggregation - close this PR
2. Series will always act like a unit - test/approve/merge this PR
3. Series will be configurable to act like both - test/approve/merge this PR and then open a new ticket to make it configurable

LucyGwilliamAdmin:
I'm for option 3 but interested in @phillipgardiner from a UR point of view?

Also @ann what are your thoughts?

AnnCorp:
Yes, I agree option 3 seems best - fits with our general ethos of things being customisable and configurable.

brockfanning:
@LucyGwilliamAdmin All automated tests are going to fail until we update sdg-build (merging the disaggregation-report PR would fix it). But otherwise this is ready for manual testing. It's now using an optional site config called `series_toggle`. If set to "true" then the Series column should display as radio buttons. If "false" or omitted, the Series column should display as checkboxes.

LucyGwilliamAdmin:
Ok, have just had a look at this on Kyrgyzstan site:
This is how it looks with `series_toggle: true`:
image

Here's how it looks with `series_toggle` omitted:
image

For some reason Series is translated for toggle, but not for dropdown.

This is just from looking at functionality on Kyrgyzstan site as it is - I will take more in-depth look on UK feature branch

brockfanning:
Ah, good catch. I can imagine why that's happening. The label for the toggle version is hardcoded to display the `indicator.series` translation key (referencing [this](https://github.com/open-sdg/sdg-translations/blob/master/translations/en/indicator.yml#L31). But we don't provide any translations for disaggregations in SDG Translations. The assumption has been that these are mostly custom, and so each country would need to translate their own. To translate "Series" in the disaggregations, there would need to be a "data.Series" translation key.

One possible solution is to add this "data.Series" translation key to SDG Translations. Another is to rely on the country to do that translation.

That makes me wonder though if we shouldn't have used the SDMX code "SERIES" instead of "Series". Using "Series" is sort of arbitrary, and "SERIES" at least matches a standard (SDMX). Maybe if we add a "data" group to SDG Translations, the keys should all correspond to SDMX ids? Just musing, I'll think on this some more.

brockfanning:
@LucyGwilliamAdmin The tests are passing again for this one. The translation issue is a tricky one to wrap my brain around. Since this feature is most needed for a country using an SDMX input, I'll talk through the problem from that perspective:

1. The translations are being imported from the SDMX DSD
2. The translations of "Series" are available at the translation key "concept.SERIES"
3. The "dimension_map" on the SDMX input is converting "SERIES" to "Series"
4. So when it gets to Open SDG, it is just "Series".
5. Open SDG checks to see if "Series" is a translation key, and decides it is not (because it does not have a dot)
6. Open SDG checks to see if "data.Series" is a translation key, and decides it is not (because a translation does not exist for that key)
7. Open SDG just displays "Series", regardless of language

So I think the translation issue would resolve itself if it could remain "concept.SERIES" when it gets to Open SDG. But in this PR all the code is specifically looking for "Series" to apply this special behavior to. We'll need to figure out some way around this problem.

brockfanning:
I guess the easiest solution for all concerned would be for us to add a "data" translation group to SDG Translations (data.yml), and include a "Series" translation. This would ensure that a "data.Series" translation is always available.

brockfanning:
TODO: Add automated test for indicator with both Units and Series.

LucyGwilliamAdmin:
@brockfanning was just looking to get a screenshot for slides and found this indicator on Kyrgyzstan site - had series and units.

Live site is still using toggle: https://sustainabledevelopment-kyrgyzstan.github.io/en/8-10-2-2/

Staging site using dropdown: https://sdg-kyrgyzstan.github.io/open-sdg-site-starter/en/8-10-2-2/

Each of the series has a different units type but user is still able to select either

brockfanning:
Great, thanks @LucyGwilliamAdmin. Does this sound like what you're looking for?

Given I am looking at an indicator with both Series and Units
And Series is in the toggle form (radio buttons)
Then I should only see units of measurement that exist within the selected Series data
And if I switch to a different Series
And the selected unit does not exist in the new Series data
Then a different unit (the first one?) should be selected on my behalf

brockfanning:
@LucyGwilliamAdmin @AnnCorp Lucy points out an important issue, I think, that should probably be resolved when this feature is released. So I'm thinking this could wait until 1.2. My goals for this would be:
1. Only subcategories available in a series will display
     * If the series is changed, this updates accordingly
2. Only units available in a series will display
     * If the series is changed, this updates accordingly
3. Only subcategories available in a unit+series will display
     * If the series OR unit is changed, this updates accordingly

The third item above would be a nice improvement to current behavior. (Currently, all subcategories display, regardless of the selected unit.)

brockfanning:
Instead of what I detailed above, we're in agreement to try a message approach, like detailed in #699.

brockfanning:
This still needs work - I've noticed a case where it appears to display data from a series that is not selected.

LucyGwilliamAdmin:
@brockfanning where is this?

brockfanning:
@LucyGwilliamAdmin I added a test to demonstrate the issue. You can see if failing [here](https://github.com/open-sdg/open-sdg/pull/678/checks?check_run_id=1027554170#step:5:267). Also if you can do the `make serve` thing then you can run through [these steps](https://github.com/brockfanning/open-sdg/blob/series-column/tests/features/Series.feature#L23-L32) manually.

brockfanning:
@LucyGwilliamAdmin I just pushed a commit that appears to have fixed that issue - so I think this is ready for review now.

LucyGwilliamAdmin (approved):
Have tested again with Kyrgyzstan site and all seems good

Issue #688, opened 2020-05-18
Default selection of disaggregations


LucyGwilliamAdmin:
Most default selecting seems fine except for when a new units type is selected – see [3.3.3 on UK test site](http://sdgdev-813006012.eu-west-1.elb.amazonaws.com/opensdg1-data-test/3-3-3/) – nothing shows on graph when Percentage (%) of total cases is selected. Also generally for [4.1.1](http://sdgdev-813006012.eu-west-1.elb.amazonaws.com/opensdg1-data-test/4-1-1/), may also be fixed if first thing is fixed

LucyGwilliamAdmin:
@brockfanning Again, will this be fixed by fixing #781 ?

LucyGwilliamAdmin:
I think this is fine now we have no data message:
https://sdgdata.gov.uk/3-3-3/

brockfanning:
Agreed.

Issue #577, opened 2020-03-25
Disaggregations not resetting when unit type is changed


LucyGwilliamAdmin:
Currently disaggregations aren't resetting when unit type is changed. For example, see 3.3.3 on UK site.

By default looks like this:
![image](https://user-images.githubusercontent.com/43404358/77561174-46412880-6eb6-11ea-9aa7-bc7a721bcad0.png)

Then when other units type is selected (Percentage of total cases):
![image](https://user-images.githubusercontent.com/43404358/77561273-6244ca00-6eb6-11ea-8d48-e90ba7b92483.png)

Then back to default units type:
![image](https://user-images.githubusercontent.com/43404358/77561472-a59f3880-6eb6-11ea-8549-aa90ef4dab6c.png)

Note: I haven't selected/unselected any disaggregation options

LucyGwilliamAdmin:
I think I would expect the first and third graph to look the same, especially the fact that the UK would still show as headline (grey bar).

Do we want the automated disaggregation selection process to be restarted whenever a new units type is selected?

LucyGwilliamAdmin:
Another example on Kyrgyzstan site: [7.3.1.2](https://sdg-kyrgyzstan.github.io/open-sdg-site-starter/en/7-3-1-2/)

When first land on page there are no disaggregation options available with default unit type, but then when you click on other unit type and back on default, disaggregation options from other unit type still display, making it look to the user like they have options.

brockfanning:
This seems like a tricky question - do you know if there is any upcoming user testing that might inform how it should work?

The first thought that occurs to me is that, under normal circumstances, changing the Unit doesn't reset anything. But if the other Unit doesn't have the same disaggregations available, then the Unit radio button includes some sort of message like "(selected disaggregations will be reset)" or something to that effect. Or maybe a popup. I'm not sure how technically feasible that is, just brainstorming.

LucyGwilliamAdmin:
@brockfanning don't think there's anything coming up which could help with this..
@AnnCorp what are your thoughts?
Phill (not sure why I can't tag you) - have you come across anything like this before?

phillipgardiner:
[...]

PR #699, opened 2020-05-18
Use the selected unit when calculating minimum field values

brockfanning:
This partially address #688. Specifically this makes sure that starting disaggregations will take into account a particular unit, when necessary. What was happening was that a starting unit was getting selected, and then the starting fields were getting selected, but they didn't always match.

### DIFF
--- a/_includes/assets/js/indicatorModel.js
+++ b/_includes/assets/js/indicatorModel.js
@@ -141,7 +141,7 @@ var indicatorModel = function (options) {
       }
       else {
         if (headline.length === 0) {
-          startingFields = helpers.selectMinimumStartingFields(this.data, this.selectableFields);
+          startingFields = helpers.selectMinimumStartingFields(this.data, this.selectableFields, this.selectedUnit);
         }
       }
       if (startingFields.length > 0) {
@@ -172,7 +172,7 @@ var indicatorModel = function (options) {
       });
     }
 
-    if (selectionUpdateNeeded) {
+    if (selectionUpdateNeeded || options.unitsChangeSeries) {
       this.updateFieldStates(this.selectedFields);
     }
 
--- a/_includes/assets/js/model/fieldHelpers.js
+++ b/_includes/assets/js/model/fieldHelpers.js
@@ -238,10 +238,17 @@ function selectFieldsFromStartValues(startValues, selectableFieldNames) {
 /**
  * @param {Array} rows
  * @param {Array} selectableFieldNames Field names
+ * @param {string} selectedUnit
  * @return {Array} Field items
  */
-function selectMinimumStartingFields(rows, selectableFieldNames) {
-  var filteredData = rows.filter(function(row) {
+function selectMinimumStartingFields(rows, selectableFieldNames, selectedUnit) {
+  var filteredData = rows;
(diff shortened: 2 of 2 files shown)


LucyGwilliamAdmin:
Ok, great - this fixes this scenario: [4.1.1](http://sdgdev-813006012.eu-west-1.elb.amazonaws.com/selected-unit/4-1-1/)

LucyGwilliamAdmin:
There still seems to be an issue with [3.3.3](http://sdgdev-813006012.eu-west-1.elb.amazonaws.com/selected-unit/3-3-3/) (no disaggregation automatically selected when unit changed and no headline data available)

brockfanning:
@LucyGwilliamAdmin The issue on 3.3.3 reminds me of the discussions in #577. Right now all the logic for choosing "minimum" starting disaggregation only happens when the page first loads, but not when units are changed. I could try to make it select minimum disaggregation even when units are changed. But then that disaggregation will remain selected if they switch back to the original unit. So the user experience would be something like:
1. Load 3.3.3 and see headline data for unit "Number of cases"
2. Switch to unit "Percentage of total cases" and see disaggregation "Country of birth" set to "Africa".
3. Switch back to unit "Number of cases" and see "Africa" still selected

If you'd be happy with this, let me know and I can keep going. Otherwise maybe more discussion is needed on what we want the UX to be in these cases. I guess one alternative might be some sort of message displayed to the user like "No general data is available for this unit. Try selecting sub-categories on the left."?

LucyGwilliamAdmin:
@brockfanning ahh ok - think we said for 577, that it's something to be user researched? if so happy to approve this one and close the new issue

brockfanning:
@LucyGwilliamAdmin I do think there are some user-experience questions to answer. There's a clear problem, in that the user is shown a blank chart with no indication of what to do - but to me it's not obvious what the best behaviour is there.

LucyGwilliamAdmin:
@AnnCorp @phillipgardiner @SavvasStephanides do you have any thoughts on this?

SavvasStephanides:
Are we talking about displaying a blank graph for certain disaggregations? I'm not sure what's the best way to tackle this, but this is what I've seen other websites do:

Group 1

I don't see anything in the UK Gov's design guidelines regarding graphs

brockfanning:
Between the options of a) automatically selecting disaggregations on behalf of the user, and b) displaying a message to the user that no data is available, I like b) better. I worry about automatically selecting disaggregations for the user, and the confusion that might cause.

However either way I think that could be done in a separate PR.

phillipgardiner:
Hi All, Sorry i'm struggling a bit to get my head around this issue. However from what I can work out something along the lines of option B would be a preference. Automatically selecting dissagregations for users can cause alot of confusion as actions are happening without them realising etc.

AnnCorp:
Agree with Brock and Phill - option B sounds best.

LucyGwilliamAdmin:
Yeah, I like b - think what @SavvasStephanides mocked up looks good

brockfanning:
@LucyGwilliamAdmin The code for that would be separate from what's already in this PR, so if it's OK with you I think we could continue with that in a new one. Is this one OK to merge, otherwise?

PR #702, opened 2020-05-19
limit toc depth to 2

LucyGwilliamAdmin:


brockfanning:
@LucyGwilliamAdmin I'm not sure offhand. I remember investigating a while back, and I believe it turned out that there is not too much we can do to customise. A quick fix might be to avoid using Markdown headings (eg, `## My heading`) for things you don't want to appear in the table of contents. I wonder if using HTML headings might be a workaround? (eg, `<h2>My heading</h2>`)

PR #703, opened 2020-05-20
Try using html headings instead

LucyGwilliamAdmin:


LucyGwilliamAdmin:
@AnnCorp trying this to get rid of sub headings in side bar - if this doesn't work I'll make them shorter as you suggested.

LucyGwilliamAdmin:
@AnnCorp great this worked - @brockfanning's idea :)

PR #705, opened 2020-05-20
Sitewide link styling

SavvasStephanides:
This change 

* sets all the links, apart from header and footer links, to be blue and underlined. 

* Adds styling for visited pages

* Some refactoring to make this update slightly easier.

### Default contrast

image

### High contrast

image

Note: the orange colour on black background [passes all the contrast tests](https://webaim.org/resources/contrastchecker/?fcolor=000000&bcolor=F47738)


brockfanning:
@SavvasStephanides Do you think it could be called just _links.scss, and imported from open-sdg.scss, like the others?

SavvasStephanides:
@brockfanning I was debating that with myself. I wanted to signify that it's styling changes to links as a _subset_ of typography

brockfanning:
@SavvasStephanides Before the visited colors were in I tested locally and it looked great! A couple comments before I try it out again to see the visited colors:
1. Not to obsess over the naming of one file, but... I'm not understanding the need for this to be a subset of typography. To me it seems like this is just about links, which I'd argue is distinct enough as its own concept/element. Naming it `_typography_links.scss` makes it seem like there could be other `_*_links.scss` files, but it's not clear to me what those would be. If you could give examples of other `_*_links.scss` files that might be needed in the future, that might help me understand.
2. I've tried to make all the variables in the `_colors.scss` file end with `-color` (or `-colors` if it's an array of them). Do you have any objections to sticking to that as a standard?

SavvasStephanides:
> Not to obsess over the naming of one file, but... I'm not understanding the need for this to be a subset of typography. To me it seems like this is just about links, which I'd argue is distinct enough as its own concept/element. Naming it _typography_links.scss makes it seem like there could be other _*_links.scss files, but it's not clear to me what those would be. If you could give examples of other _*_links.scss files that might be needed in the future, that might help me understand.

Thanks @brockfanning. Basically the code behind this file was initially developed in `_typography.scss`, since they were mostly changes in the text of the links. I later migrated them into its own file so I named it `_typography_links.scss`. I appreciate I may have had the wrong idea by name it like this and naming it just `_links.scss` might be more "inclusive". I'll change the filename to that, but do keep in mind that it's going to be quite the migration to include all the link properties in that file (we can do it in a later refactor).

> I've tried to make all the variables in the _colors.scss file end with -color (or -colors if it's an array of them). Do you have any objections to sticking to that as a standard?

I don't disagree with what you did and I have no problem sticking to it as a standard but my idea was to have a structure such as _element-modifier-extra property_, for example `$link-color-default`. So our file would look a bit like this:

```scss
$link-color-default
$link-color-highcontrast
$link-color-visited
$link-backgroundColor-default
$link-backgroundColor-highcontrast
```

which I think would give our file a little more structure. What do you think?

brockfanning:
@SavvasStephanides I started writing a reply and it got so long, I figured it should be a separate issue. :) See #712

SavvasStephanides:
@brockfanning Renamed the `scss` file. Should be ready for re-review

brockfanning:
@SavvasStephanides This looks good, but since we're talking about the sass variable naming convention revamp, it will probably be least painful to rebase this after something like #715 is merged. There's probably going to be merge conflicts regardless, but it's probably best to resolve them now before merging into 1.1.0-dev. Make sense?

SavvasStephanides:
Yeah that makes perfect sense. That was my concern since we decided to have two different branches for `1.0` and `1.1`.

brockfanning:
@SavvasStephanides I just updated the 1.1.0-dev branch with our Sass variable naming convention. As expected there are conflicts.

SavvasStephanides:
@brockfanning This should be ready for re-review

_sass/variables/_colors.scss
```diff
@@ -21,9 +21,17 @@ $color-success: #5cb85c !default;
 $goal-colors: #E52B3D #DDA63A #4C9F38 #C42130 #FF3E24 #28BCE1 #FCC30B #A21942 #FD6925 #DD1367 #FD9C27 #BF8B2E #417D46 #0A96D8 #57BE2F #00689C #1F4A6A !default;
 
 $text-color: #333 !default;
-
-$link-color: #337ab7 !default;
+$text-dark-color: black !default;
```
brockfanning, line 24: `text-dark-color` has been replaced by `color-dark`.

_sass/variables/_colors.scss
```diff
@@ -21,9 +21,17 @@ $color-success: #5cb85c !default;
...
+$text-on-dark-background-color: white !default;
```
brockfanning, line 25: `text-on-dark-background-color` has been replaced by `color-light`.

_sass/variables/_colors.scss
```diff
@@ -21,9 +21,17 @@ $color-success: #5cb85c !default;
...
+$background-color: white !default;
```
brockfanning, line 26: `background-color` has been replaced by `backgroundColor`.

_sass/variables/_colors.scss
```diff
@@ -21,9 +21,17 @@ $color-success: #5cb85c !default;
...
+$link-color: #1d70b8 !default;
 $link-color-dark: #0275d8 !default;
+$link-color-hover: #003078 !default;
+$link-color-visited: #4c2c92 !default;
+$link-color-visited-highcontrast: #F47738 !default;
```
brockfanning, line 31: Our convention is camelcasing the modifiers, eg `link-color-visited-highContrast`.

_sass/variables/_colors.scss
```diff
@@ -21,9 +21,17 @@ $color-success: #5cb85c !default;
...
+$dark-highlight: #0275d8 !default;
```
brockfanning, line 32: `dark-highlight` has been replaced by `link-color-dark`.

_sass/variables/_colors.scss
```diff
@@ -21,9 +21,17 @@ $color-success: #5cb85c !default;
...
+$footer-color: #000000 !default;
```
brockfanning, line 33: `footer-color` has been split up and replaced by `footer-text-color` and `footer-link-color`.

_sass/variables/_colors.scss
```diff
@@ -21,9 +21,17 @@ $color-success: #5cb85c !default;
...
+$footer-color-highcontrast: #FFFFFF !default;
```
brockfanning, line 34: This has been split up and replaced by `footer-text-color-highContrast` and `footer-link-color-highContrast`.

_sass/variables/_colors.scss
```diff
@@ -72,6 +80,7 @@ $menu-link-color-mobile: #E5E6E7 !default;
 $menu-link-color-mobile-active: white !default;
 $menu-link-backgroundColor-mobile: #414042 !default;
 $menu-link-backgroundColor-mobile-active: #58595B !default;
+$menu-hover-border-bottom: #aaaaaa !default;
```
brockfanning, line 83: Per our convention this should be something like `menu-link-borderBottomColor-hover`.

SavvasStephanides:
Whoops! I focused more on resolving the merge conflicts. Thanks Brock!

_sass/base/_links.scss
```diff
@@ -0,0 +1,50 @@
...
+.navbar ul.navbar-nav li:not(.contrast) a:hover{
+    border-bottom: 2px solid $menu-link-borderBottomColor-hover;
+}
+
+#search-results .search-result a {
+    text-decoration: none;
+}
+
+footer a{
+    color: $footer-link-color-highContrast;
```
brockfanning, line 41: This shouldn't be the highContrast version, since this is not inside body.contrast-high.

_sass/base/_links.scss
```diff
@@ -0,0 +1,50 @@
...
+}
+
+footer a:hover{
+    color: lighten($footer-link-color-highContrast, 20%);
```
brockfanning, line 45: Same here, this shouldn't be highContrast because it's not inside body.contrast-high.

_sass/base/_links.scss
```diff
@@ -0,0 +1,50 @@
...
+}
+
+body.contrast-high footer a:visited{
+  color: $footer-link-color-highContrast;
```
brockfanning, line 49: From this I take it that we don't need to distinguish visited links in the menus?

SavvasStephanides: Not in the footer.

SavvasStephanides:
Good catches Brock!

SavvasStephanides:
@brockfanning You can go ahead with another review Brock

brockfanning:
@SavvasStephanides Looks great to me. I went ahead and approved - though I think we still need to make a tweak to the search results so that they're underlined. If you prefer to do that in a separate ticket that's fine, or you could push it up here before merging.

SavvasStephanides:
Thanks Brock. I'll push these changes now and work on the search changes later

PR #706, opened 2020-05-20
Allow goal banner color and footer background to be overriden

brockfanning:
This is to make it easier for countries to keep their existing (pre-1.0.0) goal banner and footer styles. I also moved some goal-banner stuff into a dedicated file, instead of having in the goal-tiles file.


brockfanning:
@SavvasStephanides Oops, accessibility issues. Let me resolve those first before you review.

brockfanning:
@SavvasStephanides Ok, ready again for review.

_sass/variables/_colors.scss
```diff
@@ -8,6 +9,7 @@ $background-color: white !default;
 $link-color: #337ab7 !default;
 $dark-highlight: #0275d8 !default;
 $footer-color: #000000 !default;
+$footer-background-color: white !default;
```
SavvasStephanides, line 12: Worth including the high contrast version as well?

brockfanning:
We're talking about a Sass variable refactor, so I'll close this for now.

PR #714, opened 2020-05-21
Download the geojson file instead of opening it

### LINKED ISSUE #684: Geojsons open in new tab instead of downloading
Is there a way to force the file to download rather than opening in a new tab? This would be more consistent with the rest of the download buttons (see [3.4.2](http://sdgdev-813006012.eu-west-1.elb.amazonaws.com/opensdg1-data-test/3-4-2/) on test site)

brockfanning:
Fixes #684


SavvasStephanides:
Works great so happy to approve. Just a question though: what's the reason for building the button using Javascript as opposed to simple HTML?

brockfanning:
@SavvasStephanides To be honest I can't remember - it may have just been easier at the time, because the download button was previously done as a Leaflet plugin, which has to be Javascript.

Issue #712, opened 2020-05-21
Sass variable conventions


brockfanning:
Followup from #705 

@SavvasStephanides 
Cool, I'm definitely open to changing things around a particular convention - especially now, before 1.0.0 is released! Let me talk through some stuff though, to get it straight in my head.

Our main goal here is to allow countries to override colors for all the bits and pieces of the platform, independently. So, for example, we want to allow a country to override the footer background color _only_, while leaving everything else the same.

So our convention may be a bit different from a typical web site or design system. For example a typical web site or design system may have variables like `color-primary`, `color-primary-dark`, etc. But we need more granular "component-level" variables.

At the same time, the design-system-ish variables might be useful too. For example, what if a country wants to change the background colors across the site. Having to override the background color of every component seems cumbersome. It would be much better if they could change one variable.

So I'm imagining two separate files of color variables. One file, say `_colors-global.scss`, is the design-system-y approach, which has component-agnostic variables like `color-primary`, etc. (or whatever we decide on as a standard). The other file, say `_colors-by-component.scss` has the granular component-level variables, like `footer-background-color` (or again, whatever we decide on as a standard). **The second file would not contain any actual colors** (hex values, etc) and instead would just have references to the variables in `_colors-global.scss`.

Or maybe instead of two separate files they could just be different sections of `_colors.scss`?

That way, if a country wants to override background colors across the site, they can override one of the "colors-global" variables. And if they instead just want to override the background color of a particular component, they can override a "colors-by-component" variable.

Anyway just brainstorming here. Does that make sense to you?

Regarding the actual standard we use, I'm flexible. I do think they should all include some indication of whether they are a foreground color ("color") or background color ("background-color"). And if they are specific to "hover", "focus", etc, they'll need to include that. And also they'll need something for "high-contrast". And if they're component-specific, that needs to be there too, probably at the beginning.

I like your `element-modifier-extra property` idea, but maybe with the component when needed. So normally `element-modifier-extra property`, but if it's component-specific, then `component-element-modifier-extra property`?

SavvasStephanides:
Thanks for the reply @brockfanning.

> Our main goal here is to allow countries to override colors for all the bits and pieces of the platform, independently. So, for example, we want to allow a country to override the footer background color only, while leaving everything else the same.

[...]

PR #715, opened 2020-05-22
Update Sass variables to fit a naming convention

brockfanning:
This is an attempt to fit a naming convention discussed in #712.

### DIFF
--- a/_sass/variables/_colors.scss
+++ b/_sass/variables/_colors.scss
@@ -54,10 +54,17 @@ $loader-color: #FF6700 !default;
 // Component-level color variables.
 // [component] - [element] - [color|backgroundColor|etc] - [...modifiers]
 $goal-banner-backgroundColors: $backgroundColor $backgroundColor $backgroundColor $backgroundColor $backgroundColor $backgroundColor $backgroundColor $backgroundColor $backgroundColor $backgroundColor $backgroundColor $backgroundColor $backgroundColor $backgroundColor $backgroundColor $backgroundColor $backgroundColor !default;
-$indicator-banner-backgroundColors: $goal-banner-backgroundColors !default;
+$goal-banner-backgroundColor-highContrast: $color-dark-highContrast !default;
+$goal-banner-color: $text-color-dark !default;
+$goal-banner-color-highContrast: $color-light-highContrast !default;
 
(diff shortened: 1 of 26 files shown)


_sass/variables/_colors.scss
```diff
@@ -1,82 +1,111 @@
...
 $goal-colors: #E52B3D #DDA63A #4C9F38 #C42130 #FF3E24 #28BCE1 #FCC30B #A21942 #FD6925 #DD1367 #FD9C27 #BF8B2E #417D46 #0A96D8 #57BE2F #00689C #1F4A6A !default;
-$goal-banner-background-colors: white white white white white white white white white white white white white white white white white !default;
 
 $text-color: #333 !default;
-$text-dark-color: black !default;
-$text-on-dark-background-color: white !default;
-$background-color: white !default;
+$text-color-dark: black !default;
+$text-color-light: white !default;
+
```
SavvasStephanides, line 23: Would these belong to the "top-level" variables? Also I think `color` implies that it's for text coloring.

brockfanning: Should we make them `color-dark` and `color-light` instead?

SavvasStephanides: Trying to figure out under what circumstances this colours are used. It seems to be a widely used colour, but not as "top-level", but rather when there's a box that has text in it, like the "ALPHA" box at the very top or the indicator statuses. Is this correct?

brockfanning: Yes `text-color-light` (or `color-light` if we change it) would be for any situation where the background is dark.

SavvasStephanides: That makes sense. I think it would be helpful if you specify that this is for those specific elements. Not sure what they're called. Phases? What do you think?

brockfanning: Sounds good, I just made that change, and the color-dark/color-light change.

SavvasStephanides: Looks good. Are the `color-dark` and `color-light` variables more general or are they specific to the disclaimers/phases as well? I'm under the impression that they are.

brockfanning: @SavvasStephanides Right, they are more general.

PR #718, opened 2020-05-26
Issue templates

LucyGwilliamAdmin:
@AnnCorp @brockfanning @SavvasStephanides @phillipgardiner 
I've made a start on the issue templates - any feedback is appreciated

The issues may be a little difficult to visualise here as they are in Markdown format but I've created them in this [test repo](https://github.com/LucyGwilliamAdmin/quicker-start/issues/new/choose) so you can see what they look like in action.


phillipgardiner:
Hello @LucyGwilliamAdmin @AnnCorp I think these are looking good. My main concern is how we handle feature requests. Users may request features because it is something they want or their country wants. It may not be suitable for everyone? Is it worth adding something to the feature one around whats the user need? / do they have any evidence of why this feature is required?. I guess we just want to be able assess those that are personal opinion v's observed behaviour issues.

AnnCorp:
@LucyGwilliamAdmin re feature request template think change 'Describe the solution you'd like' to say 'need' not 'like' and followi with another line to ask Why need it (the 'so that' from user stories). Think coudl then caveat in info @brockfanning is writing re how feature requests would be considered and prioritised (some may make it into Open SDg but some could just end up as country customisations).

.github/ISSUE_TEMPLATE/general.md
```diff
@@ -0,0 +1,9 @@
+---
+name: General
+about: Leave a general comment
```
brockfanning, line 3: I would maybe change this to "Leave a general comment or question" so that it can also cover support questions.

.github/ISSUE_TEMPLATE/data_bug_report.md
```diff
@@ -0,0 +1,31 @@
...
+name: Report a bug - data
+about: Report a bug with your data build
+title: ''
+labels: 'bug'
+assignees: ''
+
+---
+**Please complete the following information**
+Data repository URL: [e.g. github.com/ONSdigital/sdg-indicators]
+Data or metadata issue:
```
brockfanning, line 11: Is "Data or metadata issue" necessary? How is it different from what they would write under "Describe the bug"?

.github/ISSUE_TEMPLATE/data_bug_report.md
```diff
@@ -0,0 +1,31 @@
...
+Input format: [e.g. CSV, YAML (Markdown), SDMX]
+
+**Describe the bug**
+A clear and concise description of what the bug is.
+
+**To Reproduce**
```
brockfanning, line 17: I think maybe we could remove "To Reproduce", "Expected behavior", and "Screenshots". These seem more relevant for site bugs. What do you think?

.github/ISSUE_TEMPLATE/data_bug_report.md
brockfanning: Oh! I see, this is meant to ask whether the problem is related to data or metadata. Maybe this could be clarified, or possibly removed if we think it might be covered in the bug description.

.github/ISSUE_TEMPLATE/data_bug_report.md
LucyGwilliamAdmin: Think that makes sense!

LucyGwilliamAdmin:
@AnnCorp @brockfanning @phillipgardiner I have made the changes you have suggested

brockfanning:
@LucyGwilliamAdmin A Git-related request: Can you edit the PR and change "base: master" to "1.0.0-dev"?

LucyGwilliamAdmin:
> @LucyGwilliamAdmin I just noticed the "labels" options - that's cool! We do have a label called "enhancement" which could be added to the feature request template.

Good idea - just added it :)

brockfanning:
Oops, my Git request messed things up. Let me see if I can resolve, one sec.

PR #723, opened 2020-06-01
Use some status-related Sass variables

brockfanning:
We have a few Sass variables for the status types, which we're not using:
* $status-backgroundColor-notstarted
* $status-color-notstarted
* $status-backgroundColor-notstarted-highContrast
* $status-backgroundColor-notapplicable-highContrast

This PR adds usage of those variables. It also tweaks our Sass to always use "background" instead of "background-color", so that users can override with more complex backgrounds like gradients.

### DIFF
--- a/_sass/layouts/_reporting_status.scss
+++ b/_sass/layouts/_reporting_status.scss
@@ -1,7 +1,9 @@
 .notapplicable {
   background: $status-backgroundColor-notapplicable;
 }
-
+.notstarted {
+  background: $status-backgroundColor-notstarted;
+}
 .inprogress {
   background-color: $status-backgroundColor-inprogress;
 }
--- a/_sass/layouts/_reporting_status.scss
+++ b/_sass/layouts/_reporting_status.scss
@@ -1,13 +1,17 @@
 .notapplicable {
+  background-color: $status-backgroundColor-notapplicable;
   background: $status-backgroundColor-notapplicable;
 }
 .notstarted {
+  background-color: $status-backgroundColor-notstarted;
   background: $status-backgroundColor-notstarted;
 }
 .inprogress {
+  background-color: $status-backgroundColor-inprogress;
   background: $status-backgroundColor-inprogress;
 }
 .complete {
+  background-color: $status-backgroundColor-complete;
   background: $status-backgroundColor-complete;
 }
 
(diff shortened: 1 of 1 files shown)


SavvasStephanides:
> We have a few Sass variables for the status types, which we're not using:
> 
> $status-backgroundColor-notstarted
> $status-color-notstarted
> $status-backgroundColor-notstarted-highContrast
> $status-backgroundColor-notapplicable-highContrast

Good idea to include these. However, not sure the position of "notstarted" or "notapplicable" is correct here. Maybe it would be better as `$status-notstarted-backgroundColor` instead (since "notstarted" is directly related to status). What do you think?

> It also tweaks our Sass to always use "background" instead of "background-color", so that users can override with more complex backgrounds like gradients.

You don't need to worry about that. If both `background` and `background-color` are present in the same element, `background` will _always_ be preferred, with the only exception if `background` defines a solid color. `background-color` acts as a fallback for browser incompatibility or if a background image is not found, which is always nice to have. [See here for proof](https://codepen.io/SavvStudio/pen/vYLBZpB?editors=1100)

brockfanning:
@SavvasStephanides I guess it's a matter of interpretation whether notstarted/complete/etc represents a distinct component, or a "modifier" of the "status" component. When we did the naming convention refactor we treated them like modifiers, so I'm following that precedent here.

Understood about background-color vs background: My point was to make it easier for countries to do stuff with just Sass variables. If we use "background" in our Sass, rather than "background-color", then the countries can do gradients just by overriding the correct variable. But if we use "background-color" in our Sass, then countries have to override that Sass with their own Sass, if they want to do a gradient. I doubt many countries need this but I figured it didn't hurt to make the change. If you think it's objectionable then I can remove it.

SavvasStephanides:
> I guess it's a matter of interpretation whether notstarted/complete/etc represents a distinct component, or a "modifier" of the "status" component. When we did the naming convention refactor we treated them like modifiers, so I'm following that precedent here.

You're right. It just seemed a bit out of place when setting more than one attribute:

```scss
$status-backgroundColor-notstarted
$status-color-notstarted
```

Maybe they could be treated as elements instead?

```scss
$status_notstarted-backgroundColor
$status_notstarted-color
$status_notstarted-color-highContrast
```

> Understood about background-color vs background: My point was to make it easier for countries to do stuff with just Sass variables. If we use "background" in our Sass, rather than "background-color", then the countries can do gradients just by overriding the correct variable. But if we use "background-color" in our Sass, then countries have to override that Sass with their own Sass, if they want to do a gradient. I doubt many countries need this but I figured it didn't hurt to make the change. If you think it's objectionable then I can remove it.

I see your point. Putting it into context makes it a lot clearer, thanks. I'm fine with keeping these changes, but maybe change the variable name from `backgroundColor` to `background` to make it clear that you can set more than just a solid color (you can also include background images or gradients). Still though, it would be nice to have a fallback to a `background-color` (maybe an extra variable?).

brockfanning:
The only reason I hesitate to refactor our variable names again is that I know users that are already working against 1.0.0-beta4 to prepare for the launch, so it would be a bit of an inconvenience for them if we change things again. Do you think it's important enough for a re-refactor? To me it's "6 of one, half a dozen of the other", but I'm open to any arguments.

About the background-color fallbacks, that sounds good and I just pushed up background-color fallbacks before all the background rules. Does that work?

SavvasStephanides:
> The only reason I hesitate to refactor our variable names again is that I know users that are already working against 1.0.0-beta4 to prepare for the launch, so it would be a bit of an inconvenience for them if we change things again. Do you think it's important enough for a re-refactor? To me it's "6 of one, half a dozen of the other", but I'm open to any arguments.

That's a fair point. No, it's not at all important. It was just a consideration. Let's leave it as is.

> About the background-color fallbacks, that sounds good and I just pushed up background-color fallbacks before all the background rules. Does that work?

That'd work. It would still work if you had `background-color` after the background rules, but this works just as well. I'll have a quick check and approve

PR #726, opened 2020-06-02
crop extra white space out of logo

LucyGwilliamAdmin:


LucyGwilliamAdmin:
@AnnCorp if we merge this in first to see how it looks, then I'll get an idea of what it needs to be resized to

PR #729, opened 2020-06-03
Change size of logo

LucyGwilliamAdmin:


LucyGwilliamAdmin:
@AnnCorp it's smaller but still seems like there's a big gap - I'll remove some more white space

PR #730, opened 2020-06-03
Fix Open SDG logo

LucyGwilliamAdmin:


LucyGwilliamAdmin:
@AnnCorp just realised even though I edited the image, I must have dragged wrong file across when reuploading, so that's why it didn't look like there was any change in amount of white space.

This should fix it though

PR #750, opened 2020-06-18
Use increased font size

brockfanning:


brockfanning:
@SavvasStephanides Ok, made that change. I think we should follow up this PR with a Sass-wide update replacing all font-size rules with variables. Do you agree?

brockfanning:
@LucyGwilliamAdmin This could definitely use a thorough visual test, if you have time to do a feature branch.

SavvasStephanides:
Looks fine from my part.

> Ok, made that change. I think we should follow up this PR with a Sass-wide update replacing all font-size rules with variables. Do you agree?

I'd leave it to what's relevant for this ticket. This ticket is for setting the default font, plus it may be too much for a hotfix.

brockfanning:
@SavvasStephanides Right, sorry, I meant in a follow-up PR.

SavvasStephanides:
@brockfanning In that case, yes I agree.

PR #763, opened 2020-07-04
Update tooltips of active selections when the year changes

### LINKED ISSUE #385: Values on the labels in maps do not get updated
When changing the year, the map-color and the bar chart are getting updated, but the value on the label is not.

grafik

grafik

brockfanning:
Fixes #385

### DIFF
--- a/_includes/assets/js/plugins/jquery.sdgMap.js
+++ b/_includes/assets/js/plugins/jquery.sdgMap.js
@@ -86,6 +86,34 @@
       this.map.fitBounds(layer.getBounds());
     },
 
+    // Build content for a tooltip.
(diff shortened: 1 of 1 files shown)


MoBosse:
@brockfanning If I am not mistaken, this causes a loading error in IE again.
Is it possible that [this ](https://github.com/open-sdg/open-sdg/blob/master/_includes/assets/js/plugins/jquery.sdgMap.js#L90)line should actually say `getTooltipContent: function(feature) {`? Beacause the console in IE gives me a message saying that a ":" is expected there

brockfanning:
Great catch, thank you @MoBosse.

PR #776, opened 2020-07-14
Vertical goal-by-target layout

### LINKED ISSUE #633: Add single column version of goal page with targets
see Phill's mockup

brockfanning:
Fixes #633 

The new layout is called "goal-by-target-vertical". An easy way to switch all goals to this layout would be this config:

```
create_goals:
  layout: goal-by-target-vertical
```


brockfanning:
@LucyGwilliamAdmin Would you be able to make a feature branch for this one?

brockfanning:
Thanks @jwestw!

@phillipgardiner Any more UX/design feedback before I merge this?

phillipgardiner:
@brockfanning hey. Just had a look through this design on the feature branch - useful as it brings it to life. I do have a couple of comments but as always open to discussion. These comments are mainly when you look to reduce the screen size (e.g ipad, mobile). As you do so everything becomes a bit squashed on the screen, 

1) The 'reported online' and 'exploring data sources' boxes start to jump about and are no longer aligned with the indicator name
2) The 'reported online' and 'exploring data sources' boxes become very close the the horizontal lines etc.

I'm not 100% sure how to get around this, it looks great full screen but becomes quite hard to scan read once the screen size is reduced. I wonder if it is possible to do something like...when the screen size is reduced the 'reported online' and 'exploring data sources' boxes move from their original position to be aligned with the indicator number instead above the indicator name? I have attach a screenshot to try and show what I mean.

I'm totally open to other options!
Screenshot (33)

brockfanning:
@phillipgardiner Moving the status above the name would be tricky for technical reasons (because the order of the markup is number, then name, then status). As an alternative I've tried to get the status to align left on smaller screens. Here is what it looks like:

screenshot-127 0 0 1_4000-2020 07 21-11_05_34

What do you think of this approach? It does spotlight the stylistic difference between the "tags" ("My untranslated tag", etc.) and the status. Though if we want to change the style of the tags we should probably do that in a separate issue.

phillipgardiner:
@brockfanning good idea . I think that is definitely preferential to how they were before. 

Can you remind me what the 'My untranslated tag' etc are please. We don't use them in the UK I believe, so not entirely sure i have seen them.

brockfanning:
@phillipgardiner There's a short description of them in the docs - see "Tags" in this list: https://open-sdg.readthedocs.io/en/latest/metadata-format/#recommended-special-fields

phillipgardiner:
> @phillipgardiner There's a short description of them in the docs - see "Tags" in this list: https://open-sdg.readthedocs.io/en/latest/metadata-format/#recommended-special-fields

Great thanks, yes it might be good to have a separate ticket to look at the style of these tags alongside everything else to see if they may need changing.

brockfanning:
Ok, sounds good, I'll go ahead and merge this then.

PR #778, opened 2020-07-20
Configurable reporting status title and description

### LINKED ISSUE #638: Add configurable intro text on Reporting status page

brockfanning:
Fixes #638 

Here is a screenshot:
reporting-status-title-and-description

@phillipgardiner What do you think? I used the "lead-copy" class to make the intro text a bit bigger, like in the frontpage banner.

### DIFF
--- a/_includes/components/reporting-status-introduction.html
+++ b/_includes/components/reporting-status-introduction.html
@@ -1,0 +1,11 @@
+<h1>
+  {% if site.reporting_status.title %}
+    {{ site.reporting_status.title | t }}
+  {% else %}
+    {{ page.t.status.reporting_status }}
+  {% endif %}
+</h1>
+
+{% if site.reporting_status.description %}
+<p class="lead-copy">{{ site.reporting_status.description | t }}</p>
+{% endif %}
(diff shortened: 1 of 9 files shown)


phillipgardiner:
@brockfanning @LucyGwilliamAdmin great to get this done. I'm debating with myself weather this should be lead-text or just  normal 16px size. My thinking is on a couple of other pages (mainly indicator pages at the moment) we also have some text at the top of the page. This is currently 16px, so I'm thinking should we be consistent with all these intro texts for the time being (even if we end up making a change to one at a later date). I think it also depends what we expect people to write here, if its just something like 'See how indicators are progressing and what data is collected' type thing, again i'm not sure if this justifies using lead-text size. Happily hear others views on this or discuss on tech call.

brockfanning:
@phillipgardiner Now that I think more about it I agree that it should be normal text. Another argument beyond consistency is that it is fairly easy for countries to add `<p class="lead-copy">` around their description, if they want to get the larger text, but if we hard-code it, then it's not so easy to remove it. So for maximum flexibility we should leave it normal. I've pushed a commit that changes it to a normal `<div>`.

brockfanning:
@LucyGwilliamAdmin @jwestw Are you happy with the documentation/syntax of the configuration settings in this PR?

PR #784, opened 2020-07-22
Optional setting for language toggle as links

### LINKED ISSUE #783: Optional non-dropdown version of language toggle

When there are only 2 or 3 languages, the language toggle may not need to be a dropdown. In these cases the dropdown may be creating unnecessary complication.

In cases where the dropdown is not wanted, it should be possible to optionally replace it with simple links displayed horizontally.

To improve the UX for implementations with only 2 or 3 languages.

None yet.

brockfanning:
Fixes #783 

Here is a screenshot:
language-toggle-links

@phillipgardiner A couple of comments from me:
1. Do we need some way to mark the "active" language?
2. Even though we're working on an alternate header that splits the language toggle away from the menu (in a separate ticket) - if a country wants to keep the current header but switch to links for the language toggle, do we need some way to separate the the first language ("English") from the last menu item ("About")?


phillipgardiner:
@brockfanning are you actively working on this right now? In response you your questions above: Yes and yes. The original mock up was as a talking point only. If we are to proceed it will need a bit more work to answer questions such as yours above. Is this something you need now?

brockfanning:
@phillipgardiner No rush at all - I just threw this up because it was technically easy to do, and I thought it would be good to show how this could be separated from other changes (like the alternate header we discussed).

brockfanning:
@LucyGwilliamAdmin We're discussing this over at #783 but it may help the discussion if you could deploy some feature branches of this. Since there are so many variations, a couple of feature branches would be ideal. If you agree, here is some config to use:

In the first feature branch:
```
header_language_toggle: dropdown
footer_language_toggle: links
```

In the second feature branch:
```
header_language_toggle: links
footer_language_toggle: dropdown
```

LucyGwilliamAdmin:
@brockfanning added links to issue

LucyGwilliamAdmin:
Have just tried this out:

Part of dropdown is covering logo on mobile:
image

LucyGwilliamAdmin:
@brockfanning I'm trying to test this but seem to be getting lots of errors:
https://github.com/ONSdigital/sdg-indicators/runs/1040992408?check_suite_focus=true

I've created pages in all 3 language so not sure what problem is

brockfanning:
@LucyGwilliamAdmin I would double-check the `permalink` in all the pages. For the extra languages it needs to be the normal permalink, but with the language abbreviation. For example if the normal page is `/about` then the French version needs a permalink of `/fr/about`.

LucyGwilliamAdmin:
Ok, have just tried this and something still isn't right.

https://github.com/ONSdigital/sdg-indicators/runs/1042264762?check_suite_focus=true

I'm getting errors such as:

- internally linking to /sdg-indicators/fr/fr/updates/, which does not exist (line 0)
- internally linking to /sdg-indicators/ru/fr/updates/, which does not exist (line 0)

i.e. there's 2 language codes in URL

brockfanning:
@LucyGwilliamAdmin The Russian pages need to have `language: ru` and the French pages need to have `language: fr`. I suspect that may be the issue.

brockfanning:
Right now there's a bug in Open SDG where the language toggle cannot be accessed with the keyboard. Hopefully that's already fixed in this PR, but if not let's remember to fix it.

LucyGwilliamAdmin:
@brockfanning thanks that fixed it.

On mobile, dropdown icon is clashing with logo:
image

Also on mobile, the indenting at the bottom is off a little bit:
image

brockfanning:
@LucyGwilliamAdmin I think that may be coming from this custom CSS: https://github.com/ONSdigital/sdg-indicators/blob/develop/assets/css/custom.scss#L8

Without that I think that the logo may continue to shrink with the screen. I haven't tested though.

LucyGwilliamAdmin:
@brockfanning ah yes - also forgot to update our footer file

LucyGwilliamAdmin:
@brockfanning have just removed the css and the footer override but still seems to be a bit of an indent in the footer:
image

Is that supposed to be there? I remember there being a discussion but might have the outcome wrong (thought no indent was decided)

brockfanning:
@LucyGwilliamAdmin I think that indenting is actually happening in the footer already. I agree it would be nice to be rid of it, but maybe that could be a separate ticket, since it's not caused by this PR (correct me if I'm wrong on that).

Some good news - it looks like this PR fixes the problem of not being able to access the language toggle with the keyboard.

LucyGwilliamAdmin (approved):
Ahh right - that's true

Happy to approve then! 
Thanks for your help solving my issues 😊

PR #785, opened 2020-07-22
Add guidance for making updates by uploading files

LucyGwilliamAdmin:


LucyGwilliamAdmin:
@brockfanning I've added some uploading guidance as discussed on Slack.

The page is getting pretty long now, I'm wondering if it's worth having separate pages for uploading and Prose.io

brockfanning:
@LucyGwilliamAdmin I am fine with splitting the file into multiple files, if you would like to. I'm also OK with doing the "split" by using different sections of the same page, like you have in this PR.

I'm wondering if we might need three parts:
1. Editing the file on Prose.io
2. Editing the file on Github.com
3. Uploading a new version that you have edited offline

Maybe we could start a discussion of whether, and to what extent, we might want to start moving away from Prose.io, even if we don't have an immediate replacement for it. I would interested to know how many users (if any) are actually using Prose.io to update their data/metadata. Seems like a good thing to cover on an upcoming call.

LucyGwilliamAdmin:
@brockfanning I've made some changes to this, but I'm not too sure what's possible with read the docs.

There's now 3 markdown files:

- making-updates.md: just talks about the requirements
- making-updates/uploading-files.md
-making-updates/using-prose.md

I've updated the mkdocs.yml file, and this is the part I'm unsure of:
I was thinking could have making-updates stay where it is with new content, but then have separate pages for the other 2 files at the next layer down

LucyGwilliamAdmin:
@brockfanning I'm thinking this is quite out of date now - I'm wondering if the sections should now be?...

- GitHub
- Upload of files
- Metadata config form

brockfanning:
@LucyGwilliamAdmin These files are in the "data provider" section which makes me think they should focus on data and metadata changes, as opposed to "indicator config" changes, which is where the config form would come in. But then again, where would we put indicator config docs?  In the developer section? Maybe the data provider section is OK - I could be convinced either way.

You didn't mention Prose.io - are you thinking we should remove that? (or is that part of the Github section?)

LucyGwilliamAdmin:
@brockfanning I did leave prose out here but maybe we should continue to provide guidance - I think it would be good to get an idea of who's using it? I know Moldova are but not sure about anyone else.

I see what you're saying about indicator config - it's diff from the actual metadata

LucyGwilliamAdmin:
@brockfanning do you think these instructions are still accurate? I'm wondering if it's worth getting is as it is as think additional to what's up at the moment? especially for the upload file part?

@AnnCorp what are your thoughts? I think this was pretty much based off what I've put together for other countries in past?

LucyGwilliamAdmin:
I'm closing this due to all the changes that have been made to processes

PR #786, opened 2020-07-23
Display a no-data message when chart has no data

### LINKED ISSUE #781: Add a message on the chart whenever there would be no data

When the user gets to a combination of units/disaggregations that contains no data, the chart appears empty. It appears to the user that the chart may be broken.

Instead of appearing as an empty chart, it would be better to display a very clear message. A mockup for this type of message was posted [here](https://github.com/open-sdg/open-sdg/pull/699#issuecomment-630864221). 

To improve the user experience when the data is not 100% complete across all subcategories and units.

We discussed more automated approaches where the combinations of subcategories/units that would result in empty charts would not even display in the first place. But the major downside with this approach is that it results in things changing on behalf of the user, which can lead to confusion and is a bad UX practice. For example, if a certain subcategory is selected, and the unit is changed to a unit without that subcategory, the selected subcategory would be changed without the user's knowledge.

**Additional context**

This is also important for the Series functionality being developed, because this adds another common scenario - where one Series has different units from another Series.

brockfanning:
Fixes #781 

captured (1)

### DIFF
--- a/_includes/assets/js/chartjs/noDataMessage.js
+++ b/_includes/assets/js/chartjs/noDataMessage.js
@@ -1,0 +1,18 @@
+Chart.plugins.register({
+	afterDraw: function(chart) {
+  	if (chart.data.datasets.length === 0) {
+    	// No data is present
+      var ctx = chart.chart.ctx;
+      var width = chart.chart.width;
+      var height = chart.chart.height
+      chart.clear();
+
+      ctx.save();
+      ctx.textAlign = 'center';
+      ctx.textBaseline = 'middle';
+      ctx.font = "normal 40px 'Open Sans', Helvetica, Arial, sans-serif";
+      ctx.fillText('No data to display', width / 2, height / 2);
+      ctx.restore();
+    }
+  }
+});
--- a/_includes/assets/js/chartjs/noDataMessage.js
+++ b/_includes/assets/js/chartjs/noDataMessage.js
@@ -47,6 +47,11 @@ Chart.plugins.register({
         yLine += lineHeight;
       }
       ctx.restore();
+
+      $('#selectionsChart').addClass('chart-has-no-data');
+    }
+    else {
+      $('#selectionsChart').removeClass('chart-has-no-data');
     }
   }
 });
--- a/_includes/assets/js/chartjs/noDataMessage.js
+++ b/_includes/assets/js/chartjs/noDataMessage.js
@@ -17,10 +17,17 @@ function getTextLinesOnCanvas(ctx, text, maxWidth) {
   return lines;
 }
 
+// This plugin displays a message to the user whenever a chart has no data.
 Chart.plugins.register({
   afterDraw: function(chart) {
     if (chart.data.datasets.length === 0) {
-      // No data is present
+
+      // @deprecated start
+      if (typeof translations.indicator.data_not_available === 'undefined') {
+        translations.indicator.data_not_available = 'This data is not available. Please choose alternative data to display.';
+      }
+      // @deprecated end
+
       var ctx = chart.chart.ctx;
       var width = chart.chart.width;
       var height = chart.chart.height
--- a/_includes/assets/js/chartjs/noDataMessage.js
+++ b/_includes/assets/js/chartjs/noDataMessage.js
@@ -1,3 +1,22 @@
+function getTextLinesOnCanvas(ctx, text, maxWidth) {
+  var words = text.split(" ");
+  var lines = [];
+  var currentLine = words[0];
+
+  for (var i = 1; i < words.length; i++) {
+      var word = words[i];
+      var width = ctx.measureText(currentLine + " " + word).width;
+      if (width < maxWidth) {
+          currentLine += " " + word;
+      } else {
+          lines.push(currentLine);
+          currentLine = word;
+      }
+  }
+  lines.push(currentLine);
+  return lines;
+}
+
 Chart.plugins.register({
   afterDraw: function(chart) {
     if (chart.data.datasets.length === 0) {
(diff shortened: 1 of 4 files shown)


brockfanning:
@LucyGwilliamAdmin @jwestw @phillipgardiner @AnnCorp It turns out that this is trivial with just a bit of code (as you can see in this PR). The hardest part - I think - is going to be deciding on what the message should say, and then translating it. So, the floor is open to ideas on what it should say. One option is to use the text from this original mockup: https://github.com/open-sdg/open-sdg/pull/699#issuecomment-630864221

AnnCorp:
@brockfanning I think the text in the original mockup looks fine if we think 'sidebar'; is the right term - @phillipgardiner any thoughts?

brockfanning:
The only potential issue with "sidebar" I think imagine is if the user is on mobile, where's not technically a sidebar.

It's worth detailing the user actions that can result in a chart without data. As far as I know, there are 2 possible scenarios right now:

1. The indicator starts off with a disaggregation (aka "subcategory") already selected, and then the user unchecks that disaggregation.
2. The indicator has multiple units, but one unit only has data for particular disaggregations, and the user selects that unit without having the necessary disaggregations checked.

Also, once we introduce the Series functionality, there will be 2 more:

3. The indicator has multiple series, but one series only has data for a particular unit, and the user selects a different unit.
4. The indicator has multiple series, but one series only has data for a particular unit, and the user selects that series without having the particular unit selected.

brockfanning:
@AnnCorp @phillipgardiner I added links to examples of the scenarios above.

phillipgardiner:
@brockfanning @AnnCorp Ideally all error messages should let users know exactly where the error was, why it happened, and what to do about it. 
This would probably involve having a different error message for each of the above scenarios. However would we know which one the users has done to see the error message? For this reason we may have to have something slightly more generic to cover all scenarios. I would move away from needing to put anything like sub-categories, sidebar, legend (all of which are currently used on our page) in the error message as I am assuming (due to a previous fix) that no chart will ever appear blank, and therefore the users will by default of seeing the error message have already known where to make their selections.

I therefore propose something like **'This data is not available. Please choose alternative data to display.** 

This issue needs looking at more in the future as ideally you would stop the user from committing the error in the first place, thus not requiring an error message at all. To do this we would need to look at how the drop downs etc currently work.

brockfanning:
I've put up PR in sdg-translations with Google translations of the phrase: https://github.com/open-sdg/sdg-translations/pull/137

LucyGwilliamAdmin:
@brockfanning can see the translations PR has been merged so is this ready for review?

brockfanning:
@LucyGwilliamAdmin Yes, it should be all set now. It depends on the latest sdg-translation code, but I also added in some fallback text in English.

brockfanning:
@LucyGwilliamAdmin Good catch - I'll see if I can expand this PR to cover the tables as well.

brockfanning:
@LucyGwilliamAdmin This is ready to test on tables now. From what I can tell, we actually already had code that was supposed to be doing this on tables, but it was not working.

LucyGwilliamAdmin:
@phillipgardiner @brockfanning  here's feature branch for this:
http://sdgdev-813006012.eu-west-1.elb.amazonaws.com/no-data-message/3-3-3/

Under Units type click Percentage (%) of total cases.
Chart tab looks good
For table tab, I'm wondering if we should have bigger text?

brockfanning:
@phillipgardiner @LucyGwilliamAdmin I think bigger text would be good. Here's an example using an H3 in the table:

screenshot-127 0 0 1_4000-2020 08 12-08_13_54

LucyGwilliamAdmin:
@brockfanning @philipashlock  think that's better - in the screenshot I can see the footers aren't there, but on feature branch they're there for chart and table - do you think they shouldn't show when message shows?

brockfanning:
@LucyGwilliamAdmin That makes sense to me - just pushed an update to hide the table/chart footers when there is no data.

LucyGwilliamAdmin:
@brockfanning sorry, I'm now thinking about the "Download Chart Image", "Download Chart CSV" and "Download Table CSV" - do we also want to hide those when message is shown?

@phillipgardiner also what are your thoughts?

phillipgardiner:
@brockfanning @LucyGwilliamAdmin Looking good, however as @LucyGwilliamAdmin says above ideally we would not allow them to download 'Download Chart Image", "Download Chart CSV" if there is no data, as effectively they are downloading a blank file (same goes for the table buttons)'. Don't know if this is part of this ticket or a separate ticket though?

Is it possible the "Download Source CSV" is still relevant to be displayed though even when no data is displayed?

phillipgardiner:
@brockfanning @LucyGwilliamAdmin Also I may be missing something in the way maps work, but would the data be available on a map if its not available as a chart or table (e.g. should a message not appear on maps saying the data does not exist?)

LucyGwilliamAdmin:
@phillipgardiner  that's a good point - in the case of 3.3.3 on that feature branch, the data showing on the map is the default unit type - although think this may need some more thought, as not sure if actually possible to have different map depending on what series or unit is selected

brockfanning:
@LucyGwilliamAdmin @phillipgardiner To play devil's advocate here, would it be over-engineering to hide certain download buttons when there is no data? It's true that the user would get an empty CSV file, or a blank chart image, but I think the user would expect that, given what they see on the screen when they click the button. I just worry that hiding/showing certain buttons will cause a bit of confusion without much benefit. But from a technical perspective I'm sure it is possible.

Regarding the maps - right now maps are completed separate from the units/subcategories/etc in the left sidebar. There's a lot we could discuss there, but that would definitely be a separate issue. (this is related to a work-in-progress PR that's open: #663)

phillipgardiner:
@brockfanning @LucyGwilliamAdmin After reflection I think you make a good point @brockfanning about the download buttons. I'm therefore happy for them to stay and I will just keep an eye out in usability testing etc to check for any confusion etc.

PR #800, opened 2020-08-03
Add jekyll description and pros and cons to docs

jwestw:


brockfanning:
@jwestw Just letting you know that I'll have some feedback on this in the next couple of days.

docs/reasons-for-using-jekyll.md
```diff
@@ -0,0 +1,39 @@
+<h1>Reasons for using Jekyll</h1>
+
+Jekyll is a popular, free and open source static website generator which works like a content management system (for example, Drupal and WordPress) but unlike those systems, Jekyll generates at build-time rather than at request-time. Jekyll creates static websites with some templating abilities.
```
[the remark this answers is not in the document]
brockfanning: Also maybe a line at the end of this paragraph like: "In this way, Jekyll is similar to other static site generators, like [Next.js](https://nextjs.org/) and [Hugo](https://gohugo.io/)."

docs/reasons-for-using-jekyll.md
```diff
@@ -0,0 +1,39 @@
+<h1>Reasons for using Jekyll</h1>
+
+Jekyll is a popular, free and open source static website generator which works like a content management system (for example, Drupal and WordPress) but unlike those systems, Jekyll generates at build-time rather than at request-time. Jekyll creates static websites with some templating abilities. 
+
+## Advantages of Jekyll
```
brockfanning, line 5: Since all of these benefits would apply to any static site, maybe change this heading to "Advantages of static sites"?

Also perhaps a quick introduction like: "Because it is built with Jekyll, the Open SDG platform gets all the advantages of using a static site."

docs/reasons-for-using-jekyll.md
```diff
@@ -0,0 +1,39 @@
...
+
+The dynamic website server involves several software components working together, and if any one component fails the website will fail to be served properly. By contrast a web server, such as one provided by Jekyll, serves only static files, so is less likely to fail. 
+
+### Security 
+
+Security risks such as SQL injection attacks are impossible on a static website since there is no dynamic database that an attacker can exploit. 
+
+### Version-controlled 
+
+Jekyll allows all of the content for Open SDG platforms to be hosted and version-controlled on GitHub.
```
brockfanning, line 25: Static sites are a collection of files, which makes them easy to maintain using open-source version-control software like Git. Free services like Github.com are a perfect tool for any static site.

docs/reasons-for-using-jekyll.md
```diff
@@ -0,0 +1,39 @@
...
+
+Jekyll is a command-line tool, which could be difficult to use for non-technical users. Although graphical user interfaces for the management of Jekyll do exist, one is not used for Open SDG.  
+
+### Long build time 
+
+Open SDG platforms can take several minutes to build. Slow build times can slow down development (especially when debugging) because the developer has to wait for the build to complete before they can see the result. 
+
+### Limited themes 
+
+Jekyll has limited themes, relative to popular content management systems.
```
brockfanning, line 39: I don't think this is true (Jekyll is probably the most mature of all the static site generators) - but regardless I think it's irrelevant to Open SDG, because Open SDG is itself a theme. It would be technically impossible to combine Open SDG with another Jekyll theme. I'd recommend taking out this part.

LucyGwilliamAdmin (approved):

PR #802, opened 2020-08-04
Add some information about search functionality

LucyGwilliamAdmin:


LucyGwilliamAdmin:
@brockfanning from your reply on Slack, what if I add something like this to the paragraph about fuzzy functionality?

However, the project that is used for fuzzy search has implemented the feature for a set of languages, but not all. Currently, Open SDG languages supported include Arabic, English, French, German, Russian and Spanish.

brockfanning:
@LucyGwilliamAdmin That sounds good. If you think it is worth linking to the full list, you could link [here](https://lunrjs.com/guides/language_support.html). Also it might be worth mentioning what happens for other languages not yet supported. My understanding is that for non-supported languages, search will still work, but it will find fewer results, because the search query must be exact (eg, "smoking" would not match "smoke").

LucyGwilliamAdmin:
@brockfanning ok, thanks - have just edited that paragraph

PR #804, opened 2020-08-05
Download buttons for indicator SDMX

brockfanning:
This is a work in progress. The phrases still need translation before this can be merged.


brockfanning:
Translation PR started: https://github.com/open-sdg/sdg-translations/pull/138

I don't think this needs translator approval, since it copied from the nearly-identical CSV download buttons.

brockfanning:
This will be refactored to take a different approach, using `site.data.indicator_downloads`.

PR #805, opened 2020-08-05
Quickstart section revamp with tutorials

brockfanning:
This is ready for review.

### DIFF
--- a/docs/automation/triggered-site-builds.md
+++ b/docs/automation/triggered-site-builds.md
@@ -21,7 +21,7 @@ To solve this problem, the following steps can be taken to set up an automatic s
 1. Go to the site repository.
 1. Under the repository name, click "Settings".
 1. In the left sidebar, click "Secrets".
-1. Click "Add a new secret".
+1. Click "New repository secret".
 1. Under "Name", type the following (case-sensitive): `token`
 1. Under "Value", paste in the access token you copied earlier.
 1. Click the green "Add secret" button.
@@ -31,7 +31,7 @@ Then repeat this for the data repository, as follows:
 1. Go to the data repository.
 1. Under the repository name, click "Settings".
 1. In the left sidebar, click "Secrets".
-1. Click "Add a new secret".
+1. Click "New repository secret".
 1. Under "Name", type the following (case-sensitive): `token`
 1. Under "Value", paste in the access token you copied earlier.
 1. Click the green "Add secret" button.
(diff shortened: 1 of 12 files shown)


brockfanning:
@AnnCorp This is that re-write and simplification of the Quick Start, with additional tutorials, that I mentioned on the collaboration call. I think that it would be good to get this reviewed and merged when possible. If you and/or @jwestw are interested in trying out the Quick Start again, I would recommend trying this version. If you want to try it out before this PR is merged, this easiest way would be to look at the rendered markdown content on my fork. For example here is the Quick Start: https://github.com/brockfanning/open-sdg/blob/quickstart-section-revamp/docs/quick-start.md

After doing the Quick Start also any feedback would be welcome on the other tutorials: https://github.com/brockfanning/open-sdg/tree/quickstart-section-revamp/docs/tutorials

jwestw:
These are the instructions I followed when setting up the recent Rwanda site. They seem clearer and easier than before.

I have gone wrong on the step(s) where the instructions are not in the quickstart but in the Yaml file instead. So where it says "You will see some instructions at the top of the file. Update those few lines as directed" I start to get lost, maybe because it's all in plain text with no bold or headings etc. I forget to insert the URL of the data repo and instead point to the site repo throughout. 
It's not that the instructions are wrong, but do you think it might be better to put the instructions all in one place instead?

brockfanning:
@jwestw Yes that's a great point, I'll see about adding some line-specific instructions here.

jwestw:
When creating the Rwanda site the instructions on favicons didn't work. I managed to upload the correct `favicon.ico` but this didn't show up in the browser, even after 10mins or so, several cache clearings and hard refreshes. So I checked other browsers and sure enough the old favicon was still showing up. I double checked, and this was definitely *not* the picture I could see in `assets/img/favicons/favicon.ico` on `develop`. 

I checked back at https://github.com/sustainabledevelopment-rwanda/sdg-indicators and found that they also had an updated `favicon.png` so I grabbed that too and uploaded it to the new repo. This time it worked! So the instructions are not quite right it seems, or at least replacing `favicon.ico` alone is not sufficient. 

Also Lucy and I are wondering about what the other favicons are and why nothing has to be done with those.

brockfanning:
@jwestw @LucyGwilliamAdmin I've updated that section of the logo/favicon tutorial so that it also renames/uploads the PNG version. I've also clarified (with line numbers) what in the site config needs to be updated in the Quick Start.

To your question about the other versions of the favicons, they're not really being used right now. See this other ticket: #808

jwestw (commented):
I have read through these changes and it looks good. I would approve, but would prefer Lucy to proof read it first.

brockfanning:
@LucyGwilliamAdmin Did you have any feedback for this?

LucyGwilliamAdmin (changes requested):
I've just tested all of these and look pretty good - I've made a few comments about things that may need changing but nothing massive

brockfanning:
@LucyGwilliamAdmin I believe I got all the fixes in the latest commit, and also updated the config in the site starter according to your suggestion.

brockfanning:
@LucyGwilliamAdmin Correction - I had to revert the change in open-sdg-site-starter (".org" vs ".github.io") because it actually caused https://open-sdg.org/open-sdg-site-starter to have problems. The indicator pages weren't loading data, complaining of a cross-origin problem.

LucyGwilliamAdmin (approved):
All looks good to me!

PR #809, opened 2020-08-14
Custom source-file downloads

### LINKED ISSUE #194: SDMX output
Most of this work will happen (and is already happening) in the [sdg-build](https://github.com/open-sdg/sdg-build) project. But we can track concerns here specific to Open SDG.

brockfanning:
Fixes #194 

This is intended to allow for download buttons for indicator-specific SDMX files, although it is general-purpose and any indicator-specific file downloads. Testing is possible, though a little tricky because it depends on PRs in sdg-build and jekyll-open-sdg-plugins. I still need to add documentation to this PR, but here are the steps to test:

1. Add something like this to the data repository config (this would be the config to add a download button for CSV files):
        
        indicator_downloads:
          - button_label: indicator.download_source
            source_pattern: data/indicator_*.csv
            indicator_id_pattern: indicator_(.*)
            output_folder: data-csv

2. Change the data repository requirements.txt to point to my fork of sdg-build:

        git+git://github.com/brockfanning/sdg-build-1@indicator-downloads

3. Change the site repository Gemfile to point to my fork of jekyll-open-sdg-plugins:

        gem 'jekyll-open-sdg-plugins', git: 'https://github.com/brockfanning/jekyll-open-sdg-plugins.git', branch: 'indicator-downloads'

4. Change the site repository config to point to this branch:

        remote_theme: brockfanning/open-sdg@indicator-downloads

Some details on those `indicator_downloads` config options for the data repo:
* button_label: This is intended to be a translation key for the label of the download button
* source_pattern: This is a wildcard-style description of where the source files are in the data repository
* indicator_id_pattern: This is a regular expression for how to convert the source filenames into indicator ids. The `(.*)` is where the indicator id is.
* output_folder: This is the name of a folder to create and place the files in. It doesn't really make a difference what you put here. (And maybe I should automate it.)

The config - especially the regular expression - is a bit technical for my tastes. But I'm not sure how to make it simpler.

The other PRs this depends on are:
* https://github.com/open-sdg/jekyll-open-sdg-plugins/pull/39
* https://github.com/open-sdg/sdg-build/pull/167

### DIFF
--- a/_includes/assets/js/indicatorModel.js
+++ b/_includes/assets/js/indicatorModel.js
@@ -42,6 +42,7 @@ var indicatorModel = function (options) {
   this.graphLimits = options.graphLimits;
   this.stackedDisaggregation = options.stackedDisaggregation;
   this.graphAnnotations = options.graphAnnotations;
+  this.indicatorDownloads = options.indicatorDownloads;
 
   // calculate some initial values:
   this.years = helpers.getUniqueValuesByProperty(helpers.YEAR_COLUMN, this.data);
@@ -282,7 +283,8 @@ var indicatorModel = function (options) {
       graphLimits: this.graphLimits,
       stackedDisaggregation: this.stackedDisaggregation,
       graphAnnotations: this.graphAnnotations,
-      chartTitle: this.chartTitle
+      chartTitle: this.chartTitle,
+      indicatorDownloads: this.indicatorDownloads,
     });
   };
 };
--- a/_includes/assets/js/indicatorView.js
+++ b/_includes/assets/js/indicatorView.js
@@ -454,7 +454,7 @@ var indicatorView = function (model, options) {
 
     this.createDownloadButton(chartInfo.selectionsTable, 'Chart', chartInfo.indicatorId, '#chartSelectionDownload');
     this.createSourceButton(chartInfo.shortIndicatorId, '#chartSelectionDownload');
-    this.createIndicatorDownloadButtons(chartInfo.indicatorDownloads, chartInfo.shortIndicatorId, '#selectionsChart');
+    this.createIndicatorDownloadButtons(chartInfo.indicatorDownloads, chartInfo.shortIndicatorId, '#chartSelectionDownload');
 
     $("#btnSave").click(function() {
       var filename = chartInfo.indicatorId + '.png',
@@ -615,7 +615,7 @@ var indicatorView = function (model, options) {
     $('#tableSelectionDownload').empty();
     this.createDownloadButton(chartInfo.selectionsTable, 'Table', chartInfo.indicatorId, '#tableSelectionDownload');
     this.createSourceButton(chartInfo.shortIndicatorId, '#tableSelectionDownload');
-    this.createIndicatorDownloadButtons(chartInfo.indicatorDownloads, chartInfo.shortIndicatorId, '#selectionsTable');
+    this.createIndicatorDownloadButtons(chartInfo.indicatorDownloads, chartInfo.shortIndicatorId, '#tableSelectionDownload');
   };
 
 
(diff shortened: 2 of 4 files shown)


LucyGwilliamAdmin:
I'm just trying this out with SDMX but I'm getting an error in data build:

https://github.com/LucyGwilliamAdmin/open-sdg-data-starter-1/runs/1040947568?check_suite_focus=true

Worked fine with CSV on UK feature branch

brockfanning:
@LucyGwilliamAdmin I think this file needs to be updated in that branch: https://github.com/open-sdg/sdg-build/blob/rem-input-path-kg2/sdg/outputs/OutputOpenSdg.py

LucyGwilliamAdmin:
@brockfanning ok - have taken another look at this. Seems to be an issue with the button label though:

https://github.com/LucyGwilliamAdmin/open-sdg-data-starter-1/runs/1155725358?check_suite_focus=true#step:5:19

(did test using same button label as CSV and that worked, and I was able to download sdmx file)

brockfanning:
@LucyGwilliamAdmin I think it might have been because I had not merged the sdg-translations PR with the "Download SDMX" translations. I've just merged [that PR](https://github.com/open-sdg/sdg-translations/pull/138), so I'm hopeful that as long as your data repository is pointing to the latest sdg-translations (master), you won't get that error anymore.

I've also refreshed the PRs so that they don't have conflicts and use the latest code.

brockfanning:
@LucyGwilliamAdmin One question I'm wondering about, since this PR is mainly related to SDMX download: Won't we need to somehow link to the DSD? Should the indicator SDMX itself contain some reference to the DSD? And/or do we need a separate download button specifically for the DSD?

brockfanning:
Let's add a default for the `indicator_id_pattern` configuration.

brockfanning:
Also let's add documentation with links to regex helper sites and with examples.

LucyGwilliamAdmin:
> @LucyGwilliamAdmin One question I'm wondering about, since this PR is mainly related to SDMX download: Won't we need to somehow link to the DSD? Should the indicator SDMX itself contain some reference to the DSD? And/or do we need a separate download button specifically for the DSD?

@brockfanning this is a really good point - I think in a couple of places, I've seen the SDMX download be a zip file which contains the DSD and the SDMX file - let me see if I can find some examples

LucyGwilliamAdmin:
I'm also wondering if it's worth speaking to Abdulla to see if he can recommend?

brockfanning:
@LucyGwilliamAdmin That example seems like a good one to follow.

I'm actually re-thinking this whole approach though - let me run this past you:

In the past our ideal solution was more of a "push button" SDMX output, where Open SDG would take the CSV data and simply convert it to SDMX files. As we learned of the difficulties there, we switched to a "lite" version where the SDMX files must be created independently, and then Open SDG only links to them. Unless I'm wrong, the main difficulties that prompted us to make that switch (away from our ideal solution) are:

1. The creation of the DSD file
2. Mapping of potentially custom/non-global disaggregations to dimensions/codes in the DSD

So my thought is: maybe we can create an SDMX output (in the same way we have a GeoJSON output) which _requires_ that those two difficulties are already solved. In other words, it requires that the DSD already exists, and it requires that the data columns/values correspond to dimension/codes in that DSD.

I would still need to do a little technical investigation to see if this is feasible, but I think 3 or 4 hours would be enough to answer that.

From a practical perspective - this would mean that the Kyrgyzstan site could use it to accomplish what this PR was meant to do. Since Kyrgyzstan already has a DSD, and their data is clearly compatible with the dimensions/codes in that DSD (since they already have SDMX files) then in theory this would work fine for them. Similarly, other countries that may not yet have SDMX files could also use this, so long as they create a DSD, and then ensure that their CSV columns/values correspond to dimensions/codes in that DSD.

brockfanning:
As discussed, there would still be a use-case for this feature, so we'll continue with this PR, but will add any SDMX-specific functionality here.

LucyGwilliamAdmin:
@brockfanning here's how this looks for Excel:
https://lucygwilliamadmin.github.io/open-sdg-site-starter-1/en/1-1-1/

Feel like it works really well - do you think it needs more testing? e.g. diff file types?

I don't seem to be getting any error in the data repos anymore

brockfanning:
@LucyGwilliamAdmin I've merged the sdg-build PR, so I believe also that's left is to approve/merge this PR. As discussed on the last call, I don't think testing with other file types is necessary.

LucyGwilliamAdmin (approved):

PR #811, opened 2020-08-16
Support markdown in the description property of some config settings

brockfanning:


LucyGwilliamAdmin (approved):
This worked. See Home, Goals and Reporting Status here:

http://sdgdev-813006012.eu-west-1.elb.amazonaws.com/markdown-in-config-settings/

PR #813, opened 2020-08-16
Site configuration builder form

brockfanning:
This is a proof-of-concept of an idea for having a form that builds (and validates) the site configuration. The target audience here are those users not yet comfortable with YAML syntax. This proof-of-concept takes the user input in a form, and then copies the YAML to their clipboard, so they can paste it into their `_config.yml` file.

The bulk of this functionality is accomplished by describing our site configuration in JSON Schema format. The json file in this PR illustrates that. It does not yet have all of our site configuration - just enough to try this out. With our config in JSON Schema form, a library called "json-editor" automatically creates the form, and also handles the validation.

To test locally, you can do `make serve` and go to the /config page.

Or if you're not working locally, create a new page like so:

```
---
layout: site-config-builder
permalink: /config
---
```

Then visit the /config page. Here's a screenshot of the whole page with all the fields expanded:

screenshot-127 0 0 1_4000-2020 08 16-21_02_21

### DIFF
--- a/_layouts/site-config-builder.html
+++ b/_layouts/site-config-builder.html
@@ -24,7 +24,7 @@
         margin-bottom: 30px;
         padding-bottom: 30px;
       }
-      div[data-schemapath="root"] > .je-panel > .container > div > .columns > div[data-schematype] > .form-group > .je-label {
+      div[data-schemapath="root"] > .je-panel > .container > div > .columns > div[data-schematype] > .form-group > .form-label {
         font-size: 1.2rem;
         font-weight: 500;
       }
--- a/_layouts/site-config-builder.html
+++ b/_layouts/site-config-builder.html
@@ -8,22 +8,25 @@
     <link rel="stylesheet" href="https://unpkg.com/spectre.css/dist/spectre-exp.min.css">
     <link rel="stylesheet" href="https://unpkg.com/spectre.css/dist/spectre-icons.min.css">
     <link rel="stylesheet" href="https://cdn.jsdelivr.net/npm/choices.js/public/assets/styles/choices.min.css">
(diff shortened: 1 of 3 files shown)


LucyGwilliamAdmin:
@brockfanning just had a quick look at this - think it's a great idea.

https://lucygwilliamadmin.github.io/open-sdg-streamlined-qs-site/config

brockfanning:
@LucyGwilliamAdmin Cool, thanks! There's another way to improve on it, by making the form start with all the configurations of the current site. With that, there would not need to be any "import" button.

LucyGwilliamAdmin:
> @LucyGwilliamAdmin Cool, thanks! There's another way to improve on it, by making the form start with all the configurations of the current site. With that, there would not need to be any "import" button.

Yeah, I think that would be good, saves an extra step!

Do you think also in the long term, would it be possible to save changes directly without having to Export and paste into config file?

brockfanning:
I agree it would be ideal to have a "Create pull request" button or something like that. I still need to investigate that, but my suspicion is that that is impossible on a static site. But we could definitely use the same approach on a separate dynamic site, dedicated to this purpose.

Also as a compromise, if we implement #819 then we could have the config form actually download a file, instead of copying to the clipboard. Then the user could upload that file directly in Github, without having to paste any code.

jwestw:
Just seeing this now. It looks like an excellent way to implement this and I think it's an brilliant idea. When I made the Sierra Leone site and more recently the Rwanda site, I managed to mess up the config file so I suggested a "setup wizard" (I know, very 90s) to Lucy to help others to avoid making mistakes. I definitely think it would be best if there were smart suggestions for input values and drop-downs wherever choices can be limited.

brockfanning:
Continuing this in a separate PR: #908

PR #815, opened 2020-08-20
Keep track of colors used in datasets for consistency

### LINKED ISSUE #559: Stop chart colors from changing
We have received feedback from some users that the legend entries are confusing. An often repeated statement is that it is confusing that the colors in the charts and legends change when you select or deselect time series. 
As far as I know the color is determined by a "datasetIndex" (that changes, everytime the selection changes) and the corresponding position in the given list of colors. Based on this approach, I can't think of a way to fix a color to a time series, but maybe someone else has an idea about it?

brockfanning:
Fixes #559 

This could use a lot of testing, with a variety of scenarios (including units), to make sure everything still works. I have not added any automated tests, because it's tough to check colors programmatically.


LucyGwilliamAdmin:
Only had a quick look at this for now as you said it requires lots of testing, but here's a feature branch:
http://sdgdev-813006012.eu-west-1.elb.amazonaws.com/consistent-chart-colors/1-2-1/

Colours are being repeated:
image

Compared with existing site and this issue doesn't occur:
image

brockfanning:
Thanks, I'll get back to this one.

brockfanning:
@LucyGwilliamAdmin I've made an attempt to fix that duplicate color problem. This is ready again for review.

LucyGwilliamAdmin:
@brockfanning something weird is happening now.

1. Go to this page: http://sdgdev-813006012.eu-west-1.elb.amazonaws.com/consistent-chart-colors/1-2-1/
2. Expand Sex and select Female and then Male
3. Expand Age and select 15 and under, 16 to 24, 25 to 49
4. Try to select 50 to 64, checkbox doesn't become checked but now max lines warning is displayed and "50 to 64" and "50 to 64, Female" display (despite warning) and some of the Male breakdowns are removed. Colour is duplicated for "16 to 24, Female" and "50 to 64, Female"

And also at some point this error is happening:
image

brockfanning:
Before I do more work on this, I'm thinking that there may be a fundamental conflict here:

1. We have a limited number of colors (by default 12: 6 colors plus 6 striped versions)
2. There can be unlimited numbers of disaggregations
3. We want each disaggregation to have a consistent color

For example let's say a user selects 12 regions, ending with "London" which appears as striped pink. The other 11 regions are using the other 11 possible colors. So at this point, all of the colors are taken.

Then they select a 13th region: "East Midlands", and they get the error message for too many choices. So they uncheck "London". 

At this point, we would expect "East Midlands" to appear instead of "London" - but what color should it be? We can't use striped pink, or we'll be changing what London was.

LucyGwilliamAdmin:
@brockfanning yes - that's true. I'm not sure how this would be possible then, as each disagg combination would need it's own unique colour right? and there's some indicators which have loads of options and there's just not that many unique colours

brockfanning:
Yeah. As I was developing this, I assumed that we could assign the unique colors dynamically, one at a time, as the user selected them. But that breaks down when the user gets to the maximum and then unchecks something.

brockfanning:
Maybe there is a way to programmatically "bump off" an unused color in these cases. It would be a shame to lose the whole thing because of a rare edge case. If we're able to do that, colors would mostly be consistent, except for the edge case where a user gets to the maximum and then unchecks something.

brockfanning:
Note to check Firefox - currently there seems to be a bug in Firefox where datasets appear white after a while.

brockfanning:
Note to check Firefox - currently there seems to be a bug in Firefox where datasets appear white after a while.

brockfanning:
Closing in favor of #1011

PR #816, opened 2020-08-20
Add vertical hover lines to line charts

### LINKED ISSUE #637: Vertical line on graphs
Use a vertical line on graphs to allow users to see precisely where the cursor crosses the x-axis

brockfanning:
Fixes #637


LucyGwilliamAdmin:
@brockfanning @phillipgardiner 
feature branch for this:
http://sdgdev-813006012.eu-west-1.elb.amazonaws.com/line-chart-vertical-hover/1-2-1/

Chart tooltips are combined when disaggregations are selected.

I haven't noticed it before but what's the white square in the tooltip next to the disaggregation name?

phillipgardiner:
@brockfanning @LucyGwilliamAdmin What are the white little squares for are they supposed to be the same colour as the lines? Also not sure what colour the line is, but could we check its contrast with white to ensure it passes accessibility contrast ration guidelines please.

brockfanning:
@LucyGwilliamAdmin @phillipgardiner From what I can see those white squares have a border that's the same color as the line. Personally it seems to me like it should be the other way around (color inside the square with a white border).

Regarding the combining of the disaggregations in the single tooltip - is that OK?

Regarding the colors themselves, that is fully configurable by the country. Countries can use any number of colors. The default is these 6:

* #7e984f
* #8d73ca
* #aaa533
* #c65b8a
* #4aac8d
* #c95f44

Side note - the number of colors **doubled** is the maximum number of lines on a chart. It's doubled because after we run out of colors, we repeat them with striped versions.

phillipgardiner:
@brockfanning @LucyGwilliamAdmin you can check colour contrast suing a tool like this: https://webaim.org/resources/contrastchecker/.  I have quickly checked the defaults above and they all pass on a black background (e.g. tool tip) but #4AAC8D and #AAA533 fail on a white background (e.g. chart). So perhaps we need to consider some different defaults for those two? Please feel free to double check as i'm totally trusting what this tool says!

Yes I think it should be a coloured square not a coloured border. But the border could just be the same colour as the square?

brockfanning:
@phillipgardiner Those sound like good candidates for two separate tickets: "Use accessible default chart colors" and "Use disaggregation color in tooltip squares"?

phillipgardiner:
@brockfanning yeah agree although I think there is another ticket to come along shortly from DAC review about not relying on colour alone, which may mean further changes to the chart colours anyway.

brockfanning:
We'll fix the legend colors (squares) as part of this ticket.

brockfanning:
@phillipgardiner I've just pushed up a commit that fixes the white square problem. Here's a screenshot:

tooltip-square-background-color

brockfanning:
@phillipgardiner Also just to make sure it is mentioned - right now the color of the vertical line is hardcoded as that blue color (#07c). There's no particular reason for that - it's just what was in the code that I stole for this solution. :) So feel free to recommend a different color for that vertical line.

phillipgardiner:
@brockfanning looks good. The blue colour passes contrast ratio guidelines so I don't really see any reason to change from that. I did look at the ONS website and they use orange, but interestingly this fails contrast ratio checks. Therefore I would say unless the blue is a colour used elsewhere in the graphs (for example the default chart colours) we can just leave as is.

brockfanning:
* The striped colors do not appear striped in the tooltip
* The blue vertical line can appear on top of the tooltip

brockfanning:
@jwestw @LucyGwilliamAdmin This is ready for re-review.

jwestw:
Feature branch refreshed: http://sdgdev-813006012.eu-west-1.elb.amazonaws.com/line-chart-vertical-hover

phillipgardiner:
@brockfanning @jwestw Has something happened to the blue line? Its looks as if  it is now alot lighter than it used to be and as a result harder to see. It looks as if it is potentially appearing behind the grid lines, instead of in front? Think like this it may struggle with contrast ratios.

brockfanning:
@phillipgardiner I just pushed up a change that I believe helps, without needing to change the width/color. Now the vertical line is being drawn after the grid, but before the dataset/tooltips. Here is a screenshot:

vertical-line-fix

phillipgardiner:
Looks great @brockfanning - thank you

LucyGwilliamAdmin:
Refreshed feature branch: http://sdgdev-813006012.eu-west-1.elb.amazonaws.com/line-chart-vertical-hover/1-2-1/

LucyGwilliamAdmin (approved):
Looks great to me!

Checked on Chrome, Edge, IE and mobile Chrome

PR #823, opened 2020-08-23
Only link indicator name in goal-by-target

### LINKED ISSUE #795: Remove underlines on indicator numbers
On the goals page the indicator numbers are underlined, and they should not be. Removing makes the indicator numbers clearer. The links are not required as users are able to click on the relevant indicator description. Please see attachement
Screenshot (35)

brockfanning:
Fixes #795 

This updates the goal-by-target layout to only have a link on the indicator name, instead of the whole thing (including the number, tags, etc.)

I did not update the original "goal" layout (the one where the indicators are displayed in a grid, without targets) because that would take more effort. For consistency we may want to do that at some point, although I don't believe many (any?) implementations are still using that layout.

### DIFF
--- a/_layouts/goal-by-target.html
+++ b/_layouts/goal-by-target.html
@@ -70,23 +70,23 @@
       {% assign tag_classes = tag_classes | join: " " %}
 
       <div class="col-md-12 {{ tag_classes }} goal-indicator">
-        <a href="{{ indicator.url }}">
-          <span>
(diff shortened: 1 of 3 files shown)


LucyGwilliamAdmin (approved):
Looks good!

Only name is link now:
http://sdgdev-813006012.eu-west-1.elb.amazonaws.com/indicator-name-link/1/

PR #834, opened 2020-08-26
Page titles

### LINKED ISSUE #828: Allow page titles to be configured
Currently all pages except for goal pages and indicator pages have the same title.

This is an accessibilty issue:
Screen reader users rely on the page title along with the page heading (h1) to identify the page and understand the page contents.

Currently code uses title from config. Instead, we could add new field to yaml part of md page files which could be pulled in. Then if that field exists, use that as page title. If doesn't, use field from config

https://github.com/open-sdg/open-sdg/blob/master/_includes/head.html#L19-L27

Page 16 in DAC report

brockfanning:
Fixes #828

### DIFF
--- a/_includes/head.html
+++ b/_includes/head.html
@@ -16,15 +16,15 @@
         <!-- Title and meta description
         ================================================== -->
 
-        <title>
-          {% if page.indicator %}
-            {{ page.t.general.indicator | escape }} {{ page.indicator.number }} - {{ page.t.general.goal | escape }} {{ page.goal.number }}
-          {% elsif page.goal %}
-            {{ page.t.general.goal | escape }} {{ page.goal.number }} - {{ page.goal.short | escape }}
-          {% else %}
-            {{ site.title | escape }}
-          {% endif %}
-        </title>
+        {% if page.indicator %}
+        <title>{{ page.t.general.indicator | escape }} {{ page.indicator.number }} - {{ page.t.general.goal | escape }} {{ page.goal.number }}</title>
+        {% elsif page.goal %}
+        <title>{{ page.t.general.goal | escape }} {{ page.goal.number }} - {{ page.goal.short | escape }}</title>
+        {% elsif page.title %}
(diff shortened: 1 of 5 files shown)


brockfanning:
@LucyGwilliamAdmin This is ready for review - note that it needs the latest code from jekyll-open-sdg-plugins, as you can see here where it references the 'master' branch in Github in the Gemfile: https://github.com/open-sdg/open-sdg/pull/834/files#diff-1dde71f804d0805d6f4281510aea6c2dR7

brockfanning:
@LucyGwilliamAdmin @jwestw I forgot to detail the way to get page titles into automatically-generated pages, like Reporting Status, the Goals page, and the frontpage. You would add a `title` property under the `create_pages` bit of your config.yml. For example, see [here](https://github.com/brockfanning/open-sdg/blob/page-titles/tests/site/_config.yml#L25).

For any pages that are not automatically-generated, you would just make sure to add the `title` property in the "frontmatter" for that page. Eg:

```
title: My page title
layout: page
permalink: /my-page-url
```

jwestw:
This seems to be fixed now. Feature branch for reference:
http://sdgdev-813006012.eu-west-1.elb.amazonaws.com/test_page_titles/

Do we need to add instructions on how to add titles to automatically generated pages?

jwestw (approved):
Works well. 
Feature branch for reference. 
http://sdgdev-813006012.eu-west-1.elb.amazonaws.com/test_page_titles/

LucyGwilliamAdmin:
@jwestw did you try changing the title for any of the non automatically generated pages here:
https://github.com/ONSdigital/sdg-indicators/tree/test_page_titles/_pages

jwestw:
Hi Lucy. After reading your message I have tried this. It works too. For example I changed "Cookies and Privacy" to "Cookies Policy". http://sdgdev-813006012.eu-west-1.elb.amazonaws.com/test_page_titles/about/cookies-and-privacy/

PR #835, opened 2020-08-26
H1 in news and post layouts

### LINKED ISSUE #829: Add h1 heading to news layout so consistent with other pages
Page layout has h1 heading which is pulled from page's title field See [here](https://github.com/open-sdg/open-sdg/blob/master/_layouts/page.html). However, the news layout doesn't have a h1 heading.

Think we need to add same thing from page layout [here](https://github.com/open-sdg/open-sdg/blob/master/_layouts/news.html#L7), which will allow title to be pulled in from countries news page.

Not having a h1 heading is an accessibility issue as screenreaders rely on it to provide context.

Page 18 in DAC report

brockfanning:
Fixes #829

### DIFF
--- a/_layouts/news.html
+++ b/_layouts/news.html
@@ -4,6 +4,9 @@ layout: default
 <div id="main-content" class="container newslist">
   <div class="row">
     <div class="col-md-9 push-md-3">
+      {% if page.title %}
+      <h1>{{ page.title | t }}</h1>
+      {% endif %}
       {{content}}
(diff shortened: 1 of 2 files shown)


LucyGwilliamAdmin (approved):
Title is now being pulled from title yaml field and is marked up as H1

http://sdgdev-813006012.eu-west-1.elb.amazonaws.com/updates-page-h1/updates/

image

PR #836, opened 2020-08-26
Add an anchor for any back-to-top links

### LINKED ISSUE #830: Add element with id="top" to pages
This will allow countries to add accessible "Back to top" links.

Currently, there is no element with id="top", so the same page link moves the viewport to the top of the page but does not change the currently focused element. Keyboard only users will be navigated back down the page to the next links in sequence as soon as they press ‘tab’. Screen reader users using cursor keys will be navigated to the content below the link.

Ensure there is an element which has the ID ‘top’ and that it has a tabindex=-1 to ensure that it is not included in tab sequence but can be focused with the .focus method.

Page 54 on DAC report.

Example from DAC report:
```
<div id="top" tabindex=-1>
    <h1>Publications</h1>
    …
</div>
```

brockfanning:
Fixes #830

### DIFF
--- a/_includes/header.html
+++ b/_includes/header.html
@@ -63,3 +63,4 @@
   </div>
 
 </header>
+<div id="top" tabindex=-1></div>


LucyGwilliamAdmin (approved):
Pages now have element with id="top":
http://sdgdev-813006012.eu-west-1.elb.amazonaws.com/back-to-top-anchor/publications/#top
image

Tested with screenreader and focus moves to top when Back to top is clicked

Issue #832, opened 2020-08-26
Add visually hidden text to provide additional context about reporting status


LucyGwilliamAdmin:
**Issue from DAC report:**
The statuses of the indicators are visually formatted to look like badges which allows sighted users to infer the purpose of the information however, there is no alternative to this visual information.

**Solution from DAC report:**
Please consider adding some visually hidden text to provide the additional context which is provided visually.

Page 60 on DAC report

**Example from DAC report:**

```
<span class="visuallyhidden"> Indicator status: </span>
<span class="status complete">Reported online</span>
```

This would need to implemented on all goal page layouts

LucyGwilliamAdmin:
Fix for this has been merged so I am closing

Issue #831, opened 2020-08-26
Ensure relationship between targets and indicators is programmatically determinable


LucyGwilliamAdmin:
**Issue from DAC report:**
There are rows with numbers indicating hierarchical sections, but this structure of the information is not programmatically determinable. Screen reader users will be relayed all the text within sections and subsections linearly. The context of section 1.1.1 ‘belonging to’ 1.1 and introduced by the text ‘By 2030, eradicate extreme poverty…’ is not presented, nor is the separation of section 1.1 from section 1.2, as indicated visually via a grey bottom border.

Additionally, the relationship between the number ‘1.1.1’, the link and the status badge is indicated by the visual positioning of this information within a row, but this relationship is not programmatically determinable.
The sequential relationship between numbered items is also not programmatically determinable. Additionally, it is not possible for screen reader users to easily navigate to required sections.

**Solution from DAC report:**
As this is a sequential numbered list, please ensure this relationship is programmatically determinable. This can be implemented via ordered list mark-up. An alternative way to programmatically determine the structural hierarchy is to implement headings.

Page 57 on DAC report

**Example from DAC report:**

```
<div class="row no-gutters goal-target">
    <h3 class="col-md-1">
        <span class="visuallyhidden">Target 1.2</span>
        <span aria-hidden="true">1.2</span>
    </h3>
    <div class="col-md-11">…</div>
    <div class="row no-gutters goal-indicator indicator-cards">
        <h4 class="col-md-1 col-md-offset-1">
            <span class="visuallyhidden">Indicator 1.2.1</span>
            <span aria-hidden="true">1.2.1</span>
        </h4>
        <div class="col-md-8">…</div>
        <div class="col-md-2 status-column">…</div>
    </div>
    <div class="row no-gutters goal-indicator indicator-cards">
        <h4 class="col-md-1 col-md-offset-1">
            <span class="visuallyhidden">Indicator 1.2.2</span>
            <span aria-hidden="true">1.2.2</span>
        </h4>
        <div class="col-md-8"><a href="/1-2-2">…</div>
        <div class="col-md-2 status-column">…</div>
    </div>
</div>
```

PR #838, opened 2020-08-26
Refactor of goal-by-target-vertical for accessibility

brockfanning:
This is a rebuild of the `goal-by-target-vertical` layout to address the items in #831 and #832.

Because the markup is completely different here, there may be visual differences, though I've tried to keep them to a minimum. This should be tested for cross-browser issues.

I'm not sure if this approach is exactly what was intended. This approach puts all of the target pieces (number and name) in an `h3` and all of the indicator pieces (number, name, status, and tags) in an `h4`. I also considered, to take the example of the indicators, putting the status/tags outside the heading, and the number/name inside the heading. But here I went with the approach of putting all the pieces inside each heading.

That may be confusing, so here's a pseudo-code version of that.

What I considered:
```
<h3>
    target number
    target name
</h3>
<h4>
    indicator number
    indicator name
</h4>
<p>
    status
    tags
</p>
<h4>
    indicator number
    indicator name
</h4>
<p>
    status
    tags
</p>
etc...
```

What I did instead:
```
<h3>
    target number
    target name
</h3>
<h4>
    indicator number
    indicator name
    status
    tags
</h4>
<h4>
    indicator number
    indicator name
    status
    tags
</h4>
etc...
```

### DIFF
--- a/_layouts/goal-by-target-vertical.html
+++ b/_layouts/goal-by-target-vertical.html
@@ -20,7 +20,7 @@
   </div>
 </div>
 
-<div id="main-content" class="container goal-indicators goal-{{ page.goal.number }}">
+<div id="main-content" class="container">
 
   {{ content }}
 
@@ -30,15 +30,12 @@
     {% assign goal_indicators = page.indicators | where: 'goal_number', page.goal.number | group_by: 'target_number' %}
     {% for group in goal_indicators %}
       {% assign target = group.name | sdg_lookup %}
-      <div class="row no-gutters goal-target">
-        <div class="col-md-1">
-          {{ target.number }}
-        </div>
-        <div class="col-md-11">
-          {{ target.name }}
-        </div>
+      <div class="row no-gutters target-and-indicators">
+        <h3 class="target">
+          <span class="target-number">{{ target.number }}</span>
+          <span class="target-name">{{ target.name }}</span>
+        </h3>
         {% for indicator in group.items %}
-        <div class="row no-gutters goal-indicator indicator-cards">
 
           {% assign status_css = indicator.reporting_status | slugify %}
           {% if indicator.reporting_status == 'notapplicable' %}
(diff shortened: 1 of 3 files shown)


brockfanning:
@LucyGwilliamAdmin @jwestw This is ready for review. Is it possible to put this on a feature branch and show it to the DAC folks? I'm not 100% sure my approach is correct. This is intended to fix both #831 and #832 in the `goal-by-target-vertical` layout.

We'll also need feedback from @phillipgardiner on the design but only after DAC approves the markup.

LucyGwilliamAdmin:
@brockfanning no worries - I will add it to our list to discuss

Link to feature branch:
http://sdgdev-813006012.eu-west-1.elb.amazonaws.com/accessible-goal-by-target/1/

Mark up now looks like this:
image

Some visual changes:

1. Full screen
image
image

2. Smaller screen




3. Mobile

phillipgardiner:
@brockfanning What are the visual changes? I've spotted that the black staus boxes are smaller and there doesn't seem to be as much gap between two status boxes when on top of each other - is there anything else?

brockfanning:
@phillipgardiner I did my best to maintain the style without going to too much trouble. But there may be some unintended differences. I see from Lucy's screenshots that the indentation is now maintained on mobile (which I kind of like). The smaller reporting status was unintentional.

brockfanning:
@phillipgardiner But just wanted to stress that DAC may reject the markup, so I wouldn't want you to spend too much time on this yet.

phillipgardiner:
@brockfanning OK I'll wait on the outcome of DAC then but fundamentally it looks OK.

LucyGwilliamAdmin:
@brockfanning @phillipgardiner response from DAC on this:

I would suggest that the ‘Target 1.1’ and ‘Indicator 1.1.1’ were used so that the headings are descriptive but not too long. Screen reader users filter the page by elements such as headings, so the extra text ‘Target’ and ‘Indicator’ would help users navigating in this way to understand what these numbers represent. Your interpretation is correct, so there would be a list of Targets, each item of this top level list would be another list of indicators. Each item within this Indicators list would be the heading Indicator 1.1.1, the link and the status. However, I don’t think you would need to do both of these things. I discussed a similar issue with a screen reader user before and from that discussion I think it may be easier for screen reader users to navigate headings rather than lists nested within lists as this may become confusing. Especially as the numbered indicators continue from each other.


So I think this means have "1.1" as h3 and add visually hidden text "Target 1.1" and have "1.1.1" as h4 and add visually hidden text "Indicator 1.1.1", then as long as the elements fall within the hierarchy this would be fine i.e. nothing further. Brock is this how you interpret DAC's reply?

brockfanning:
That interpretation sounds right to me, and I've given this a shot. @jwestw This could use a feature branch update when you have a chance.

jwestw:
Feature branch created : http://sdgdev-813006012.eu-west-1.elb.amazonaws.com/jw-test-goal-by-target-vertical/

Not testing done yet.

jwestw:
I see text "1.1" but I don't see any visually hidden text, like "Target 1.1"
image

And same for indicators, I don't see  visually hidden text "Indicator 1.1.1".

Am I missing something in the code?

LucyGwilliamAdmin:
@jwestw @brockfanning I've just had a quick look at this. I am able to see the visually hidden text here:
image

(James, we're using a library so sr-only is stuff just for screenreaders)

However, when I tab through page with screenreader "Target 1.1",  "Indicator 1.1.1", etc. aren't being read out by screen reader. Not sure if it's because the elements aren't focusable?

brockfanning:
@LucyGwilliamAdmin Yes I think you're right - tabbing only goes to focusable items. I think if you want to read out content in between you would use the up/down keys.

LucyGwilliamAdmin (approved):
@brockfanning yes, your right - I've just tried using arrows and I'm able to access non-focusable elements.

For h3 screen reader is reading out "Target ..." and for h4, screen reader is reading "Indicator ..." so happy to approve.

brockfanning:
Thanks very much @LucyGwilliamAdmin.

@phillipgardiner I think we're all clear from a markup perspective. Do you see any visual/design issues in the feature branch?

phillipgardiner:
@brockfanning Looks OK to me 👍

PR #862, opened 2020-09-01
Better keyboard accessibility for tabbed content

### LINKED ISSUE #840: Make sure tab panels are included in page sequence
**Page 55 in DAC report**

**Issue from DAC report:**
Tab-panel content does not take focus upon activation. Tabs have been implemented as same page links, but these do not work and upon activation, focus is taken to the other tabs before the tab-panel content.

**Solution from DAC report:**
Include the tabpanel in the page sequence so that it can be focused by screen reader users. Ensure that focus is taken to the tabpanel upon activation.

**Example from DAC report:**

```
<div class="tab-content">
    <div role="tabpanel" class="tab-pane active" id="national"
tabindex="0">…</div>
</div>
```

Includes data tabpanel, metadata tabpanel and reporting status tab-panel

brockfanning:
Fixes #840

### DIFF
--- a/_includes/assets/js/accessibility.js
+++ b/_includes/assets/js/accessibility.js
@@ -119,3 +119,11 @@ var accessibilitySwitcher = function() {
   };
 
 };
+
+$(document).ready(function() {
+  // Focus on tab content when tab is selected.
+  $(".nav-tabs a").on("shown.bs.tab", function(event) {
+    var href = event.target.getAttribute('href');
+    $(href).focus();
+  });
+})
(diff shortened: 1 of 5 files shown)


LucyGwilliamAdmin (approved):
When tabbing through and then selecting a tab, focus on tab content instead of next tab label:
http://sdgdev-813006012.eu-west-1.elb.amazonaws.com/tab-pane-tabindex/1-2-1

LucyGwilliamAdmin:
@brockfanning for some reason this isn't working on the accessibility feature branch: http://sdgdev-813006012.eu-west-1.elb.amazonaws.com/dac2020/1-2-1/

Any ideas?

PR #863, opened 2020-09-01
Fieldsets around radios and checkboxes

### LINKED ISSUE #842: Include dropdowns and radio buttons within a fieldset with a legend
**Page 83 in DAC report**

**Issue from DAC report:**
The checkboxes in the dropdowns, and the radio buttons under unit types are not contained within a programmatically associated semantic grouping to present their relationship to assistive technology. Screen reader users navigating the page ‘out of context’ filtering the page by element type, will encounter a number of independent checkboxes, but with no additional group context to describe these as additional sub-category options.

**Solution from DAC report:**
Ensure that form groups are programmatically determined by included them within a fieldset with a legend. This will ensure that users navigating ‘out of context’ will be given the additional context of the legend to present the purpose of the components.

**Example from DAC report:**

Dropdowns:

```
<div class="variable-selector" data-field="Sex" data-has-data="true">
    <button class="accessBtn" tabindex="0" aria-expanded="false"><h5>Sex<i
       class="fa fa-chevron-down"></i></h5></button>
    <div class="bar"><div class="selected"></div></div>
    <div class="variable-options" style="display: block;">
         <div>
             <button data-type="select">Select all</button>
             <button data-type="clear">Clear all</button>
         </div>
         <fieldset>
             <legend><span class="visuallyhidden">Add subcategory by Sex</span></legend>
             <label class="default" data-has-data="true">
                <input type="checkbox" value="Female" data-field="Sex">Female
            </label>
            <label class="default" data-has-data="true">
                <input type="checkbox" value="Male" data-field="Sex">Male
            </label>
        </fieldset>
    </div>
</div>
```

Radio buttons:

```
<fieldset>
<div id="units">
    <legend><span class="visuallyhidden">Data units type</span></legend>
    <h4>Units type</h4>
    <label>
        <input type="radio" name="unit" value="English Local Authority 
          Councillors" tabindex="0" checked=""> English Local Authority 
          Councillors
    </label>
    …
</div>
```

brockfanning:
Fixes #842 

Note that for the screenreader-only text, I tried to re-use existing language so that we would not need to do any new translations. So the text is not exactly as recommended by DAC in #842.

I did a quick test locally and did not see any visual problems or functionality issues.

### DIFF
--- a/_includes/components/indicator/fields-template.html
+++ b/_includes/components/indicator/fields-template.html
@@ -10,11 +10,14 @@
           <button data-type="select">{{ page.t.indicator.select_all }}</button>
           <button data-type="clear">{{ page.t.indicator.clear_all }}</button>
         </div>
-        <% _.each(fieldItem.values, function(item) { %>
-          <label>
-            <input type="checkbox" <%=item.checked ? 'checked' : ''%> value="<%=item.value%>" data-field="<%=fieldItem.field%>" /><%=translations.t(item.value)%>
-          </label>
-        <% }); %>
+        <fieldset>
+          <legend><span class="sr-only">{{ page.t.indicator.sub_categories }} - <%=translations.t(fieldItem.field)%></span></legend>
+          <% _.each(fieldItem.values, function(item) { %>
+            <label>
+              <input type="checkbox" <%=item.checked ? 'checked' : ''%> value="<%=item.value%>" data-field="<%=fieldItem.field%>" /><%=translations.t(item.value)%>
+            </label>
+          <% }); %>
+        </fieldset>
       </div>
 
       <% if(allowedFields.indexOf(fieldItem.field) == -1) { %>
--- a/_includes/components/indicator/fields-template.html
+++ b/_includes/components/indicator/fields-template.html
@@ -1,13 +1,13 @@
 <script type="text/template" id="item_template">
   <% _.each(fields, function(fieldItem) { %>
-    <div class="variable-selector<% if(allowedFields.indexOf(fieldItem.field) == -1) { %> disallowed child<% }%>" data-field="<%=fieldItem.field%>">
-      <button class='accessBtn' tabindex='0' aria-expanded='false'><h5><%=translations.t(fieldItem.field)%><i class="fa fa-chevron-down"></i></h5></button>
-      <div class="bar">
-        <div class="selected"></div>
-      </div>
-      <div class="variable-options">
-        <fieldset>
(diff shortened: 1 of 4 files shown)


brockfanning:
Why no series?

LucyGwilliamAdmin:
@brockfanning I'm starting to look at this on UK site but I'm not sure what I'm expecting to hear from screen reader.

http://sdgdev-813006012.eu-west-1.elb.amazonaws.com/sidebar-fieldsets/8-4-1/

On this feature branch, "Sub-categories - Material category" isn't read out until I've expanded the drop down and tabbed to the first item of drop down (Biomass)

"Units type" is read out when I focus on first unit type (Material footprint (Million metric tonnes)).

I'm not actually able to tab through the different unit types, but have compared that with current uk live site and don't seem to be able to do it there either.

brockfanning:
I've moved the "Subcategories - x" legend above the "Select all" and "Clear all" buttons. Still the user needs to tab onto the "Select all" button before the legend is read, and it gets read along with the "Select all" button. I've also changed to commas instead of dashes, which seems to slow down the screenreader. So instead of reading, for example, "subcategories age button select all" it reads "subcategories... age... button select all".

For the unit types, I believe keyboard navigators would use the up/down arrows to go between the unit types.

LucyGwilliamAdmin:
@brockfanning I'm going through this with screenreader:
https://www.w3.org/WAI/tutorials/forms/grouping/#:~:text=The%20element%20provides%20a,in%20the%20group%20are%20required.

Comparing our radio buttons with those in the example, think that's right.

However I think from looking at examples, "Sub categories - x" should be read out when x (Material category in this case) is focused on rather than the "Select all" button

brockfanning:
We'll try putting the fieldset around the button element.

brockfanning:
@LucyGwilliamAdmin @jwestw This is ready to test again. I've changed the fields to have the fieldset around the whole thing, including the dropdown button.

jwestw:
Feature branch refreshed:
http://sdgdev-813006012.eu-west-1.elb.amazonaws.com/sidebar-fieldsets

jwestw:
Looking at the files changed and inspecting the elements in Chrome shows that fieldsets are around the whole radio and check box code. 

image

Using the screen reader on checkboxes seemed to work beautifully. It said what the categories were as they were focused on (I think this is what we were trying to get fixed) what each of the checkboxes correspond to and if it was checked or not.

jwestw (approved):
All looks good to me. See my last comments.

PR #864, opened 2020-09-01
More obvious focus colors

### LINKED ISSUE #849: Change colour of assistive technology focus border
**Page 120 in DAC report**

**Issue from DAC report:**
When tabbing through page with assistive technology, components are highlighted with black border to indicate focus. The black focus highlight is not visible against the black ‘high contrast’ button.

**Solution from DAC report:**
Please ensure that a mechanism is in place to allow users to identify their focus on the page. In this case, the black border is not visible against the black button. The GDS focus highlight is recommended. Consider implementing a consistent visible focus indicator to help users identify when page elements are in focus, using the ‘Focus state’ colour ($govukfocus- colour #FFDD00). For more information please visit [Colour - GOV.UK Design System](https://design-system.service.gov.uk/styles/colour/), Focus state.

**Example from DAC report:**
No example provided

brockfanning:
Fixes #849 

This could use design input, as I arbitrarily chose the colors here. I'm using our link color for the normal focus color, and our button highlight color for the high-contrast focus color. Here's an example of each:

Normal contrast:

screenshot-127 0 0 1_4000-2020 09 01-15_43_30

High contrast:

screenshot-127 0 0 1_4000-2020 09 01-15_43_49

### DIFF
--- a/_includes/search.html
+++ b/_includes/search.html
@@ -1,6 +1,6 @@
 <div id="search" class="menu-target">
   <form action="{{ page.baseurl }}search">
-    <input name="q" id="indicator_search" title="{{ page.t.search.search }}" />
+    <input type="text" name="q" id="indicator_search" title="{{ page.t.search.search }}" />
     <label for="indicator_search">
       <button aria-label="{{ page.t.search.search }}" id="search-btn" type="submit"><i class="fa fa-search" aria-hidden="false"></i></button><span>{{ page.t.search.search }}:</span>
     </label>
--- a/_layouts/goals.html
+++ b/_layouts/goals.html
@@ -7,7 +7,7 @@
     <div class="goal-tiles">
         <div class="row no-gutters">
             {% for goal in page.goals %}
-            <div class="col-xs-4 col-md-2">
+            <div class="col-xs-4 col-md-2 goal-tile">
                 <a href="{{ goal.url }}">
                     <img src="{{ goal.icon }}" id="goal-{{ goal.number }}" alt="{{ goal.short | escape }} - {{ page.t.general.goal }} {{ goal.number }}" />
                 </a>
@@ -18,7 +18,7 @@
             out more symmetrically.
             {% endcomment %}
             {% if page.goals.size == 17 %}
-            <div class="col-xs-4 col-md-2">
+            <div class="col-xs-4 col-md-2 goal-tile">
                 <img src="{{ site.goal_image_base }}/{{ page.language }}/18.png" id="goal-18" alt="The Global Goals for Sustainable Development" />
             </div>
             {% endif %}
--- a/_sass/base/_typography.scss
+++ b/_sass/base/_typography.scss
@@ -108,3 +108,6 @@ body.contrast-high {
     box-shadow: 0 -2px $focusOutlineColor-highContrast,0 4px $focusColor-highContrast !important;
   }
 }
+input[type="radio"]:focus, body.contrast-high input[type="radio"]:focus {
+  box-shadow: 0px 0px 0px 4px $focusOutlineColor !important;
+}
--- a/_sass/base/_typography.scss
+++ b/_sass/base/_typography.scss
@@ -111,3 +111,9 @@ body.contrast-high {
 input[type="radio"]:focus, body.contrast-high input[type="radio"]:focus {
   box-shadow: 0px 0px 0px 4px $focusOutlineColor !important;
 }
+input[type="text"]:focus,
+body.contrast-high input[type="text"]:focus {
+  background-color: $backgroundColor !important;
+  box-shadow: none !important;
+  outline: 3px solid $focusOutlineColor !important;
+}
--- a/_sass/base/_typography.scss
+++ b/_sass/base/_typography.scss
@@ -37,7 +37,6 @@ h3{ font-size: $h3-fontSize; }
 
   img {
     max-width: 100%;
-    margin-top: 10px;
   }
 }
 
--- a/_sass/base/_typography.scss
+++ b/_sass/base/_typography.scss
@@ -88,22 +88,22 @@ h3{ font-size: $h3-fontSize; }
 [href]:focus,
 [tabindex]:focus,
 iframe:focus,
-.navbar #search input:focus,
-.navbar #search button:focus {
-  color: $focusColor;
-  outline: .25rem solid $focusOutlineColor;
-  background-color: $focusBackgroundColor;
-  outline-offset: 0;
+input:focus,
+button:focus {
+  color: $focusColor !important;
+  outline: .25rem solid $focusOutlineColor !important;
+  background-color: $focusBackgroundColor !important;
+  outline-offset: 0 !important;
 }
 body.contrast-high {
   [contentEditable="true"]:focus,
   [href]:focus,
   [tabindex]:focus,
   iframe:focus,
-  .navbar #search input:focus,
-  .navbar #search button:focus {
-    color: $focusColor-highContrast;
-    outline: .25rem solid $focusOutlineColor-highContrast;
-    background-color: $focusBackgroundColor-highContrast;
+  input:focus,
+  button:focus {
+    color: $focusColor-highContrast !important;
+    outline: .25rem solid $focusOutlineColor-highContrast !important;
+    background-color: $focusBackgroundColor-highContrast !important;
   }
 }
--- a/_sass/base/_typography.scss
+++ b/_sass/base/_typography.scss
@@ -90,9 +90,11 @@ iframe:focus,
 input:focus,
 button:focus {
   color: $focusColor !important;
-  outline: .25rem solid $focusOutlineColor !important;
+  outline: 3px solid transparent !important;
   background-color: $focusBackgroundColor !important;
   outline-offset: 0 !important;
+  box-shadow: 0 -2px $focusOutlineColor,0 4px $focusColor !important;
+  text-decoration: none !important;
 }
 body.contrast-high {
   [contentEditable="true"]:focus,
@@ -102,7 +104,7 @@ body.contrast-high {
   input:focus,
   button:focus {
     color: $focusColor-highContrast !important;
-    outline: .25rem solid $focusOutlineColor-highContrast !important;
     background-color: $focusBackgroundColor-highContrast !important;
+    box-shadow: 0 -2px $focusOutlineColor-highContrast,0 4px $focusColor-highContrast !important;
   }
 }
--- a/_sass/base/_typography.scss
+++ b/_sass/base/_typography.scss
@@ -83,3 +83,24 @@ h3{ font-size: $h3-fontSize; }
   clip: rect(0,0,0,0);
   border: 0;
 }
+
+[contentEditable="true"]:focus,
+[href]:focus,
+[tabindex]:focus,
+iframe:focus,
+.navbar #search input:focus,
+.navbar #search button:focus {
+  outline: .25rem solid $focusColor;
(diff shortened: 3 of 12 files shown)


phillipgardiner:
@brockfanning GDS guidance specifies using #FFDD00 [https://design-system.service.gov.uk/styles/colour/](url). It looks as if they ensure it is black text on a yellow focus to ensure acceptable contrast ratio. Therefore they change the text colour on focus too. If used we could use this colour could be used on both our normal pages and our high contrast pages. You can see an example by tabbing through this page: [https://www.gov.uk/service-manual](url). Perhaps we can discuss the in and outs further on the tech call today @AnnCorp?

phillipgardiner (commented):
@brockfanning GDS guidance specifies using #FFDD00 [https://design-system.service.gov.uk/styles/colour/](url). It looks as if they ensure it is black text on a yellow focus to ensure acceptable contrast ratio. Therefore they change the text colour on focus too. If used we could use this colour could be used on both our normal pages and our high contrast pages. You can see an example by tabbing through this page: [https://www.gov.uk/service-manual](url). Perhaps we can discuss the in and outs further on the tech call today @AnnCorp?

brockfanning:
We'll mimic https://www.gov.uk/service-manual

brockfanning:
@LucyGwilliamAdmin I believe this one is ready for a feature branch.

@phillipgardiner I suspect there may be many edge cases where we need to tweak things, so I think this will need a lot of manual review, tabbing through the various types of pages in both normal and high-contrast modes.

LucyGwilliamAdmin:
@brockfanning @phillipgardiner:

Feature branch:
http://sdgdev-813006012.eu-west-1.elb.amazonaws.com/new-focus-colour/

Only had quick look so far but some comments:
Top-left logo not focussing properly:
image

Horizontal white space missing between goals grid:
image

LucyGwilliamAdmin:
Goals are also joined on Goals page: http://sdgdev-813006012.eu-west-1.elb.amazonaws.com/new-focus-colour/goals/

LucyGwilliamAdmin:
Focus is overhanging on goal icon on indicator pages:
image

LucyGwilliamAdmin:
For tags in high-contrast mode, there isn't much difference between tags colour and focus colour:
image

brockfanning:
@LucyGwilliamAdmin I pushed another commit that is intended to help with the logo and goal-icon issues.

LucyGwilliamAdmin:
Feature branch has been updated:
http://sdgdev-813006012.eu-west-1.elb.amazonaws.com/new-focus-colour

I think all of the things I brought up have been fixed. @phillipgardiner are you also taking a look at this one?

phillipgardiner:
@LucyGwilliamAdmin @brockfanning Yeah I will take a look through but it will have to be tomorrow now.

brockfanning:
I didn't address the similarity between the focus yellow and the high-contrast tag yellow. This same problem affects the active tabs on indicator pages in high-contrast mode.

@LucyGwilliamAdmin Do you mind switching the feature branch to use the default contrast icons? Ie, remove `contrast_type: long` from the config. I think this PR also causes a problem with the "A A" icons that we'll need to look at.

LucyGwilliamAdmin:
> @LucyGwilliamAdmin Do you mind switching the feature branch to use the default contrast icons? Ie, remove `contrast_type: long` from the config. I think this PR also causes a problem with the "A A" icons that we'll need to look at.

@brockfanning done 👍

phillipgardiner:
@brockfanning @LucyGwilliamAdmin Had a look through in the end and got a couple of things to add, as well as the tag issue.

1) I think we may need to invert the high contrast SDG logo. Currently the contrast of yellow with with white is hard to read. Looking at the gov.uk wesbite they invert thier white writing to black.

Screenshot (45)

2) When the goals images are highlighted, it is really hard to see which one is highlighted (e.g. in this screenshot it is the 3rd one along. But i'm not sure what we can do about this? 
Screenshot (46)

And finally I noticed on GOV.UK their exists an underline which is thick and black? Is there code different somehow?
Screenshot (47)

Sorry got to shoot know but i'll pick up any comments tomorrow.

brockfanning:
I just pushed some commits to add the black underline and to make the goal icon outlines much thicker. (Though I couldn't figure out how to add the black underline to the goal icons.)

I'm not sure offhand what the best solution is for the issue with the logo.

phillipgardiner:
@brockfanning quick work indeed 👍 

@jwestw Are you able by any chance to update the feature branch to reflect these latest changes? Feature branch http://sdgdev-813006012.eu-west-1.elb.amazonaws.com/new-focus-colour

@brockfanning @AnnCorp  in regards to the logo, I guess the possible solutions are 1) We leave how it is and see if it is a problem to DAC. 2) I could probably provide an inverted logo image if there is a way of swapping it for the original on focus?

LucyGwilliamAdmin:
E.g like this:

AnnCorp:
@phillipgardiner happy to create, if there isn't already an official inverted version available.

brockfanning:
If I understand right, there is already an inverted version here: https://github.com/ONSdigital/sdg-indicators/blob/develop/assets/img/high-contrast/SDG_logo.png

However it's a PNG with white text on a transparent background (which is why it's hard to see in the Github link above). So when the focus yellow gets applied, it appears as white text on a yellow background.

One solution would be to get the PNG a black background instead of transparent. But this is something that many other countries will experience too, so I would hope there is a CSS-based solution. Maybe we can write CSS code that specifically targets the logo and gives it a yellow outline while leaving the background black.

jwestw:
> @jwestw Are you able by any chance to update the feature branch to reflect these latest changes? Feature branch http://sdgdev-813006012.eu-west-1.elb.amazonaws.com/new-focus-colour

That should be done now. 
http://sdgdev-813006012.eu-west-1.elb.amazonaws.com/new-focus-colour/

brockfanning:
Thanks @jwestw looks good.

Regarding my note above about a logo-specific focus treatment, here are screenshots of the logo for only an outline and no background color:

logo-focus

logo-focus-highContrast

phillipgardiner:
@jwestw thanks for updating the feature branch 👍 

@brockfanning loving the black underlines, couple of comments form me though of things I've spotted..

- Skip to main content should be link (e.g. underlined) this means visually people know its a link too
- The yellow highlight no longer appears around the goal icons on the reporting status page. Incidentally I think the thicker border works great, i don't think it is a biggy that the black lines do not appear in this instance as that is a specific GDS accessibility thing
- In regard to the logos above, I think this may be one that needs checking with DAC. I believe the black logo would be OK like this, but white logo would might fail as that yellow on white background may fail contrast ratio. Like I say not sure so needs checking with DAC around what you should do with logos, as I think the rules are a bit confusing (e.g. do you have to change the logo or can you just highlight it)

LucyGwilliamAdmin:
Just noticed a couple of things with this:

1. When I click on Sex or Age, they become highlighted (so even not when tabbing through with keyboard - I'm not sure if this was expected so raising just in casw:
http://sdgdev-813006012.eu-west-1.elb.amazonaws.com/new-focus-colour/1-2-1/

2. When tabbing through here and get to Material footprint (Million metric tonnes), only the radio button is highlighted and is just very slightly visible:
http://sdgdev-813006012.eu-west-1.elb.amazonaws.com/new-focus-colour/8-4-1/

brockfanning:
I just pushed up a few more commits based on the feedback:
* Special treatment of radio buttons (I gave it a bit more yellow outline)
* Added the yellow outline to reporting status goal icons
* I went ahead and added the special treatment for the logo (so that DAC has something to try out)

Regarding the skip-to-content link, actually the GDS focus code specifically hides the underlines on links (aka "text decoration"). So I think that would be a departure from the GDS.

Lucy to your comment about the things being highlight just from clicking - yes that is a good point to remember. Elements can receive "focus" in ways other than keyboard navigation. Just clicking any element will give it focus usually. This means that, in the case of links, this focus style will always be visible on click (if only for a split second). And for things that are clicked as part of interaction (like drop-downs, buttons, etc.) they will remain focused until the user clicks on something else. (This may be why we had such a subtle focus outline before.)

Among the things still to do, we still need to figure out how to handle the high-contrast toggle. The problem is that the normal A and the inverted A both look the same when focused. We need to decide how to make them distinct even when one is focused, right?

AnnCorp:
@brockfanning re the last point, is it worth you contacting DAC directly for some advice on this point?

brockfanning:
I think that is more of a design question that we can figure out for ourselves. I've just pushed up an attempt along with some fixes for the language toggle and contrast toggle.

@jwestw Whenever you're able I think this is ready for another feature branch update.

phillipgardiner:
@jwestw @brockfanning is there another feature branch to check all this on?

@brockfanning one thing iv'e notice is on the gov.uk implementation of tabs: https://design-system.service.gov.uk/components/tabs/. When the tab is focused, it is only the words that are highlighted rather than the whole tab. Is this possible on our tabs please?

jwestw:
Feature branch is refreshed: http://sdgdev-813006012.eu-west-1.elb.amazonaws.com/new-focus-colour

brockfanning:
Thanks very much @jwestw.

@phillipgardiner I did manage to get the search bar change in there, so it should be available to review in the feature branch. I took a quick look at the tabs and at first glance it seems like it might be technically difficult - at least I couldn't figure it out quickly. So that would take more time I think.

phillipgardiner:
@brockfanning not sure its made it through yet but one thing I have spotted is on the indicator page, the goal icon is highlighted. This needs the thick border treatment the same as on the homepage etc - sorry!

brockfanning:
@phillipgardiner Good catch, I just pushed up a commit to help with that.

brockfanning:
Still need to add special treatment for checkboxes.

brockfanning:
@LucyGwilliamAdmin @jwestw This is ready again for review.

jwestw:
http://sdgdev-813006012.eu-west-1.elb.amazonaws.com/new-focus-colour

jwestw:
I think these are all the checks:

- On indicator pages the goal icon is highlighted- needs thick border treatment the same as other pages *DONE*
- Special treatment of radio buttons - more yellow outline *DONE (I think) I've checked this on check-box buttons and radio buttons. The button itself is highlighted in thick yellow, but the text is not*
- Added the yellow outline to reporting status goal icons *DONE*
- Special treatment for the logo  (My interpretation of this was that there was meant to be a thick yellow highlight around the logo on every page.)  *DONE* 
- Focus fixes for language toggle and contrast toggle (I've only tests the contrast toggle) *DONE*
- Special focus treatment of text input (If this means the search box (where text can be input) is highlighted properly, then it's *DONE* )

Looks good in:
- Edge
- IE 
- Chrome

Also looks fine in Windows High contrast mode

jwestw (approved):
Approving on basis of the last checks I've done

phillipgardiner:
@brockfanning  in regard to previous comment.....one thing iv'e notice is on the gov.uk implementation of tabs: https://design-system.service.gov.uk/components/tabs/. When the tab is focused, it is only the words that are highlighted rather than the whole tab. Is this possible on our tabs please?

Did you create a separate ticket for this? If not happy to do it.

brockfanning:
@phillipgardiner No I didn't make a ticket for that, thanks for creating. It wasn't immediately obvious to me how fast that would be or if it's even feasible. Usually CSS fixes are pretty easy to figure out but this one gave me some trouble. Because it may take more time I thought it should probably be separate so that it doesn't delay the accessibility tickets.

PR #865, opened 2020-09-01
Use darker text color and default to white text in high-contrast

### LINKED ISSUE #852: Make main text colour black
Low vision users struggled to read the font. Please consider implementing a heavier font with a greater contrast.

brockfanning:
Fixes #852

### DIFF
--- a/_sass/base/_typography.scss
+++ b/_sass/base/_typography.scss
@@ -1,6 +1,10 @@
 body{
   font-family: 'Open Sans', Helvetica, Arial, sans-serif;
   font-size: $fontSize;
+  color: $text-color;
+}
+body.contrast-high {
+  color: $text-color-highContrast;
 }
 
 h1{ font-size: $h1-fontSize; }
--- a/_sass/layouts/_header.scss
+++ b/_sass/layouts/_header.scss
@@ -341,6 +341,11 @@ body.contrast-high {
         }
       }
     }
+    #search {
+      input, label {
+        color: $color-dark-highContrast;
+      }
+    }
   }
 
   @media screen and (max-width: 768px) {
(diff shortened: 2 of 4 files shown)


brockfanning:
Use #0b0c0c

brockfanning:
Default to white on high-contrast.

brockfanning:
This is ready for review again.

jwestw (changes requested):
I have tested this. In normal contrast mode it appears that the text is black (visual inspection) though when I inspect element, it doesn't show what colour it is. I see that #0b0c0c refers to a pure black though. 

In high contrast mode I see no greys any more, only black, yellow and white. However @brockfanning I have spotted a problem this has caused. In the high contrast mode if you try to search the search bar is white but so is the text, meaning it cannot be seen. Can we change that?

brockfanning:
Thanks very much @jwestw. This is a good example of things that may result from the change of defaulting to white text in high-contrast mode. I've pushed a commit to deal with that one (search bars) but there may still be others.

brockfanning:
@jwestw Is this on a feature branch? If so, maybe Phil could also help look for issues?

LucyGwilliamAdmin:
@brockfanning @phillipgardiner James' feature branch: http://sdgdev-813006012.eu-west-1.elb.amazonaws.com/jw-black-text/

LucyGwilliamAdmin (approved):
I've also just looked through and can't see any other issues

phillipgardiner:
@brockfanning @LucyGwilliamAdmin @jwestw 

The only additional thing I have spotted is the the alpha banner text is set to: #3D3D3D this would need changing to: #0b0c0c.

brockfanning:
@phillipgardiner This PR has already been merged, but I made #883 to fix that issue.

phillipgardiner:
@brockfanning OK, thanks

PR #871, opened 2020-09-04
More efficient unique function for better support of large datasets

### LINKED ISSUE #818: Support larger datasets

As the dataset gets larger, performance will get worse. This is unavoidable, but maybe there are some inefficiencies in our code, or architectural factors, that are limiting us too much. We should investigate what the "bottlenecks" are when Open SDG tries to load a very large dataset.

If possible, better performance with larger datasets.

When the platform struggles with larger datasets it appears to the user that the site is broken.

Alternatively we could recommend a maximum number of rows of data.

**Additional context**

Here's an example of a very large dataset to test against: https://onsdigital.github.io/sdg-indicators/4-2-1/

brockfanning:
Fixes #818


brockfanning:
@jwestw Could you try a feature branch for this one? This should make a dramatic improvement in the load time for indicator 4.2.1.

_includes/assets/js/model/utils.js
```diff
@@ -7,10 +7,12 @@
...
-  return rows
-    .map(function(row) { return row[prop]; })
-    .filter(isElementUniqueInArray)
-    .filter(function(row) { return row; })
+  var uniques = new Set();
+  rows.forEach(function(r) {
+    uniques.add(r[prop])
+  });
+  return Array.from(uniques)
+    .filter(function(value) { return value; })
```
leeoniya, line 15: what's the purpose of this filter?

brockfanning: @leeoniya That is to filter out null values. I wonder if there is a better way, hopefully less mysterious?

brockfanning: My this would be better:

```
var uniques = new Set();
rows.forEach(function(r) {
  if (r[prop]) {
    uniques.add(r[prop])
  }  
});
return Array.from(uniques).sort();
```

leeoniya: oh, interesting. i didn't realize that the return value merely needs to *coerce* to `true`. i usually do an explit check to return a boolean. `.filter(v => v != null)` works for filtering just `null` and `undefined`. i guess that counts as less mysterious :D

https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Array/filter#Description

leeoniya: your `if` version works, too abd should compress better. again, i usually try to avoid coercion and do `if (r[prop] != null) {`.

you may want to bench thr `if` version against the `filter` version. if the vast majority of the values are non-null, running 92k `if`s may be slower than just plucking at the end. although i think the perf diff here will be negligible for datasets below millions of datapoints.

brockfanning: Makes sense, thanks. I just committed that change. The array being filtered here is not the full list of rows - it's only the unique values from that Set - so I think going with the most readable code is fine here. Thanks for pointing that out!

jwestw (approved):
Processing of the chart now seems to be instantaneous. I have only checked 4.2.1. Not aware of other large datasets.  
http://sdgdev-813006012.eu-west-1.elb.amazonaws.com/jw-test-large-datasets/4-2-1/

PR #873, opened 2020-09-06
aria-described-by for indicator disaggregation variable hint

### LINKED ISSUE #859: Link variable-hint element to dropdown button element
**Page 78 in DAC report**

**Issue from DAC report:**
The hint text describing the dependency is not programmatically associated with the field.

**Solution from DAC report:**
Use aria-describedby attributes to programmatically associate instructions with dropdown fields

**Example from DAC report:**

```
<div class="variable-selector disallowed child" data-field="Region"
  data-has-data="true">
    <button class="accessBtn" tabindex="0" aria-expanded="false" aria-describedby="variable-hint">
        <h5>Region<i class="fa fa-chevron-down"></i></h5>
    </button>
    <div class="bar">…</div>
    <div class="variable-options">…</div>
    <div class="variable-hint" id="variable-hint">Available when 'Country' is selected</div>
    <div class="no-data-hint">No options available because of filters above</div>
</div>
```

brockfanning:
Fixes #859

### DIFF
--- a/_includes/assets/js/indicatorView.js
+++ b/_includes/assets/js/indicatorView.js
@@ -133,7 +133,14 @@ var indicatorView = function (model, options) {
       $(element).find('.bar .selected').css('width', width);
 
       // is this an allowed field:
-      $(element)[_.contains(args.allowedFields, currentField) ? 'removeClass' : 'addClass']('disallowed');
+      if (_.contains(args.allowedFields, currentField)) {
+        $(element).removeClass('disallowed');
+        $(element).find('> button').removeAttr('aria-describedby');
+      }
+      else {
+        $(element).addClass('disallowed');
+        $(element).find('> button').attr('aria-describedby', 'variable-hint-' + currentField);
+      }
     });
   });
 
--- a/_includes/components/indicator/fields-template.html
+++ b/_includes/components/indicator/fields-template.html
@@ -3,7 +3,7 @@
     <fieldset>
       <legend class="sr-only">{{ page.t.indicator.sub_categories }} - <%=translations.t(fieldItem.field)%></legend>
       <div class="variable-selector<% if(allowedFields.indexOf(fieldItem.field) == -1) { %> disallowed child<% }%>" data-field="<%=fieldItem.field%>">
-        <button class='accessBtn' tabindex='0' aria-expanded='false' aria-described-by='variable-hint-<%=fieldItem.field%>'>
+        <button class='accessBtn' tabindex='0' aria-expanded='false' aria-describedby='variable-hint-<%=fieldItem.field%>'>
           <h5><%=translations.t(fieldItem.field)%><i class="fa fa-chevron-down"></i></h5>
         </button>
         <div class="bar">
--- a/_includes/components/indicator/fields-template.html
+++ b/_includes/components/indicator/fields-template.html
@@ -5,7 +5,7 @@
       <div class="variable-selector<% if(allowedFields.indexOf(fieldItem.field) == -1) { %> disallowed child<% }%>" data-field="<%=fieldItem.field%>">
         <button class='accessBtn' tabindex='0' aria-expanded='false'
           <% if(allowedFields.indexOf(fieldItem.field) == -1) { %>
-          aria-describedby='variable-hint-<%=fieldItem.field%>'
+          aria-describedby='variable-hint-<%=fieldItem.field.replace(/ /g, '-')%>'
           <% }%>
         >
           <h5><%=translations.t(fieldItem.field)%><i class="fa fa-chevron-down"></i></h5>
@@ -26,7 +26,7 @@
         </div>
 
         <% if(allowedFields.indexOf(fieldItem.field) == -1) { %>
-          <div class="variable-hint" id="variable-hint-<%=fieldItem.field%>">
+          <div class="variable-hint" id="variable-hint-<%=fieldItem.field.replace(/ /g, '-')%>">
             {%- capture var_hint_replacement -%}
               {% raw %}<%= translations.t(_.findWhere(edges, { To: fieldItem.field }).From) %>{% endraw %}
             {%- endcapture -%}
--- a/_includes/components/indicator/fields-template.html
+++ b/_includes/components/indicator/fields-template.html
@@ -3,7 +3,11 @@
     <fieldset>
       <legend class="sr-only">{{ page.t.indicator.sub_categories }} - <%=translations.t(fieldItem.field)%></legend>
       <div class="variable-selector<% if(allowedFields.indexOf(fieldItem.field) == -1) { %> disallowed child<% }%>" data-field="<%=fieldItem.field%>">
-        <button class='accessBtn' tabindex='0' aria-expanded='false' aria-describedby='variable-hint-<%=fieldItem.field%>'>
+        <button class='accessBtn' tabindex='0' aria-expanded='false'
+          <% if(allowedFields.indexOf(fieldItem.field) == -1) { %>
+          aria-describedby='variable-hint-<%=fieldItem.field%>'
+          <% }%>
+        >
           <h5><%=translations.t(fieldItem.field)%><i class="fa fa-chevron-down"></i></h5>
         </button>
         <div class="bar">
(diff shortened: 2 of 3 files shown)


jwestw:
Hi I have created a feature branch for this and I see in indicator 1.4.2 that aria-described-by text is present.
<button class="accessBtn" tabindex="0" aria-expanded="true" aria-described-by="variable-hint-Type of Documentation"> <h5>Type of Documentation<i class="fa fa-chevron-down"></i></h5> </button>
I am not sure if this means the fix is successful. I need some more explanation on this one. @brockfanning please ping me on Slack. Thanks

brockfanning:
@jwestw I think I was confused when I did this one, I'll revisit. For one, it should be "aria-describedby", and also I don't think various elements are pointing to each other as intended.

jwestw:
Okay. When you make the changes, please let me know explicitly what I am looking out for. Thanks

brockfanning:
@jwestw @LucyGwilliamAdmin This is ready for review again. With this PR, whenever the `Available whenever "[x]" is selected` message appears, the drop-down button for that field should have an `aria-describedby` attribute which points to the id of that message. For example, if the "Region" field is dependent on the "Country" field, then the markup for the Region drop-down button should look something like this:

```
<button class="accessBtn" tabindex="0" aria-expanded="true" aria-describedby="variable-hint-Region">
      <h5>Region<i class="fa fa-chevron-down"></i></h5>
</button>
```

And below it there would be a corresponding div:

```
<div class="variable-hint" id="variable-hint-Region">Available when 'Country' is selected</div>
```

Then if a "Country" is selected and the message goes away, the `aria-describedby` on the "Region" button should also go away.

jwestw:
Feature branch: http://sdgdev-813006012.eu-west-1.elb.amazonaws.com/jw-test-variable-hint-described-by

jwestw:
I notice that when using NVDA screen reader on[ 2.1.1 ](http://sdgdev-813006012.eu-west-1.elb.amazonaws.com/jw-test-tab-keyboard-navigation/2-1-1/) it reads the titles in each of the category selector buttons. When tabbing down to 'sex' which is only available when age is selected, it reads "Sex" and states "collapsed" or when you click it "expanded" even though it has not actually gone into an expanded state because it's unavailable until 'Age' is selected. 
I got it to read 'Available when 'Age' is selected' by clicking the down arrow key ↓  on the while the category button is selected. This is possibly intended behaviour. 
I think "Expanded" and "Collapsed" should not be read out if the button is not available though.

LucyGwilliamAdmin:
> I notice that when using NVDA screen reader on[ 2.1.1 ](http://sdgdev-813006012.eu-west-1.elb.amazonaws.com/jw-test-tab-keyboard-navigation/2-1-1/) it reads the titles in each of the category selector buttons. When tabbing down to 'sex' which is only available when age is selected, it reads "Sex" and states "collapsed" or when you click it "expanded" even though it has not actually gone into an expanded state because it's unavailable until 'Age' is selected.
> I got it to read 'Available when 'Age' is selected' by clicking the down arrow key ↓ while the category button is selected. This is possibly intended behaviour, and if so, could be approved.
> However, I think "Expanded" and "Collapsed" should not be read out if the button is not available.

@jwestw  I agree that "Available when XX is selected" isn't read out when focussing on sub-category but think it actually should be, without having to use down arrow. I'm not sure about "Expanded" and "Collapsed" - may be something to check with DAC

jwestw:
> @jwestw I agree that "Available when XX is selected" isn't read out when focussing on sub-category but think it actually should be, without having to use down arrow. I'm not sure about "Expanded" and "Collapsed" - may be something to check with DAC

Reading out "Expanded" and "Collapsed" would be fine if the box/button was available and something was happening. But they are read out when the button is not able to expand and collapse.

LucyGwilliamAdmin:
> > @jwestw I agree that "Available when XX is selected" isn't read out when focussing on sub-category but think it actually should be, without having to use down arrow. I'm not sure about "Expanded" and "Collapsed" - may be something to check with DAC
> 
> Reading out "Expanded" and "Collapsed" would be fine if the box/button was available and something was happening. But they are read out when the button is not able to expand and collapse.

@jwestw Yeah, I get that - but they can technically be expanded and collapsed and vision users can see what items are available when XX has been selected.

> I got it to read 'Available when 'Age' is selected' by clicking the down arrow key ↓ while the category button is selected. This is possibly intended behaviour, and if so, could be approved.

@jwestw how did you do this? Just tried to recreate but when I press down arrow when item is focused on it just scrolls down the page

jwestw:
> @jwestw how did you do this? Just tried to recreate but when I press down arrow when item is focused on it just scrolls down the page

I just recreated it again by tabbing through the to the unavailable item, when the focus is on the button it reads the title, but when I pressed down, it reads "Available when [X] is selected". So I simply pressed the down arrow button to make this occur.

LucyGwilliamAdmin:
> > @jwestw how did you do this? Just tried to recreate but when I press down arrow when item is focused on it just scrolls down the page
> 
> I just recreated it again by tabbing through the to the unavailable item, when the focus is on the button it reads the title, but when I pressed down, it reads "Available when [X] is selected". So I simply pressed the down arrow button to make this occur.

Ah right, that doesn't work for me for some reason

LucyGwilliamAdmin:
@brockfanning can you recreate what James has done to get "Available when XX is selected" read out?

And is when it's being read out (when dropdown is focused on OR when down arrow is pressed after dropdown is focussed on) something we need to check?

LucyGwilliamAdmin:
@jwestw @brockfanning I've just realised, I thin James is getting that read out because it's just reading out the next thing in sequence and not the aria-describedby attribute. So maybe aria-described attribute is not working?

LucyGwilliamAdmin:
@brockfanning I'm not sure if it's the naming of the variable-hint id that's causing a problem? I just changed id="variable-hint-a" and then aria-describedby="variable-hint-a" in the inspect panel, and the screen reader read it out

brockfanning:
@jwestw @LucyGwilliamAdmin Would it be possible to get a step-by-step to recreate the problem? I just tried the following with NVDA:

1. Go to http://sdgdev-813006012.eu-west-1.elb.amazonaws.com/jw-test-variable-hint-described-by/2-1-1/
2. Tab to the "Age" drop-down and uncheck the selected box
3. Tab to the "Sex" drop-down"

At this point it reads out "available when age is selected" (after a lot of other stuff).

LucyGwilliamAdmin:
@brockfanning no probs - do the following with NVDA:

1. Go to http://sdgdev-813006012.eu-west-1.elb.amazonaws.com/jw-test-variable-hint-described-by/5-2-1/
2.  Tab to Abuse sub-category

At this point it doesn't read out "available when type of partner abuse is selected".

However, it also works for me on 2.1.1

Makes me think even more, that it could be something to do with the naming. In 5.2.1 id has space (`aria-describedby="variable-hint-Abuse sub-category"`), in 2.1.1 it doesn't (`aria-describedby="variable-hint-Sex"`)

brockfanning:
Ah! Yes you're right, that would explain it. I'll need to "slugify" the name of the disaggregation for that id.

brockfanning:
@LucyGwilliamAdmin @jwestw A side-note, but maybe something we could fix here: When the button has focus, the new yellow/black focus treatment slightly overlaps the "Available when" message. It's minor so I'm fine with ignoring it. But if we wanted to fix it I think we would have to add some more margin above the "Available when" message.

LucyGwilliamAdmin:
@brockfanning checked screen reader issue - sounds great now! I think it might be worth fixing overhang if pretty straightforward to do so.

brockfanning:
@LucyGwilliamAdmin I just pushed a commit to help with that.

LucyGwilliamAdmin:
@brockfanning that looks good - no longer overlapping

I'm wondering if we need to do something similar here (part of black underline is hidden by disaggregation percentage bar):
image

@phillipgardiner do you have any thoughts on this?

LucyGwilliamAdmin:
Although that could be quite unrelated?

phillipgardiner:
@brockfanning @LucyGwilliamAdmin Obviously ideally it would not be obscured, but we may just put up with it for MVP. Also I have added a note to the tab card #892 but ideally for drop downs only the outline should be highlighted not the whole control, see example: https://design-system.service.gov.uk/components/select/

brockfanning:
@LucyGwilliamAdmin @phillipgardiner Since it's an easy tweak, I just made a commit to add more margin about that percentage bar. But let me know if you'd prefer we do that separately, and I could remove it.

LucyGwilliamAdmin:
@brockfanning @phillipgardiner just refreshed feature branch. Looks good to me and I'm happy for tweak to stay here

LucyGwilliamAdmin (approved):
Looks/sound great!

PR #877, opened 2020-09-07
Remove alert role from disclaimer

### LINKED ISSUE #860: Remove "alert" role from phase banner
The phase banner does not need to be implemented as an alert; this does not conform with GDS Design System guidelines and is causing an accessibility issue with search results: Instead of focus going to "## results found", it goes to the top of the page

Remove role="alert" from here:
https://github.com/open-sdg/open-sdg/blob/master/_includes/components/dev-disclaimer.html

brockfanning:
Fixes #860

### DIFF
--- a/_includes/components/dev-disclaimer.html
+++ b/_includes/components/dev-disclaimer.html
@@ -1,5 +1,5 @@
 <div class="container">
-  <div class="disclaimer-alert" role="alert">
+  <div class="disclaimer-alert">
     <strong class="phase-tag">{{ site.disclaimer.phase | default: page.t.header.alpha | t }}</strong>
     {% assign disclaimer_message_default = page.t.header.disclaimer | replace_first: '%email_address', site.email_contacts.suggestions %}
     {{ site.disclaimer.message | default: disclaimer_message_default | t }}


jwestw:
Okay, so I have finally worked out what a "Phase Banner" is. 

Feature branch set up here http://sdgdev-813006012.eu-west-1.elb.amazonaws.com/jw-test-remove-alert-role

I am inspecting the element that is the phase banner and I see no alert-role as such, but I do see 
```<div class="disclaimer-alert">``` 
But ```role="alert">``` isn't there. 

However, I cannot find the file "_includes/components/dev-disclaimer.html" in my feature branch. Not sure why that is.

brockfanning:
@jwestw The issue that this fixes, #860, was only to remove `role="alert"`, so that would be the thing to check/review on the feature branch.

It's normal for particular files to not be in a feature branch. When your feature branch is built, it takes all missing files from this PR. Ideally *all* files will be missing, so that it takes 100% of the files from this PR. In those rare cases where your feature branch actually does have one of the files affected by the PR, that means you are "overriding" that file. In those cases, you would need to manually figure out what changed in the PR and make the same change in your overridden version of the file. Hope that makes sense - it will probably become more clear with practice.

LucyGwilliamAdmin (approved):
Looks good

Issue #854, opened 2020-08-28
Use text symbol instead of image for sort arrows in table


LucyGwilliamAdmin:
**Page 43 in DAC report**

**Issue from DAC report:**
Important information including symbols indicating functionality are implemented as background images within the CSS which disappear in Windows high contrast mode. The full screen and search icons do not invert colours in high contrast mode.

Note: this is not using Open SDG high-contrast mode but Windows high contrast mode, so to recreate, turn on High contrast mode in Windows settings and open page in Internet Explorer

**Solution from DAC report:**
Please use a text symbol inline rather than a background image within the CSS. This means that the colours can be set within the HTML to ensure that content does not disappear when alternative styles are applied.

![image](https://user-images.githubusercontent.com/43404358/91593225-ad787280-e957-11ea-855d-e97499bad95e.png)

brockfanning:
This one might merit discussion with DAC. This is a very narrow audience: Windows users, using Windows high-contrast, in Internet Explorer. IE will be officially abandoned by Microsoft in less than a year. Do we want to spend the effort on such a narrow case?

Also we should consider Windows high-contrast mode in general? We have devoted considerable time to our own custom high-contrast mode. Is there any way we can focus on that development, rather than trying to support OS-specific high-contrast modes?

jwestw:
Hi Brock - absolutely agree with you here. IE will soon be dropped. Edge will take over, which itself is based on Chromium base, so we shouldn't cater for IE. 

With regards to Windows high contrast mode I have no idea how widely used it is, so can't really comment here.  

Side note: When I used high contrast mode in Windows, Chrome offers the High Contrast Extension (by: chrome-a11y-extensions).

LucyGwilliamAdmin:
Fix for this has been merged so I'll close it

PR #880, opened 2020-09-08
Accessibility fixes for sorting tables

### LINKED ISSUE #869: Use native HTML component for the table sort button
**Page 100 in DAC report**

**Issue from DAC report:**
Column headings within the tables are focusable and act as sort ascending/descending buttons upon activation. However, they have no semantic role to describe to expose this behaviour to assistive technology in order for screen reader users and voice activation users to engage with this functionality. Setting a table heading as focusable meant that screen reader users could not use the standard keyboard shortcuts to navigate the table.
These shortcuts ensure that the column heading information is relayed as users navigate the table data to ensure the information is understood.

**Solution from DAC report:**
Please use native HTML components wherever possible, for example, the button element.

**Example from DAC report:**

```
<div style="display: none;">
    <span id="ColumnSortInfo">Click to sort by this column</span>
</div>
…
<th scope="col" class="sorting_asc" rowspan="1" colspan="1" style="white- 
  space: nowrap;">
    <button aria-sort="ascending" aria-controls="DataTables_Table_0"
       aria-describedby="ColumnSortInfo">
        <span>Year</span>
    </button>
    <span class="sort" aria-hidden="true"></span>
</th>
```

brockfanning:
Fixes #869 and #854 

I'm not 100% clear on the benefits here, but here are the changes in this PR:
1. The up/down arrows are now done with Fontawesome icons instead of with background images. This is intended to fix the DAC complaint in #854.
2. Instead of tabbing to the `<th>` elements for to a sort, the users now tab to a `<span role="button">` element inside the `<th>`. The button is `aria-describedby` a hidden div. (Which will eventually have to be translated.) This is intended to fix the DAC complaint in #869.

### DIFF
--- a/_includes/assets/js/indicatorView.js
+++ b/_includes/assets/js/indicatorView.js
@@ -724,9 +724,9 @@ var indicatorView = function (model, options) {
       var table_head = '<thead><tr>';
 
       var getHeading = function(heading, index) {
-        var span = '<span class="sort" />';
-        var span_heading = '<span tabindex="0" role="button" aria-describedby="column-sort-info">' + translations.t(heading) + '</span>';
-        return (!index || heading.toLowerCase() == 'units') ? span_heading + span : span + span_heading;
+        var arrows = '<span class="sort"><i class="fa fa-sort-down"></i><i class="fa fa-sort-up"></i></span>';
+        var button = '<span tabindex="0" role="button" aria-describedby="column-sort-info">' + translations.t(heading) + '</span>';
+        return (!index || heading.toLowerCase() == 'units') ? button + arrows : arrows + button;
       };
 
       table.headings.forEach(function (heading, index) {
--- a/_includes/assets/js/indicatorView.js
+++ b/_includes/assets/js/indicatorView.js
@@ -725,7 +725,7 @@ var indicatorView = function (model, options) {
 
       var getHeading = function(heading, index) {
         var span = '<span class="sort" />';
-        var span_heading = '<span>' + translations.t(heading) + '</span>';
+        var span_heading = '<span tabindex="0" role="button" aria-describedby="column-sort-info">' + translations.t(heading) + '</span>';
         return (!index || heading.toLowerCase() == 'units') ? span_heading + span : span + span_heading;
       };
 
@@ -757,6 +757,8 @@ var indicatorView = function (model, options) {
       initialiseDataTable(el);
 
       $(el).removeClass('table-has-no-data');
+
+      $(el).find('th').removeAttr('tabindex');
     } else {
       $(el).append($('<h3 />').text(translations.indicator.data_not_available));
       $(el).addClass('table-has-no-data');
--- a/_includes/components/indicator/data-panes.html
+++ b/_includes/components/indicator/data-panes.html
@@ -3,6 +3,9 @@
     {% include components/charts/chart.html graph_type=page.indicator.graph_type %}
   </div>
   <div role="tabpanel" class="tab-pane" id="tableview">
+    <div style="display: none;">
+      <span id="column-sort-info">Click to sort by this column</span>
+    </div>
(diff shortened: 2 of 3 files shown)


brockfanning:
@jwestw Sorry this is a vague one, and I'm not really clear on the benefits myself. One concrete thing you could test though is using Internet Explorer along with Windows high-contrast mode. (ie, not the Open SDG high-contrast mode, but instead the Windows high-contrast mode). Previously, in IE + Windows high-contrast caused the table header sort arrows to be hidden. Now the sort headers should be visible.

jwestw:
Looks good for fixing #869 
image

jwestw:
#854 doesn't seem to be fixed for internet explorer. In normal/low contrast mode the arrow is not appearing now in Internet Explorer in both low and high contrast. Fixed in Chrome though. 

IE/Low contrast mode: 

image

Windows/IE high contrast:

image


Windows High Contrast + Chrome + Chrome high contrast extension:

image

brockfanning:
@jwestw Just to make sure it's not another problem, in IE do the arrows appear in normal contrast mode? If not, it could be a separate issue where Fontawesome icons are incompatible with certain IE security settings.

jwestw:
@brock The first image is from IE in Normal Contrast mode, so no the arrows are not showing up.

jwestw:
MS Edge in high contrast mode:

image


MS Edge in Normal Contrast mode:

image

jwestw:
**Regarding security settings in IE**

Unfortunately I can't test this as I cannot change security settings on my work machine and I do not have IE installed on any other device.

LucyGwilliamAdmin:
@brockfanning I've just revisited the report for these 2 issues.

The benefit for:

1. using text symbol instead of image is "that the colours can be set within the HTML to ensure that content does not disappear when alternative styles are applied." So I think this comes back to users being able to customise styles as they need.

2. having header as a `button` means that controls can be used to navigate the table. I've just done some testing with screen reader:

    a. On feature branch, user can sort by column when column header is focussed by pressing space bar. This previously didn't work.

    b. Also on feature branch, user can now navigate through table using ctrl+alt+arrowkey. This isn't possible on UK live site. 

    As you change row, the year of the new row is read out and as you change column, the name of the new column is read out (both along with value). However reading the screen reader user comments in the report ("I would expect the headings to be read along with the information so it’s clear to what information refers and to save confusion."), I'm not sure if they would expect column header to be read out every time, and maybe same for row.

    For example, for [1.2.1 on feature branch](http://sdgdev-813006012.eu-west-1.elb.amazonaws.com/jw-test-accessibility-sorting-tables/1-2-1/), when moving from 1 to 2, screen reader says "Male column 4 18.6". When moving from 2 to 3, it says "2006 row 3 18)

    image

    So in general terms, when moving between columns, screen reader reads out "[DISAGG-NAME] column [COL-NUMBER] [VALUE]". When moving between rows, screen reader reads out "[YEAR] row [ROW-NUMBER] [VALUE]".

brockfanning:
@LucyGwilliamAdmin Do you think it's OK to merge or are there still changes that need to be made?

LucyGwilliamAdmin:
@brockfanning I'm just looking at this: https://www.w3.org/WAI/tutorials/tables/#why-is-this-important (the first bullet point), where it says "Screen readers speak one cell at a time and reference the associated header cells, so the reader doesn’t lose context". I'm wondering if should always say "[DISAGG-NAME] [YEAR] [VALUE]" whenever navigating to a new cell?

I'm trying to find an example online but no luck so far

LucyGwilliamAdmin:
@brockfanning actually, just tried screen reader on first table here: https://www.w3.org/WAI/tutorials/tables/two-headers/ and screen reader reads out the same as what feature branch is reading out.

LucyGwilliamAdmin (approved):

PR #882, opened 2020-09-09
Add aria-label to contrast toggle

brockfanning:


LucyGwilliamAdmin (approved):
Feature branch here: http://sdgdev-813006012.eu-west-1.elb.amazonaws.com/hc-aria-label

Screen reader reads: "Disable high contrast link disable high contrast" for default contrast button and "Enable high contrast link enable high contrast" for high contrast button.

This is also what current longer buttons read out on UK live site so think this is correct.

jwestw:
My NVDA screen reader simply says 'A' and 'A' when hovering over the two high/low contrast buttons.

LucyGwilliamAdmin:
@jwestw what about when you focus on them by tabbing through?

PR #883, opened 2020-09-10
Use text color for disclaimer text

brockfanning:


LucyGwilliamAdmin (approved):
Looks good!

http://sdgdev-813006012.eu-west-1.elb.amazonaws.com/disclaimer-text/

PR #884, opened 2020-09-11
Custom javascript for better tab accessibility

### LINKED ISSUE #840: Make sure tab panels are included in page sequence

brockfanning:
Fixes #840 

@jwestw @LucyGwilliamAdmin This is ready for review. Because this is a moderately significant bit of javascript and does change the tab behavior, I've got this as an optional feature, `accessible_tabs`. Eventually I would foresee this is the default behavior, but for now I don't want to cause problems for the countries during upgrades.

So, when testing this in a feature branch, you would need to add `accessible_tabs: true` to the _config.yml file.

### DIFF
--- a/_includes/assets/js/indicatorView.js
+++ b/_includes/assets/js/indicatorView.js
@@ -769,7 +769,12 @@ var indicatorView = function (model, options) {
 
       $(el).removeClass('table-has-no-data');
 
-      $(el).find('th').removeAttr('tabindex');
+      $(el).find('th')
+        .removeAttr('tabindex')
+        .click(function() {
+          var sortDirection = $(this).attr('aria-sort');
+          $(this).find('span[role="button"]').attr('aria-sort', sortDirection);
+        });
     } else {
       $(el).append($('<h3 />').text(translations.indicator.data_not_available));
       $(el).addClass('table-has-no-data');
--- a/_includes/assets/js/tabs.js
+++ b/_includes/assets/js/tabs.js
@@ -1,38 +1,38 @@
 {% if site.accessible_tabs %}
 $(document).ready(function() {
     $('.nav-tabs').each(function() {
-        var $tabsList = $(this);
-        var $tabs = $tabsList.find('li > a');
-        var $panes = $tabsList.parent().find('.tab-pane');
+        var tabsList = $(this);
+        var tabs = tabsList.find('li > a');
+        var panes = tabsList.parent().find('.tab-pane');
 
-        $panes.attr({
+        panes.attr({
             'class': 'tabPanel',
             'role': 'tabpanel',
             'aria-hidden': 'true',
         }).hide();
 
-        $tabsList.attr({
+        tabsList.attr({
             'role': 'tablist',
         });
 
-        $tabs.each(function(idx) {
-            var $tab = $(this);
-            var tabId = 'tab-' + $tab.attr('href').slice(1);
-            var $pane = $tabsList.parent().find($tab.attr('href'));
+        tabs.each(function(idx) {
+            var tab = $(this);
+            var tabId = 'tab-' + tab.attr('href').slice(1);
+            var pane = tabsList.parent().find(tab.attr('href'));
 
-            $tab.attr({
+            tab.attr({
                 'id': tabId,
                 'role': 'tab',
                 'aria-selected': 'false',
                 'tabindex': '-1',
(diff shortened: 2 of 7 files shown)


jwestw:
Feature branch: http://sdgdev-813006012.eu-west-1.elb.amazonaws.com/jw-test-tab-keyboard-navigation

LucyGwilliamAdmin:
@brockfanning So now when tabbing through, focus goes from tab name to tab content unless a diff tab is selected using left and right arrow keys, which is great:
http://sdgdev-813006012.eu-west-1.elb.amazonaws.com/tab-nav/1-1-1/

However, there is something in the report (page 56) which I'm unsure about:

"Tab-panels are not focusable and **screen reader users are not able to navigate from the tabs using cursor keys**, so in order to read the content within the tab it is necessary to tab to the next focusable element on the page, and then use the cursor keys to navigate back up the page."

I'm not sure if this means that when the tab name is focussed on, the user should be able to use down arrow to start reading through tab content?

This is not the case on the feature branch or on the [w3 tabs example](https://www.w3.org/TR/wai-aria-practices/examples/tabs/tabs-2/tabs.html) though.

LucyGwilliamAdmin (approved):
I've compared this with a couple of accessibility tab examples, and tab and screen reader act in same way 👍 

http://sdgdev-813006012.eu-west-1.elb.amazonaws.com/tab-nav/1-1-1/

jwestw:
A couple of things I noticed on this:
- when tables are sortable, there is not indication that it has been completed. Clicking the button to sort them doesn't make screen reader say "sorted" nor is a click sound played. 
- although the column headers are read out through the screen reader I didn't find a way to have the data read out. Is that intended behaviour?

brockfanning:
@jwestw I just added a commit that might help with the sorting issue. Can you give it a try?

About moving around in the table, I believe the standard keyboard shortcuts are CTRL+ALT+arrow keys. This is how non-sighted users would get the data read to them, as I understand it.

jwestw:
Feature branch refreshed: http://sdgdev-813006012.eu-west-1.elb.amazonaws.com/tab-nav

jwestw:
The first issue is improved. Reader now says that you can sort the column. However I didn't find a way to make it say anything when it is clicked to make data sorted ascending or descending. 

I couldn't get CTR + ALT + arrows to work, but SHIFT + Arrows did work. As the data is selected it reads it out, but the highlight doesn't move. 
The way data items in the table are selected seems odd to me but this could just be the behaviour of NVDA, I am not sure. Behaviour is:
- SHIFT + Down selects data first going down a row, then cycling through that row's columns. 
- SHIFT + left/right selects individual numerals/characters in the data so when it hits '8.15' it reads out '8 selected' then 'dot selected'.
- I don't see the selection move (but maybe this isn't important) 

Again, all of this behaviour might be NVDA specific and be out of our control

LucyGwilliamAdmin:
@brockfanning @jwestw table sorting is fixed here, isn't it:

https://github.com/open-sdg/open-sdg/pull/880

brockfanning:
Yes, technically table-sorting is unrelated to this ticket, but I thought I would throw that tweak in. All I'm doing in that last commit is copying the `aria-sort` attribute (which is generated by jquery.datatables) to the custom `<span role="button">` that we added in #880. My hope was that the `aria-sort` property would resolve the problem. If we need to do much else than it should probably be in a separate ticket.

@jwestw You're right that we probably don't want to get too involved with the workings of a particular screenreader. And I would guess it's fine that the cells don't get highlighted, since it is likely only non-sighted people using that. (Also cells aren't "focusable" anyway, so couldn't normally be highlighted.)

LucyGwilliamAdmin:
@brockfanning @jwestw is there anything else to fix on this one?

brockfanning:
Not that I know of.

PR #886, opened 2020-09-13
Search button outside of label

### LINKED ISSUE #847: Move search button outside of search input label
**Page 114 in DAC report**

**Issue from DAC report:**
The search button is within the label for the search component, which is confusing, as the button becomes a part of the input label.

**Solution from DAC report:**
Please implement the search button outside of the search input label as it already has an aria-label attribute, this is sufficient.

**Example from DAC report:**

```
<form action="/search">
    <input name="q" id="indicator_search" title="Search">
    <label for="indicator_search">
        <span>Search</span>
    </label>
    <button aria-label="Search" id="search-btn" type="submit"><i class="fa 
      fasearch" aria-hidden="false"></i></button>
</form>
```

instead of

```
<form action="/search">
    <input name="q" id="indicator_search" title="Search">
    <label for="indicator_search">
        <button aria-label="Search" id="search-btn" type="submit"><i class="fa 
           fasearch" aria-hidden="false"></i></button><span>Search:</span>
    </label>
</form>
```

brockfanning:
Fixes #847 

As part of this some changes to the CSS were necessary, and so I took the opportunity to remove a bit of CSS that did not seem needed. So manual cross-browser testing will be necessary to ensure that the search bar looks the same as before.

This implements the markup changes suggested in #847. There should be no visual or functional change.

### DIFF
--- a/_includes/search.html
+++ b/_includes/search.html
@@ -1,8 +1,7 @@
 <div id="search" class="menu-target">
   <form action="{{ page.baseurl }}search">
     <input type="text" name="q" id="indicator_search" title="{{ page.t.search.search }}" />
-    <label for="indicator_search">
-      <button aria-label="{{ page.t.search.search }}" id="search-btn" type="submit"><i class="fa fa-search" aria-hidden="false"></i></button><span>{{ page.t.search.search }}:</span>
-    </label>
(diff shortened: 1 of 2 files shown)


LucyGwilliamAdmin (approved):
Checked on Chrome, Edge, Firefox, IE, mobile Chrome, mobile Huawei browser and looks good on all!

PR #887, opened 2020-09-13
Add role='status' to search results info

### LINKED ISSUE #861: Give search results element "status" role
Links to #860 - when search results load, focus isn't taken to "## results found". To fix this give element role="status" here:
https://github.com/open-sdg/open-sdg/blob/master/_layouts/search.html#L6

DAC report (page 116) mainly mentions that part, but may also apply to when there are no results e.g. code starting here:
https://github.com/open-sdg/open-sdg/blob/master/_layouts/search.html#L22

brockfanning:
Fixes #861 

This add `role="status"` to the paragraph with info about the number of results found on the search page.

### DIFF
--- a/_layouts/search.html
+++ b/_layouts/search.html
@@ -3,7 +3,7 @@
 
 <script type="text/template" class="results-template">
   <% if(resultsCount) { %>
-    <p><%= resultsCount %> results found</p>
+    <p role="status"><%= resultsCount %> results found</p>
     <% _.each(searchResults, function(result){ %>
       <div class="row search-result">
         <div class="col col-xs-12">
@@ -19,7 +19,7 @@
     <% }); %>
 
   <% } else { %>
-    <p>
+    <p role="status">
       {{ page.t.search.no_results }}
       <% if(didYouMean) { %>
       - {{ page.t.search.did_you_mean }}:


LucyGwilliamAdmin:
@brockfanning have looked at this one can see you've implemented the solution DAC suggested.

However, the focus is still taken to the top of the page and I need to navigate down to "## results found". Looking at DAC's comments on page 116 of the report, I think they expect for focus to automatically go to "## results found" when a term is searched for.

jwestw:
Feature branch: http://sdgdev-813006012.eu-west-1.elb.amazonaws.com/search-role-status

brockfanning:
@LucyGwilliamAdmin I'm curious what the report says there (I can't find it at the moment). As-is, that element is a `<p>` so I'm not sure that it can even be focused on. I know that very first search result could be focused on, if that's wanted. But personally it seems a little unusual for this page to start with the focus down past the navigation, while all other pages start with the focus at the top.

LucyGwilliamAdmin:
@brockfanning have just pinned it to Slack channel - page 116

LucyGwilliamAdmin:
I've also just found this:

https://www.w3.org/WAI/WCAG21/working-examples/aria-role-status-searchresults/

In the example, when pressing enter to search screen reader reads "0 results returned" whereas for this feature branch it reads out the landing info. I think it's because the page reloads, rather than the search results just appearing. Might be worth discussing this with DAC?

LucyGwilliamAdmin:
Sorry for all the comments...

Also just found this (giving element tabindex=0):
https://developer.mozilla.org/en-US/docs/Web/HTML/Global_attributes/tabindex

Not sure if it will work but might be worth a try?

LucyGwilliamAdmin:
Sorry for all the comments...

I'm not sure what happens when the search button is pressed - does it go to a link? Could we include an anchor in the link? 

Maybe something like this: go to link and click focus on "Search button" and press enter. Screen reader then reads "1 result found"
http://sdgdev-813006012.eu-west-1.elb.amazonaws.com/search-role-status-overwrite/search/?q=smoking#search-results-count

image

Element is focused on highlighted temporarily, but isn't keyboard focusable

LucyGwilliamAdmin:
@brockfanning Reply from DAC:
> I just double checked this one and if the page refreshes then it doesn’t need to be a status message, the refresh behaviour will alert users to the fact that the page has updated – plus you have updated the page heading to alert users to the changes on the page. We would recommend that the page title match the page heading, but otherwise the implementation works fine!

So I think we can close this one and the related issue?

PR #890, opened 2020-09-13
Use aside instead of div for news sidebar

### LINKED ISSUE #851: Mark up and move Categories side bar (news page)
**Page 161 in DAC report**

**Issue from DAC report:**
The ‘Posts by Category’ side navigation region (see updates image) is not marked-up. As it is placed after the rest of the main page content within the HTML, screen reader users will encounter it after the rest of the page is relayed. The context of its visual positioning alongside the main content and at the top of the page is not presented to screen reader users.

**Solution from DAC report:**
Please ensure the sidebar is marked-up within an ‘aside’ region. In addition, it should be included above the main page content to ensure that users encounter it in page sequence prior to the main content to which it relates.

**Example from DAC report:**

```
<aside id="sidebar">
    <h4>Posts by Category</h4>
    <ul> …</ul>
</aside>
<main id="main-content" class="container newslist">…</main>
```

brockfanning:
Fixes #851

### DIFF
--- a/_includes/components/post-categories.html
+++ b/_includes/components/post-categories.html
@@ -1,4 +1,4 @@
-<div id="sidebar">
+<aside id="sidebar">
   <h4>{{ page.t.post.posts_by_category }}</h4>
   <ul>
     {% for category in site.categories %}
@@ -8,4 +8,4 @@
       </li>
     {% endfor %}
   </ul>
-</div>
+</aside>


LucyGwilliamAdmin (approved):
`aside` is now used instead of `div` and screen reader now reads "Complimentary landmark list with 5 items" instead of "List with 5 items"

Issue #858, opened 2020-09-01
Implement long description link to table


LucyGwilliamAdmin:
**Page 32 in DAC report**

**Issue from DAC report:**
The table provides a suitable alternative to the information displayed within the chart. However, the table is not presented to users as a direct alternative, as there is no clear label to allow users to identify its purpose as an alternative. Users may not understand the association between the alternative forms of information.

**Solution from DAC report:**
Consider implementing a long description link to navigating users to the tabular form of the data using the ‘longdesc’ attribute. Please consider the implementation below in order to ensure that the context of the actionable components associated with the chart are contextualised as part of the chart group.

**Example from DAC report:**

```
<figure role="group">
    <figcaption class="visuallyhidden"> Percentage of the
 population living in households at risk of poverty </figcaption>
    <div id="chart-canvas">
        <h3 id="chart-heading" class="chart-title">Percentage of the
 population living in households at risk of poverty</h3>
        <div id="chart" class="plot-container" style="height: 420px;">
            <div class="chartjs-size-monitor">
                <div class="chartjs-size-monitor-expand"><div class=""></div></div>
                <div class="chartjs-size-monitor-shrink"><div class=""></div></div>
            </div>
            <canvas role="img" aria-labelledby="chart-heading plotLegend"
 style="display: block; height: 400px; width: 795px;" width="1192"
 height="600" class="chartjs-render-monitor" longdesc="link to table"
 ></canvas>
    …
    </div>
    <div id="plotLegend">
        <ul id="legend">
            <li data-datasetindex="0">
                <button aria-label="Remove graph of United Kingodom from
 chart" class="swatch" style="background-color: #777777">
 United Kingdom</button>
            </li>
        </ul>
    </div>
    <div id="selectionChartFooter" class="table-footer-text">
            <dl>
                <dt>Source: </dt><dd>Eurostat</dd>
                <dt>Geographical Area: </dt><dd>United Kingdom</dd>
                <dt>Unit of Measurement: </dt><dd>Percentage (%)</dd>
            </dl>
        </div>
    </div>
</figure>
```

Additionally, consider the following implementation so that screen reader users know where to access the alternative (for chart and map tab)

```
<div style="display: none;">
    <span id="details">For tabular data alternative see Table tab</span>
</div>
<ul class="nav nav-tabs data-view" role="tablist">
    <li role="presentation" class="nav-item active">
        <a class="nav-link" data-toggle="tab" href="#chartview"
[...]

PR #894, opened 2020-09-16
Chart/table accessibility

### LINKED ISSUE #853: Make users aware that table tab is a direct alternative to chart
**Page 32 in DAC report**

**Issue from DAC report:**
The table provides a suitable alternative to the information displayed within the chart. However, the table is not presented to users as a direct alternative, as there is no clear label to allow users to identify its purpose as an alternative. Users may not understand the association between the alternative forms of information.

**Solution from DAC report:**
Consider implementing a long description link to navigating users to the tabular form of the data using the ‘longdesc’ attribute. Please consider the implementation below in order to ensure that the context of the actionable components associated with the chart are contextualised as part of the chart group.

**Example from DAC report:**

```
<figure role="group">
    <figcaption class="visuallyhidden"> Percentage of the
 population living in households at risk of poverty </figcaption>
    <div id="chart-canvas">
        <h3 id="chart-heading" class="chart-title">Percentage of the
 population living in households at risk of poverty</h3>
        <div id="chart" class="plot-container" style="height: 420px;">
            <div class="chartjs-size-monitor">
                <div class="chartjs-size-monitor-expand"><div class=""></div></div>
                <div class="chartjs-size-monitor-shrink"><div class=""></div></div>
            </div>
            <canvas role="img" aria-labelledby="chart-heading plotLegend"
 style="display: block; height: 400px; width: 795px;" width="1192"
 height="600" class="chartjs-render-monitor" longdesc="link to table"
 ></canvas>
    …
    </div>
    <div id="plotLegend">
        <ul id="legend">
            <li data-datasetindex="0">
                <button aria-label="Remove graph of United Kingodom from
 chart" class="swatch" style="background-color: #777777">
 United Kingdom</button>
            </li>
        </ul>
    </div>
    <div id="selectionChartFooter" class="table-footer-text">
            <dl>
                <dt>Source: </dt><dd>Eurostat</dd>
                <dt>Geographical Area: </dt><dd>United Kingdom</dd>
                <dt>Unit of Measurement: </dt><dd>Percentage (%)</dd>
            </dl>
        </div>
    </div>
</figure>
```

Additionally, consider the following implementation so that screen reader users know
where to access the alternative.
Example:

```
<div style="display: none;">
    <span id="details">For tabular data alternative see Table tab</span>
</div>
<ul class="nav nav-tabs data-view" role="tablist">
    <li role="presentation" class="nav-item active">
        <a class="nav-link" data-toggle="tab" href="#chartview"
 aria-controls="chartview" role="tab" data-on="click" data-event-
 category="Tab change" data-event-action="Change data view" data-event-
 label="Change to Chart tab" aria-describedby="details">Chart</a>
    </li>
    …
</ul>
```

brockfanning:
Fixes #853 and #858

This adds a hidden element that indicates that the table tab is an alternative display to the chart. The screenreader should announce this when a keyboard user tabs to the "Chart" or "Map" tabs, and also when the user tabs to the Chart or Map itself.

Notes:
* I couldn't get this to work without making the chart itself focusable. So, tabbing into the Chart tab will focus the whole chart, at which point the screenreader will announce the tabular alternative again.
* Although the DAC recommendation said to use `longdesc` on the canvas, I couldn't get this to work at all. The screenreader seemed to completely ignore it. I used `aria-describedby` instead, which worked for me. We could ask DAC if `longdesc` should also be used.
* This includes an unrelated fix to the keyboard navigation of the tabs (see the `e.preventDefault` in the PR changes). Without this fix, the screenreader was announcing "T" when changing tabs via the keyboard arrows (at least in NVDA this was happening).


jwestw:
I've just made a feature branch: http://sdgdev-813006012.eu-west-1.elb.amazonaws.com/jw-test-chart-table-accessibility

LucyGwilliamAdmin:
@brockfanning @jwestw looking at this now.

The tab functionality seems to have reverted - I can now tab through each of the tabs instead of tabbing down to chart and then using left and right arrow to choose. Is this expected?

brockfanning:
Yes, I think I'll need to revisit this.

brockfanning:
@jwestw @LucyGwilliamAdmin It seems like the feature branch is not reflecting the latest code from my branch - but when I look at your branch code everything seems to be correct. The javascript file should have a whole section about tabs, but it doesn't seem to be making it through to the feature branch ([see here](http://sdgdev-813006012.eu-west-1.elb.amazonaws.com/jw-test-chart-table-accessibility/assets/js/sdg.js?v=20200916181735)).

brockfanning:
@jwestw @LucyGwilliamAdmin Ah! I remember now - there is a config setting necessary: `accessible_tabs`. That needs to be set to true.

Do you think that should be the default? I made it opt-in because I didn't want to introduce a change in behavior without the countries specifically turning it on. But given our confusion I think I could be persuaded that it's OK to make it the default.

LucyGwilliamAdmin:
@brockfanning ah yes - I remember that field from another PR. I'll add that in now. Think for now it shouldn't be default but then maybe for next major version, we could make it the default?

LucyGwilliamAdmin:
@brockfanning have just had a look of this.

Looks good in Chrome!

1. Go to http://sdgdev-813006012.eu-west-1.elb.amazonaws.com/jw-test-chart-table-accessibility/3-a-1/
2. Tab to Chart tab - screen reader reads "tab control chart tab selected for tabular data alternative see Table tab"
3. Use right arrow to try to go to Table tab - Focus moves to Table tab and screen reader reads "table tab selected"
4. Use right arrow to try to go to Map tab - Focus moves to Map tab and screen reader reads "map tab selected for tabular data alternative see Table tab"

However, in Edge I do the following:

1. Go to http://sdgdev-813006012.eu-west-1.elb.amazonaws.com/jw-test-chart-table-accessibility/3-a-1/
2. Tab to Chart tab - screen reader reads "chart tab selected for tabular data alternative see Table tab"
3. Use right arrow to try to go to Table tab - screen reader reads "h"
4. Use right arrow to try to go to Table tab - screen reader reads "a"
5. Use right arrow to try to go to Table tab - screen reader reads "r"
6. Use right arrow to try to go to Table tab - screen reader reads "t"
7. Use right arrow to try to go to Table tab - screen reader reads "line feed"
8. Use right arrow to try to go to Table tab - screen reader reads "out of tab tab" and focus moves to Table tab

And same happens when trying to move from Table to Map tab

In IE:
1. Go to http://sdgdev-813006012.eu-west-1.elb.amazonaws.com/jw-test-chart-table-accessibility/3-a-1/
2. Tab to Chart tab - screen reader reads "chart table map tab control chart tab http://sdgdev-813006012.eu-west-1.elb.amazonaws.com/jw-test-chart-table-accessibility/3-a-1/ chart view selected for tabular data alternative see Table tab"
3. Use right arrow to try to go to Table tab - focus moved to Table tab and screen reader reads "table tab http://sdgdev-813006012.eu-west-1.elb.amazonaws.com/jw-test-chart-table-accessibility/3-a-1/ table view selected table"
4. Use right arrow to try to go to Map tab - focus moved to Map tab and screen reader reads "map tab http://sdgdev-813006012.eu-west-1.elb.amazonaws.com/jw-test-chart-table-accessibility/3-a-1/ map view selected for tabular data alternative see Table tab"

LucyGwilliamAdmin:
@brockfanning I said I'd test this again on personal machine as have newer version of NVDA on there - have just done so.

Works on Chrome, Edge and Firefox fine, but does the same thing with IE

LucyGwilliamAdmin:
What does the `href` attribute on the tabs do? e.g. here:
image

I was trying to see why the URL is being read out on IE and may be something to do with that

brockfanning:
@LucyGwilliamAdmin I believe the href is used by the Bootstrap javascript to connect the tabs with the panes. But I tried removing the href in our tabs.js file, and it doesn't seem to cause any problems. This should fix that issue in IE, according to here: https://stackoverflow.com/questions/35824664/how-to-prevent-href-attribute-getting-read-by-nvda-screenreader

LucyGwilliamAdmin:
Have just retested on Chrome, IE, Firefox and Edge. All of the tab stuff sounds great!

When focused on chart, "alternative data.." is read out on all but IE - are you experiencing this too?

brockfanning:
@LucyGwilliamAdmin Yes, same thing here. I think this explains it: https://stackoverflow.com/questions/42302985/nvda-screen-reader-is-not-reading-aria-labelledby-and-aria-describedby-in-intern

LucyGwilliamAdmin:
Ah right, that's unfortunate - do you think there will be a workaround at all?

brockfanning:
According to that thread, adding `tabindex="-1"` to the referenced element should work, but I just tried it locally and it made no difference.

brockfanning:
From what I've been reading, NVDA users tend to use Firefox. And I would suspect that Windows Narrator uses tend to use Edge. Given that IE11 is so close to being abandoned by Microsoft I lean towards ignoring any IE11 issues unless they're show-stoppers.

_includes/components/charts/canvas.html
```diff
@@ -0,0 +1,6 @@
+<canvas 
+    role="img"
+    aria-labelledby="chart-heading legend-help"
```
LucyGwilliamAdmin, line 3: @brockfanning what is `chart-heading` here?

brockfanning: @LucyGwilliamAdmin Good catch - I had copied that from the DAC example but then forgot to add the id. I've just pushed that commit. Hopefully this won't cause too much repetition in what the screenreader reads.

LucyGwilliamAdmin:
> From what I've been reading, NVDA users tend to use Firefox. And I would suspect that Windows Narrator uses tend to use Edge. Given that IE11 is so close to being abandoned by Microsoft I lean towards ignoring any IE11 issues unless they're show-stoppers.

I just revisited the report, DAC test IE with Jaws so downloaded that and reads out like the other browsers so should be all good from DAC test p.o.v!

_includes/components/charts/canvas.html
LucyGwilliamAdmin: It is a little strange because its read out again directly after but not sure if we can really do anything about it?

LucyGwilliamAdmin: So the one read out is coming from chart-heading, what is the other one coming from?

brockfanning: I believe it's the `<figcaption>` that DAC recommended. Do you think we should remove that or wait and see what DAC says about it?

LucyGwilliamAdmin: I see - I've just found [this](https://www.scottohara.me/blog/2019/01/21/how-do-you-figure.html#:~:text=A%20figcaption%20provides%20a%20caption,used%20on%20the%20figure%20instead.) which may give us grounds for removing the figcaption

LucyGwilliamAdmin: I was also just starting to think maybe we could combine the h4 with the `figcaption` e.g.
image

but I think I remember them saying this will make the hierarchy incorrect

brockfanning: Ok. Personally I would be fine with removing the figure/figcaption.

LucyGwilliamAdmin: Although the heading of the table is a caption... I think maybe one to discuss with DAC but what are your thoughts?

One thing to bare in mind with this, the captured area for the image download will need to be changed to include figcaption

LucyGwilliamAdmin: Sorry didn't refresh so didn't see your comment before I posted. I'm happy either way

brockfanning: It seems so much better for without the figcaption, I removed it. It will be easy to add back if necessary.

_includes/components/charts/chart.html
```diff
@@ -7,16 +7,13 @@
     {% assign graph_template = 'components/charts/' | append: include.graph_type | append: '.html' %}
     {% assign graph_title = page.indicator.graph_titles.first.title | default: page.indicator.graph_title %}
     <div id="selectionsChart">
-      <figure role="group">
```
LucyGwilliamAdmin: @brockfanning sorry, I'm wondering we should just remove `figcaption` but still keep it wrapped in `figure`?

brockfanning: Gotcha, just added it back.

LucyGwilliamAdmin (approved):
Checked on all browsers and everything sounds/looks good!

PR #902, opened 2020-09-18
Alert role and screenreader-only warning label for too-many-datasets

### LINKED ISSUE #856: Relay max chart lines warning to screen readers and make it clearer it's appeared
Part of this relates to https://github.com/open-sdg/open-sdg/issues/163

**Page 70 in DAC report**

**Issue from DAC report:**
There is nothing to indicate to screen reader users that a warning has appeared on the page upon their selection. When users do encounter this by navigating through the page, there is nothing to indicate that it is a warning message as this is indicated visually via a yellow background and lightning bold icon with no text alternative. Low vision users also found it difficult to locate this message.

**Solution from DAC report:**
Adding a role of alert to the containing <div> will ensure this message is relayed to screen reader users when it appears onscreen as it will recognise that the CSS style ‘display:none;’ has changed to ‘display:block;’.

Additionally, please consider implementing the message in line with the sub-category components so that low vision users’ navigation with high page magnification may perceive that it has appeared.

**Example from DAC report:**

```
<div id="dataset-size-warning" style="" role="alert">
    <i class="fa fa-bolt"><span class="visuallyhidden">Warning:</span></i>
      There are too many dataset sub-categories specified to show on the chart.
      Not all will be shown until you remove some selections.
</div>
```

brockfanning:
Fixes #856 (partially)

### DIFF
--- a/_includes/components/charts/chart.html
+++ b/_includes/components/charts/chart.html
@@ -1,7 +1,8 @@
 {% if include.graph_type %}
 
-    <div id="dataset-size-warning" style="display:none">
-      <i class="fa fa-bolt"></i> {{ page.t.indicator.dataset_size_warning }}
+    <div id="dataset-size-warning" style="display:none" role="alert">
+      <i class="fa fa-bolt"><span class="sr-only">Warning:</span></i>
+      {{ page.t.indicator.dataset_size_warning }}
     </div>
 
     {% assign graph_template = 'components/charts/' | append: include.graph_type | append: '.html' %}


jwestw:
Feature branch: http://sdgdev-813006012.eu-west-1.elb.amazonaws.com/jw-test-too-many-datasets-revamp/

Testing now

jwestw:
This seems to work beautifully. It gives a clear warning that too many datasets have been selected and that they won't all be shown on the chart. It couldn't be clearer. (I think I actually prefer it to reading it!)
It didn't give the same warning when I was selecting as many data categories as possible while tabbed on the table tab, which is the expected behaviour.

jwestw (approved):

PR #904, opened 2020-09-18
Accessible charts

### LINKED ISSUE #843: Make chart tooltips accessible to all users
**Page 90 in DAC report**

**Issue from DAC report:**
Certain data points are interactive, producing a tooltip containing text details when hovered with the mouse. This functionality is not keyboard accessible. On mobile, when the content is zoomed, the tooltip is not fully displayed on the screen, but when a user scrolls it disappears.

**Solution from DAC report:**
Please ensure that any information and functionality which is available to sighted users navigating with a mouse is accessible to all users. A control could be provided which allows users to single out particular points on the graph and relay the corresponding percentage information. 

**Example from DAC report:**
No example provided

brockfanning:
Fixes #843 

This doesn't use any standard HTML approaches, because the charts are Canvas elements. But this does include the following:
1. Keyboard navigation of the tooltips using the left/right arrows
2. Screenreader announcement of the tooltips during keyboard navigation
3. Give the chart a transparent background when focused, instead of the yellow background.
4. Slight re-arrangement of the elements referenced in the canvas aria-describedby, so that they're read in this order:
    * Chart title
    * How to use tooltip keyboard navigation
    * Info about tabular alternative
    * How to use legend buttons

Like the accessible tabs this is implemented as an opt-in feature. To use it, implementations will need `accessible_charts` set to true.

Tech note: This uses a lot of Chart.js internals, which is not ideal. We could watch https://github.com/chartjs/Chart.js/issues/2850 to see if any future helper functions might be usable here.


LucyGwilliamAdmin:
Feature branch: http://sdgdev-813006012.eu-west-1.elb.amazonaws.com/accessible-charts

LucyGwilliamAdmin:
@brockfanning have looked on Chrome, IE, Edge and Firefox and functionality looks really great!

Thought I'd take a look on mobile just to see how it looks/works and I'm not really sure if it's possible to nav through the data points

LucyGwilliamAdmin (approved):
We agreed that since this is a bonus and is working on desktop version of site as expected that we will merge in despite not working on mobile

PR #901, opened 2020-09-16
Search results link fixes

brockfanning:
Fixes #845 

* Removes the category/path from the search result link
* Adds the indicator/goal number in front of the title in the link
* Fixes the problem with the focus treatment of search results

This relies on a PR for jekyll-open-sdg-plugins: https://github.com/open-sdg/jekyll-open-sdg-plugins/pull/51

So testing this would require this in the `Gemfile`:

`gem 'jekyll-open-sdg-plugins', git: 'https://github.com/brockfanning/jekyll-open-sdg-plugins.git', branch: 'numbers-in-search-results'`

Here is a screenshot:

numbers-in-search-results

PR #905, opened 2020-09-19
Full indicator name in the page title and h1

brockfanning:
This is an experimental PR intended as a companion to #901 to make the indicator page title and H1 closer to what displays in the search results, in that PR.

Visually nothing should change, but semantically the indicator name is now included inside the H1.

Also the page title (ie, the contents of the browser tabs) will now be the word "Indicator" followed by the number, then a dash and the name. Example:

`Indicator 1.1.1 - Proportion of population below the international poverty line, by sex, age, employment status and geographical location (urban/rural)`.

We could simplify that a bit more by removing "Indicator" and/or the dash. Example:

`1.1.1 Proportion of population below the international poverty line, by sex, age, employment status and geographical location (urban/rural)`

This would make it exactly like the search results in #901.

An important note is that whatever we choose for the page title will be what appears in search engines (Google, DuckDuckGo, etc.). Right now, the page title is (for example) `Indicator 1.1.1 - Goal 1` and so this is what appears in search engines.

### DIFF
--- a/_includes/components/indicator/header.html
+++ b/_includes/components/indicator/header.html
@@ -8,8 +8,10 @@
       </div>
       <div class="col-xs-8 col-md-9 col-lg-10 indicator-details">
         <div>
-          <h1>{{ page.t.general.indicator }} {{ page.indicator.number }}</h1>
-          <p class="lead-copy">{{ page.indicator.name }}</p>
+          <h1>
+            {{ page.t.general.indicator }} {{ page.indicator.number }}
+            <span class="lead-copy">{{ page.indicator.name }}</span>
+          </h1>
         </div>
       </div>
     </div>
--- a/_includes/head.html
+++ b/_includes/head.html
@@ -17,7 +17,7 @@
         ================================================== -->
 
         {% if page.indicator %}
-        <title>{{ page.t.general.indicator | escape }} {{ page.indicator.number }} - {{ page.t.general.goal | escape }} {{ page.goal.number }}</title>
+        <title>{{ page.t.general.indicator | escape }} {{ page.indicator.number }} - {{ page.indicator.name | escape }}</title>
         {% elsif page.goal %}
         <title>{{ page.t.general.goal | escape }} {{ page.goal.number }} - {{ page.goal.short | escape }}</title>
         {% elsif page.title %}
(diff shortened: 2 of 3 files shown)


brockfanning:
@LucyGwilliamAdmin @phillipgardiner @jwestw @AnnCorp We recently discussed trying this out just to see how hard it was, and if it was something we'd like to do. So I put up this PR to help the discussion.

jwestw:
I can see the entire title is wrapped in the H1 tag now, although the text appears no bigger (as intended). 

I have used Microsoft Narrator (which is good by the way) and NVDA to test this but I don't see any difference in behaviour. I can't seem to get either to focus on the H1 titles. Will continue to the experiment with that. 

Is this meant to only be an SEO change, or should it indeed be affecting the screen reader performance and experience?

LucyGwilliamAdmin:
@jwestw the headings aren't focusable - navigate to the thing that is focusable before (the goal logo in this case I think) and then use the down arrow key for the screen reader to read the heading

brockfanning:
Based on latest feedback, we're going to go ahead with this PR, so this is ready for review.

LucyGwilliamAdmin:
Feature branch from James; http://sdgdev-813006012.eu-west-1.elb.amazonaws.com/jw-test-indicator-page-title-in-H1

LucyGwilliamAdmin:
@brockfanning just looked at this one - it works - but the heading is sort of split up. For example, on Chrome for 1.1.1 it reads "heading level 1 Proportion of population below the international poverty line, by sex, age," need to press down arrow again - "heading level 1 employment status and geographical location (urban/rural)"

Same for Edge and Firefox (but may split in slightly different place

Fine for IE with Jaws though

brockfanning:
@LucyGwilliamAdmin That sounds similar to something I ran into with NVDA also: it apparently has a maximum line length, but that is configurable. I went into the preferences for NVDA and under "Browse mode" was able to increase the number of characters. If that helps with this issue, I think that means it is screenreader-specific and not related to our code.

LucyGwilliamAdmin (approved):
@brockfanning ah yes - happy to approve then 👍 

@jwestw I wen ahead and reviewed this one as it was last accessibility PR

PR #906, opened 2020-09-21
Add site title to page title

brockfanning:
Demonstration of adding site title to page title.

### DIFF
--- a/_includes/head.html
+++ b/_includes/head.html
@@ -16,15 +16,7 @@
         <!-- Title and meta description
         ================================================== -->
 
-        {% if page.indicator %}
-        <title>{{ page.t.general.indicator | escape }} {{ page.indicator.number }} - {{ page.indicator.name | escape }}</title>
-        {% elsif page.goal %}
-        <title>{{ page.t.general.goal | escape }} {{ page.goal.number }} - {{ page.goal.short | escape }}</title>
-        {% elsif page.title %}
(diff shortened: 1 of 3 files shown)


LucyGwilliamAdmin:
Here's a feature branch showing how this would look on UK site:
http://sdgdev-813006012.eu-west-1.elb.amazonaws.com/dac2020/

@phillipgardiner I know you said you'll pick this up tomorrow - this is the PR about including site title in all page titles

phillipgardiner:
@brockfanning @LucyGwilliamAdmin My thoughts on this (and i'm not expert!). The H1 should be followed by the site name.
Like you say we are probbaly best to discuss at tech call as I need to make sure I have understood the issue / potential solution correctly. The only thing I would add is I believe as a user may search using an acronym we should potentially add '(SDG)' at the end of the site name.

phillipgardiner:
Also can we clarify on the call what the name of of site actually is?

brockfanning:
We'll also add the site title to the goal and indicator pages.

brockfanning:
@LucyGwilliamAdmin This is ready for review again. I thought it might make sense to split out the page title into its own include file, to make it easier for countries that might want to have more control.

LucyGwilliamAdmin (approved):
Looks good!

http://sdgdev-813006012.eu-west-1.elb.amazonaws.com/dac2020/

PR #907, opened 2020-09-22
Backwards-compatible array-style for search_index_boost

brockfanning:
This is one more tweak to support the site-config-builder functionality. It is intended to be fully backwards-compatible. So the old syntax will still work:

```
search_index_boost:
    title: 10
```

But the documented/recommended syntax will be produced by the site builder, and also works:

```
search_index_boost:
    - field: title
      boost: 10
```

### DIFF
--- a/_includes/assets/js/search.js
+++ b/_includes/assets/js/search.js
@@ -144,8 +144,19 @@ var indicatorSearch = function() {
   // Helper function to get a boost score, if any.
   function getSearchFieldOptions(field) {
     var opts = {}
-    if (opensdg.searchIndexBoost[field]) {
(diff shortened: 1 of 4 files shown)


brockfanning:
@LucyGwilliamAdmin @jwestw This is minor and shouldn't change anything, but it's needed to continue working towards that site config form functionality.

LucyGwilliamAdmin (approved):
Works both ways and returns same results:

http://sdgdev-813006012.eu-west-1.elb.amazonaws.com/search-index-boost/search/?q=poverty
http://sdgdev-813006012.eu-west-1.elb.amazonaws.com/search-index-boost-new/search/?q=poverty

PR #916, opened 2020-09-25
Only announce no data, added data, or removed data.

### LINKED ISSUE #910: Use shorter status message to relay changes to chart/table
@brockfanning Converting to issue so that we can discuss DAC's response

DAC's consultancy response:

> The reason it is recommended to place it closer to the filters is that low vision users who require a high magnification may not notice this message – two of our low vision users did not locate the message when it appeared. However, as they are still able to select the filters and the data still appears in the table, I would agree that the message only needs to be relayed when users actually go to view the table – in which case, they should locate the message, so I would say this is fine.
 
Side note relating to DAC_On_Input_01, I just noticed with JAWS and every time a new filter is selected, all of the selected sub-categories are relayed as a status message. I would strongly advise that only the current change should be relayed. i.e. when age group ’65 to 74’ is selected, instead of:

**Current Code Ref(s):** #indicator-data-view-status

```
<span role="status" class="sr-only" id="indicator-data-view-status">Chart and table shows datasets for United Kingdom and Female and Male and 15 and under and 16 to 24 and 25 to 49 and 65 to 74 and 15 and under, Female and 16 to 24, Female and 25 to 49, Female and 65 to 74, Female and 15 and under, Male.</span>
```

Please make sure that status message just relays something like:

```
<span role="status" class="sr-only" id="indicator-data-view-status">Chart and table updated to include breakdows for age group 65 to 74</span>
```

- otherwise it becomes confusing.

brockfanning:
Fixes #910.

### DIFF
--- a/_includes/assets/js/indicatorView.js
+++ b/_includes/assets/js/indicatorView.js
@@ -327,6 +327,7 @@ var indicatorView = function (model, options) {
   }
 
   this.updatePlot = function(chartInfo) {
+    this.updateIndicatorDataViewStatus(view_obj._chartInstance.data.datasets, chartInfo.datasets);
     view_obj._chartInstance.data.datasets = chartInfo.datasets;
 
     if(chartInfo.selectedUnit) {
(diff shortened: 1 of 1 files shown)


LucyGwilliamAdmin (approved):
Checked on Chrome, IE, Firefox and Edge - correct text read out for each scenario

Sounds good!

http://sdgdev-813006012.eu-west-1.elb.amazonaws.com/shorter-disagg-readouts/5-1-1/

PR #917, opened 2020-09-25
Allow disabling of news category links

### LINKED ISSUE #911: Remove the news category sidebar (Following up on #851)
@brockfanning converting to issue so we can discuss DAC's response:

> In general, I would not say that information was supplementary if the lack of that information does not stop the user achieving their main goal. We would expect all users to be able to access the same information and functionality.

> I would say that supplementary information is perhaps an additional visual cue for sighted users that does not provide any information or functionality but is just decorative.

> This navigation region isn’t decorative as it provides functionality. This is just a usability comment, so to achieve AA standard it does not need to be met. It is just in the report to make you aware that blind users will not be able to access that information as easily as a sighted user who can immediately see this information on the page.

brockfanning:
Fixes #911 

This allows category links to be disabled, with this site config:

```
news:
  category_links: false
```

### DIFF
--- a/_layouts/news.html
+++ b/_layouts/news.html
@@ -1,9 +1,14 @@
 ---
 layout: default
 ---
+{% if site.news.category_links == false %}
+  {% assign category_links = false %}
+{% else %}
+  {% assign category_links = true %}
+{% endif %}
 <div id="main-content" class="container newslist">
   <div class="row">
-    <div class="col-md-9 push-md-3">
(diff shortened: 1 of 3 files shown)


LucyGwilliamAdmin (approved):
Looks good to me!

http://sdgdev-813006012.eu-west-1.elb.amazonaws.com/keep-categories/updates/
http://sdgdev-813006012.eu-west-1.elb.amazonaws.com/rem-categories/updates/

PR #925, opened 2020-09-29
Make the active tab panel focusable

### LINKED ISSUE #924: Make tab-panels focusable
**URL:** http://sdgdev-813006012.eu-west-1.elb.amazonaws.com/dac2020/1-2-1/

**Current Code Ref(s):** #chartview

```
<div role="tabpanel" class="tabPanel" id="chartview" aria-hidden="false" aria-labelledby="tab-chartview" style="">[…]</div>
```

**Solution:**
Include the tabpanel in the page sequence when it is triggered so that it can be focused by screen reader users.

**Example:** 

```
<div role="tabpanel" class="tabPanel" id="chartview" aria-hidden="false" aria-labelledby="tab-chartview" style="" tabindex="0">[…]</div>
```

brockfanning:
Fixes #924

### DIFF
--- a/_includes/assets/js/tabs.js
+++ b/_includes/assets/js/tabs.js
@@ -41,9 +41,15 @@ $(document).ready(function() {
                         'tabindex': '-1',
                     });
                 
-                panes.filter(':visible').attr('aria-hidden', 'true').hide();
+                panes.filter(':visible').attr({
+                    'aria-hidden': 'true',
+                    'tabindex': '-1',
+                }).hide();
 
-                pane.attr('aria-hidden', 'false').show();
+                pane.attr({
+                    'aria-hidden': 'false',
+                    'tabindex': '0',
+                }).show();
 
                 tab.attr({
                     'aria-selected': 'true',


jwestw:
Feature branch:  http://sdgdev-813006012.eu-west-1.elb.amazonaws.com/jw-test-focusable-tab-panels

jwestw:
Currently trying to spot difference in behaviour between
https://sdgdata.gov.uk/1-2-1/
and 
http://sdgdev-813006012.eu-west-1.elb.amazonaws.com/jw-test-focusable-tab-panels/1-2-1/

brockfanning:
@jwestw My impression from the ticket's description is that you should now be able to use the TAB key to get to the contents of a tab -- aka a "tab pane" (for example the Sources on an indicator page) and then use the down arrow to read the pane's contents.

LucyGwilliamAdmin:
@brockfanning just checking does `accessible_tabs: true` need to be set for testing this?

brockfanning:
Ah, good point - yes it does.

LucyGwilliamAdmin:
@brockfanning thanks

@jwestw  Can you add `accessible_tabs` to the config file of the feature branch please?

jwestw:
This is working very nicely! 

image

Hopefully my (unedited) screenshot shows the highlighted tab-pane.

jwestw (approved):
Approving according to the previous comments.

LucyGwilliamAdmin:
@brockfanning @jwestw sent to DAC for sign-off and got the following reply:

> looks resolved but I wouldn’t do a yellow background as a focus, just a yellow border

PR #933, opened 2020-09-29
aria-describedby for no-data-hint

### LINKED ISSUE #929: Ensure "No options available..." message is programmatically associated with its field
**URL:** http://sdgdev-813006012.eu-west-1.elb.amazonaws.com/dac2020/3-b-1/

**Screen shot:**

The hint text ("No options available because of the filters above") describing the dependency is not programmatically associated with the field.

**Solution:**
Please use an aria-describedby attribute to programmatically associate instructions with the field.

brockfanning:
Fixes #929

### DIFF
--- a/_includes/assets/js/indicatorView.js
+++ b/_includes/assets/js/indicatorView.js
@@ -157,6 +157,15 @@ var indicatorView = function (model, options) {
       // Indicate whether the fieldGroup had any data.
       var fieldGroupElement = $(view_obj._rootElement).find('.variable-selector[data-field="' + fieldGroup.field + '"]');
       fieldGroupElement.attr('data-has-data', fieldGroup.hasData);
+      var fieldGroupButton = fieldGroupElement.find('> button'),
+          describedByCurrent = fieldGroupButton.attr('aria-describedby') || '',
+          noDataHintId = 'no-data-hint-' + fieldGroup.field.replace(/ /g, '-');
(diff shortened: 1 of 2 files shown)


jwestw:
Feature branch: http://sdgdev-813006012.eu-west-1.elb.amazonaws.com/jw-test-no-data-hint-aria

jwestw:
The screen reader reads "No options available because of the filters above" on both https://sdgdata.gov.uk/3-b-1/ and in my feature branch http://sdgdev-813006012.eu-west-1.elb.amazonaws.com/jw-test-no-data-hint-aria/3-b-1/ . 

I can see the difference in the code, but I am not sure what behaviour changes I should be observing here.

brockfanning:
@jwestw Can you describe _when_ the screenreader announces that? If the screenreader is announcing it when you tab to the button, I think that is the intended behavior.

LucyGwilliamAdmin:
@jwestw just looked at this, and for me it's not read out on live site but is on your feature branch

jwestw:
> @jwestw just looked at this, and for me it's not read out on live site but is on your feature branch

I think we had this difference before, didn't we? In setup, it has the desired behaviour in both branches. So as long as you are seeing it in the feature branch and it wasn't appearing before, then it is fixed.

jwestw (approved):
Approving on the basis of what Lucy has found.

PR #934, opened 2020-09-29
Use aria-hidden when hiding sidebar

### LINKED ISSUE #931: Make sure sidebar is hidden from assistive technology when map/embed is showing
**URL:** http://sdgdev-813006012.eu-west-1.elb.amazonaws.com/dac2020/3-a-1/

When the maps or dashboard tabs are opened, the ‘Sub-categories’ toolbar visually disappears, but are not hidden from assistive technology. Screen reading software still relays these.

**Solution:**
Please ensure content which should be hidden is done so in a way that is hidden from assistive technology. This can be achieved with the CSS style ‘display: none;’.

brockfanning:
Fixes #931 

This doesn't use `display: none` because I believe that would mess up our animation. Instead this uses `aria-hidden` which should give the same result.

### DIFF
--- a/_includes/assets/js/indicatorView.js
+++ b/_includes/assets/js/indicatorView.js
@@ -43,7 +43,7 @@ var indicatorView = function (model, options) {
           hideSidebar = $(this).data('no-disagg'),
           mobile = window.matchMedia("screen and (max-width: 990px)");
       if (hideSidebar) {
-        $sidebar.addClass('indicator-sidebar-hidden').attr('aria-hidden', 'true');
+        $sidebar.addClass('indicator-sidebar-hidden');
         $main.addClass('indicator-main-full');
         // On mobile, this can be confusing, so we need to scroll to the tabs.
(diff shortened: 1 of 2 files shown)


LucyGwilliamAdmin:
@brockfanning I've just tried this out but it still seems to be there in the background e.g. if you go [here](http://sdgdev-813006012.eu-west-1.elb.amazonaws.com/aria-hidden-sidebar/3-a-1/) and tab through, when you get to goal icon you will have to tab a few times to get to the next visible focusable item

brockfanning:
@LucyGwilliamAdmin Good catch. I guess we're going to have to give up on our animation here. I've just removed the animation and switched to `display:none`.

LucyGwilliamAdmin (approved):
Sad that we've lost the animation but it does work now:
http://sdgdev-813006012.eu-west-1.elb.amazonaws.com/aria-hidden-sidebar/3-a-1/

PR #942, opened 2020-10-01
Alternate header

brockfanning:
This is not ready for review yet, but a feature branch may be helpful for feedback on the visuals. The feature branch should have this config:
```
header:
    template: accessible
```


LucyGwilliamAdmin:
@brockfanning @phillipgardiner 

Feature branch: http://sdgdev-813006012.eu-west-1.elb.amazonaws.com/accessible-header/

phillipgardiner:
@LucyGwilliamAdmin @brockfanning looking good. Couple of comments:
1) The logo has shrunk in size can it be made bigger?
2) The spacing between the searh and menu has reduce, can it be re-installed.

see https://sdgdata.gov.uk/ for how it should look - -thanks

brockfanning:
@phillipgardiner Those two fixes should be fairly easy. But let's focus on whether or not to do the "hamburger menu", and if so, how that should be styled. If we decide not to do the hamburger menu (which would be totally fine) then I would probably continue these tweaks in a separate PR.

brockfanning:
@phillipgardiner Makes sense - it was only because of the issues with the Menu/Search tabs in another ticket, but I'm remembering now that I was going to investigate another potential fix for that specific issue (the tab order being wrong). We can pause/close this while I try out that option.

brockfanning:
I'll close this for now. Hopefully #952 resolves that tab order issue, and then I'll start a new PR for a non-hamburger alternate header.

Issue #937, opened 2020-09-30
Ensure all map components are labelled


LucyGwilliamAdmin:
**URL:** http://sdgdev-813006012.eu-west-1.elb.amazonaws.com/dac2020/3-a-1/

**Screen shot:**
![image](https://user-images.githubusercontent.com/43404358/94693977-42da9e00-032c-11eb-97c5-f86b2c99854f.png)

The slider is a custom component with no role or programmatically associated value. It is not keyboard focusable and it has no label to identify it.

Additionally, many of the links do not have a label to allow them to be accessed by voice activation software users. The purpose of these links is indicated visually by an icon with the text label available via a tooltip which appears on mouse hover.
Some of the links on the page have a role of button (zoom in, zoom out and home), the rest do not. This is inconsistent.

**Solution:**
Please ensure that all components are labelled.

**Example:**
`<a class="leaflet-control-fullscreen-button leaflet-bar-part" href="#" title="View Fullscreen" style="outline: none;" role="button" aria-label="View fullscreen"></a>`

The slider needs to be implemented as a focusable component with its functionality made accessible via the keyboard such as the cursor keys. Its behaviour should be described to screen reader users. It is advised to use native HTML components where possible to ensure that assistive technology can engage with it as expected such as the slider role, or the input type range component. For more information, please visit [Horizontal Slider Example | WAI-ARIA Authoring Practices 1.1](https://www.w3.org/TR/wai-aria-practices-1.1/examples/slider/slider-1.html).

brockfanning:
These are all third-party libraries, but I'll look into ways to fix the problems.

LucyGwilliamAdmin:
I checked this and looks good to me: http://sdgdev-813006012.eu-west-1.elb.amazonaws.com/map-slider-accessibility

However I've sent DAC feature branches for all the fixes so far - do you want me to wait for their feedback to approve?

brockfanning:
I'm ok either way - just to make clear though, #941 only fixes part of this issue. I still need to fix the search and fullscreen buttons - I think they'll each be in their own PRs.

LucyGwilliamAdmin:
Closing as fixed

Issue #936, opened 2020-09-30
Ensure map symbols show in high contrast mode


LucyGwilliamAdmin:
**URL:** http://sdgdev-813006012.eu-west-1.elb.amazonaws.com/dac2020/3-a-1/

**Screen shot:**
<img src="https://user-images.githubusercontent.com/43404358/94693420-97c9e480-032b-11eb-90ce-c6fa229e0002.png" width="50%">

<img src="https://user-images.githubusercontent.com/43404358/94693426-9a2c3e80-032b-11eb-81a8-584470c0016c.png" width="50%">

Important information including symbols indicating functionality are implemented as background images within the CSS which disappear in Windows high contrast mode. The full screen and search icons do not invert colours in high contrast mode.

**Solution:**
Please use a text symbol inline rather than a background image within the CSS. This means that the colours can be set within the HTML to ensure that content does not disappear when alternative styles are applied.

LucyGwilliamAdmin:
Closing as fixed

PR #944, opened 2020-10-02
Map fullscreen accessibility

brockfanning:
Fixes the parts of #936 and #937 that have to do with the fullscreen button.

### DIFF
--- a/_includes/assets/js/plugins/jquery.sdgMap.js
+++ b/_includes/assets/js/plugins/jquery.sdgMap.js
@@ -362,7 +362,7 @@
         plugin.map.addControl(L.Control.zoomHome());
 
         // Add full-screen functionality.
-        plugin.map.addControl(new L.Control.Fullscreen());
+        plugin.map.addControl(new L.Control.fullscreenAccessible());
 
         // Add the year slider.
         plugin.map.addControl(L.Control.yearSlider({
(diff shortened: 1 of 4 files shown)


jwestw:
Feature branch: http://sdgdev-813006012.eu-west-1.elb.amazonaws.com/jw-test-fix-hierarchical-field-sorting

jwestw:
The same problem is occuring on this feature branch: 

image
NOTE: this is **before** I am changing _config.yml 

However, it is not occuring on https://sdgdata.gov.uk/
image

jwestw:
All tested in Windows high contrast mode. 

Chrome does not respond to Windows high contrast mode, and has its own plug-in called "High Contrast". Using this plug-in it looks like this:

image

This is something like the expected behaviour but this does not seem to be as a result of a fix because on https://sdgdata.gov.uk/3-a-1/ we can also see: 

image

So it seems it was never a problem with this browser/plugin-combo

Both Edge and IE respond to Windows high contrast mode. 

Edge is the best:

image

All the icons in buttons except the search icon are appearing. The search icon does appear in grey if you hover over it however (you can't see my mouse pointer on this screenshot, but it was over the search button). 

IE is pretty bad. 

image

As you can see nothing is fixed and I think it is worse than the originally reported problem as the other buttons' characters have disappeared:

image

As @brockfanning  has pointed out before, Windows High Contrast + IE might be a bit of an edge case. Can we advise people to use Chrome?

LucyGwilliamAdmin:
@brockfanning @jwestw as this is working in Edge and IE with WHCM, I'm sending to DAC for sign-off

LucyGwilliamAdmin (approved):

PR #945, opened 2020-10-03
Map search accessibility

### LINKED ISSUE #938: Ensure map search works as expected
**URL:** http://sdgdev-813006012.eu-west-1.elb.amazonaws.com/dac2020/3-a-1/

**Screen shot:**
image

The search ‘button’ (implemented as a link) performs two functions. On the first instance it is to trigger the search input to appear. However, when the ‘search’ icon is focused and enter is pressed, the search input flashes onto the screen and then disappears again. There is nothing to relay its collapsed or expanded status to screen reader users.
Once the search input is expanded, if there is text in the search input, the search ‘button’ link then acts as a submit button allowing users to submit their search. If there is no text within the search field, it collapses the search field. This behaviour is not described to screen reader users.

**Solution:**
Please ensure that all functionality is able to be operated using the keyboard. To allow screen reader users to understand that a link triggers content to appear, an aria-expanded = “false” attribute can be used when the button is in a collapsed state, and a descriptive label should be provided such as aria-label="show search". When this button is configured to collapse the component, an aria-expanded = “true” attribute can be used with an aria-label="hide search". When this button is used as a search button, it should have no aria-expanded attribute and an aria-label="search".

brockfanning:
Fixes #938 
Fixes the parts of #937 related to the search button
Fixes the parts of #936 related to the search button

### DIFF
--- a/_includes/assets/js/plugins/jquery.sdgMap.js
+++ b/_includes/assets/js/plugins/jquery.sdgMap.js
@@ -380,7 +380,7 @@
         plugin.map.addControl(plugin.selectionLegend);
 
         // Add the search feature.
-        plugin.searchControl = new L.Control.Search({
+        plugin.searchControl = new L.Control.SearchAccessible({
           layer: plugin.getAllLayers(),
           propertyName: 'name',
           marker: false,
--- a/_includes/assets/js/plugins/jquery.sdgMap.js
+++ b/_includes/assets/js/plugins/jquery.sdgMap.js
@@ -381,6 +381,8 @@
 
         // Add the search feature.
         plugin.searchControl = new L.Control.SearchAccessible({
+          textPlaceholder: 'Search map',
+          autoCollapseTime: 5000,
           layer: plugin.getAllLayers(),
           propertyName: 'name',
           marker: false,
--- a/_includes/assets/js/plugins/jquery.sdgMap.js
+++ b/_includes/assets/js/plugins/jquery.sdgMap.js
@@ -382,7 +382,7 @@
         // Add the search feature.
         plugin.searchControl = new L.Control.SearchAccessible({
           textPlaceholder: 'Search map',
-          autoCollapseTime: 5000,
+          autoCollapseTime: 7000,
           layer: plugin.getAllLayers(),
           propertyName: 'name',
           marker: false,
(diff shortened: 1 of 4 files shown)


LucyGwilliamAdmin:
@brockfanning looks good to me - sending to DAC for sign off

http://sdgdev-813006012.eu-west-1.elb.amazonaws.com/map-search-accessibility/3-a-1/

LucyGwilliamAdmin (changes requested):
Had feedback from DAC on this:

> The search input field needs a label. The field seems to collapse on its self without being closed which may cause problems for users who require longer to use components. ‘Location not found’ also disappears very quickly, low vision users may struggle to read that message.

brockfanning:
@LucyGwilliamAdmin I'm interpreting the latest feedback are three separate comments:
1. The search input field needs a label.
2. The field seems to collapse on its self without being closed which may cause problems for users who require longer to use components.
3. ‘Location not found’ also disappears very quickly, low vision users may struggle to read that message.

I'm clear on the first and third, but would you be able to get clarification about the second one? Specifically, what is the user action that causes it to collapse on itself?

LucyGwilliamAdmin:
@brockfanning for 2. if a user clicks to open the search box and then clicks on the map or grey area surrounding the map, the search input closes.

1. and 3. look good! I'm wondering if the time for 3. should display just even a little longer?

brockfanning:
@LucyGwilliamAdmin I've raised the time for that alert from 5 seconds to 7 seconds. Easy to tweak it further if you'd like.

I've also removed the "auto-collapse" which means that after a search is submitted, it will stay expanded. I think this is a better experience for people who are trying to add selections to the map using the search

But this doesn't actually affect the second issue mentioned, which is that the search collapses when you click outside of it. Personally I don't understand how that is related to accessibility. Isn't that more of a design decision? I feel like that is pretty standard web application behavior.

LucyGwilliamAdmin:
Looks good to me - will send to DAC for sign-off and see what they say about input closing when map is clicked

http://sdgdev-813006012.eu-west-1.elb.amazonaws.com/map-search-accessibility/3-a-1/

LucyGwilliamAdmin (approved):
DAC aren't able to look at this ahead of retest but looks good to me so I'm approving. Another one I think we should merge so that we can make it live for retest @brockfanning

PR #947, opened 2020-10-03
Support for keyboard operation of the map selection legend

brockfanning:
This is not in response to a particular issue but I suspect it will come up. The selection legend on maps can't be accessed or operated with a keyboard. This should allow users to tab onto those items, and to remove them from the list by pressing enter.

### DIFF
--- a/_includes/assets/js/plugins/leaflet.selectionLegend.js
+++ b/_includes/assets/js/plugins/leaflet.selectionLegend.js
@@ -19,27 +19,71 @@
     },
 
     addSelection: function(selection) {
+      var selectionStatus = L.DomUtil.get('selection-status');
+      var parsedSelection = this.parseSelection(selection);
+      selectionStatus.innerHTML = 'Selection added: ' + parsedSelection.name + ': ' + parsedSelection.value + '.';
       this.selections.push(selection);
       this.update();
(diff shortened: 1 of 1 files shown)


LucyGwilliamAdmin:
Feature branch: http://sdgdev-813006012.eu-west-1.elb.amazonaws.com/map-selection-legend-accessibility/3-a-1/

LucyGwilliamAdmin:
@brockfanning I'm a little unsure about what this should be doing

brockfanning:
@LucyGwilliamAdmin This affects "selection legend" that appears in the upper right corner of maps, when you select regions. Those items in the upper right should now be accessible by keyoboard (tabbing onto them). Also if you are tabbed onto one of them and you press the enter key, it should behave like you clicked on it.

LucyGwilliamAdmin:
@brockfanning yes, looks like it's keyboard accessible now which is great 👍 

However the default map isn't showing any labels on the scale

brockfanning:
Let's create an issue and postpone this one.

LucyGwilliamAdmin:
@brockfanning did we say we would close this PR and reopen if/when we come back to the issue?

brockfanning:
Yes I think so, sounds good.

PR #948, opened 2020-10-04
Map caption/footer with revamp of chart/table footers

### LINKED ISSUE #297: Unit of measurement in map view
The unit is currently not displayed anywhere in the map view. Is it possible to show the translated content of the csv-"Unit"-field somewhere in or above the map?

### LINKED ISSUE #900: Add caption to map tab
Both charts and tables have a caption - maps don't. See comparison of all three in images in #899 

Add caption above map in map tab which describes what the map shows

So that if someone takes a screenshot of map and uses it out of context, it's clear what it is showing, and also for consistency with the chart and table tabs

None

brockfanning:
This PR includes a number of things:
1. Random bug-fix of the no-data message (it was broken when we merged the accessible charts PR)
2. Test for the no-data-message (to avoid it getting broken again)
3. Convert the chart and table footers to an HTML include file for easier customization, rather than obscure javascript code we're using now (I'm not sure why we had it as javascript before. If there is some reason that I'm forgetting, then I'll be happy to remove that change from this PR.)
4. Add the above-mentioned footer to the map tab.
5. Add the same chart title H4 to the map tab.
6. Remove a lot of unneeded code.

From the perspective of the end-user, the only thing that should be different is that the map tab now has a title and footer.

From the perspective of the developer, an improvement here is that there is now an `_includes/components/indicator/data-footer.html` file that can be overridden to easily customize the chart/table/map footers.

One particular thing to watch out for is to make sure that the chart and table footers look/behave the same as before. For example:
* They have the same styling (font size, indentation, spacing, etc)
* They still disappear when a chart/table has no data
* The chart footer still appears in the downloaded chart image

Fixes #297 
Fixes #900


LucyGwilliamAdmin:
Feature branch: http://sdgdev-813006012.eu-west-1.elb.amazonaws.com/map-caption/3-a-1/

Looks good to me - with DAC for sign off

LucyGwilliamAdmin (approved):
I'm happy with the functionality and have also received this from DAC:

> This is corrected and seems helpful

So approving

Issue #921, opened 2020-09-29
Remove grid role from tables


LucyGwilliamAdmin:
**URL:** http://sdgdev-813006012.eu-west-1.elb.amazonaws.com/dac2020/3-a-1/

**Screenshot:**
![image](https://user-images.githubusercontent.com/43404358/94578220-5ecc3a00-026f-11eb-898b-3e27711ed0d6.png)

A role of ‘grid’ has been used on a table which has not been implemented as a standard grid widget. Grids require focus to be managed, and once this role is assigned the usual table shortcuts will not work with assistive technology.

**Current Code Ref(s):** #DataTables_Table_0 
`<table class="table table-hover dataTable no-footer" id="DataTables_Table_0" role="grid" style="width: 100%;">[…]</table>`

**Solution:**
Please remove the role of grid from this table. For more information, please visit [Data Grid Examples | WAI-ARIA Authoring Practices 1.1](https://www.w3.org/TR/wai-aria-practices-1.1/examples/grid/dataGrids.html) and [Grid Design Pattern in WAI-ARIA Authoring Practices 1.1](https://www.w3.org/TR/wai-aria-practices-1.1/#grid).

LucyGwilliamAdmin:
@brockfanning I sent this to DAC for sign off and had the following reply:

>The table header aria-label and aria-controls can be removed from <th> now. Also, what seems to be preventing table shortcuts for screen readers in some browsers is setting the rowspan and colspan attributes. I think these attributes are interpreted as specifying the area introduced by the table header, so when the rowspan is set on a column header sometimes the tech seems to get confused. I would remove these as they may prevent screen reading software correctly relaying the table structure in certain browsers, and they don’t add anything for simple tables.
> 
> ```
> <th scope="col" class="sorting_asc" aria-controls="DataTables_Table_0" rowspan="1" colspan="1" aria-sort="ascending" aria-label="Year: activate to sort column descending" style="white-space: nowrap;"><span tabindex="0" role="button" aria-describedby="column-sort-info">Year</span><span class="sort"><i class="fa fa-sort-down"></i><i class="fa fa-sort-up"></i></span></th>
> ```

LucyGwilliamAdmin:
@brockfanning I've just reopened and moved to do as we had some comments back from DAC. Or would you prefer a new ticket?

LucyGwilliamAdmin:
Latest fix is with DAC for sign off

brockfanning:
[...]

PR #954, opened 2020-10-10
Remove rowspan and colspan from table headers

brockfanning:
Fixes the latest comment mentioned in #921


jwestw:
Feature branch for 954: http://sdgdev-813006012.eu-west-1.elb.amazonaws.com/jw-test-remove-datatable-rowspan-colspan

jwestw:
I see the changes in the code. Also, this works with NVDA in both IE and Chrome.

jwestw (approved):

PR #957, opened 2020-10-14
Keep swatch colors in Edge/IE + high-contrast mode

### LINKED ISSUE #955: Key and map colours don't match in Windows High Contrast Mode
image

brockfanning:
This fixes #955 only in the IE and Edge browsers. In other browsers (like Chrome and Firefox) this will have no effect, but if they ever decide to support the property `forced-color-adjust` property, it will resolve it for them too.

### DIFF
--- a/_sass/base/_mixins.scss
+++ b/_sass/base/_mixins.scss
@@ -96,3 +96,10 @@
   box-shadow: 0 -2px $focusOutlineColor,0 4px $focusColor !important;
   text-decoration: none !important;
 }
+
+@mixin preserveOriginalColors {
+  @media (-ms-high-contrast: active), (forced-colors: active) {
+    -ms-high-contrast-adjust: none;
+    forced-color-adjust: none;
+  }
+}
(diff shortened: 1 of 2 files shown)


LucyGwilliamAdmin:
Looks mostly good in both Edge and IE - the arrow is a diff shape still, which is making the arrow and swatch adjacent (and not meeting colour contrast rules):
image

brockfanning:
@LucyGwilliamAdmin I just committed a fix for that issue.

LucyGwilliamAdmin:
Looks good to me - will sent to DAC for sign off

http://sdgdev-813006012.eu-west-1.elb.amazonaws.com/map-legend-swatches-keep-colors/3-a-1/

LucyGwilliamAdmin (approved):
Approving as looks good to me. DAC don't have time to look over ahead of retest starting tomorrow so think we should merge, and then create new PR if needed - what are your thoughts @brockfanning?

Mainly thinking if we get all of the DAC stuff into master branch, we can make it live on UK site, so DAC can use that for retest

brockfanning:
Sounds good to me.

PR #965, opened 2020-10-16
Always focusable tab panes

### LINKED ISSUE #962: Tab panels are focusable but not on page load – tabs have to be switched before the tab panels are focusable
From DAC:

> “The tab panels are focusable but not on page load – tabs have to be switched before the tab panels are focusable.” – here is what I mean:

> No tabindex attribute on page load:
> image
> 
> Once I click a tab panel, tabindex attributes appear:
> image
> 
> The currently open tabpanel on pageload is not in focus sequence so users who are looking for the tab that is already open, and don’t need to select it won’t find the tabpanel in focus sequence. 
> 
> A good example is: https://www.w3.org/TR/wai-aria-practices/examples/tabs/tabs-1/tabs.html.  Their tabs always have a tabindex of 0, and this works fine because the ones that aren’t open are hidden with display:none; anyway so they wont be focusable unless they are open.

brockfanning:
Fixes #962

### DIFF
--- a/_includes/assets/js/tabs.js
+++ b/_includes/assets/js/tabs.js
@@ -8,6 +8,7 @@ $(document).ready(function() {
             'class': 'tabPanel',
             'role': 'tabpanel',
             'aria-hidden': 'true',
+            'tabindex': '0',
         }).hide();
 
         tabsList.attr({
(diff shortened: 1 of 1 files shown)


LucyGwilliamAdmin (approved):
Looks good to me:

http://sdgdev-813006012.eu-west-1.elb.amazonaws.com/tab-pane-tabindex-2/3-a-1/

PR #967, opened 2020-10-16
Remove aria-label from table headers

### LINKED ISSUE #960: Remove aria-label from table header
From DAC:

> Table works much better now but the aria-label="Year: activate to sort column descending" needs to be removed from the table header <th>.
> 
> This is because, as Screen reader users navigate the table, the are relayed the table header from each data cell. Without the label they would get ‘year 2011’ from the 2011 data cell, but because there is an aria-label, they instead get ‘Year: activate to sort column descending 2011’. This doesn’t make sense as the data cells are not sortable – only the column header cell. The markup is otherwise fine as it is.

brockfanning:
Fixes #960

### DIFF
--- a/_includes/assets/js/indicatorView.js
+++ b/_includes/assets/js/indicatorView.js
@@ -587,7 +587,7 @@ var indicatorView = function (model, options) {
 
     table.DataTable(datatables_options);
     table.removeAttr('role');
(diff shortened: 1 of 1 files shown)


LucyGwilliamAdmin (approved):
Looks good to me:

http://sdgdev-813006012.eu-west-1.elb.amazonaws.com/remove-aria-label-from-table-headers/1-2-1/

PR #968, opened 2020-10-19
Different canvas fallback for mobile/desktop.

brockfanning:
This also normalizes some text to make it easier to translate later.

### DIFF
--- a/_includes/assets/js/chartjs/accessibleCharts.js
+++ b/_includes/assets/js/chartjs/accessibleCharts.js
@@ -17,6 +17,12 @@ Chart.plugins.register({
                     break;
             }
         });
+        if (window.innerWidth <= 768) {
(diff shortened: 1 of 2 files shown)


LucyGwilliamAdmin (commented):
I have just tried this out but screen reader is still reading about using left and right arrows to browse data points

http://sdgdev-813006012.eu-west-1.elb.amazonaws.com/canvas-fallback-mobile-fix/1-2-1/

brockfanning:
@LucyGwilliamAdmin I'm guessing that it's because the "aria-describedby" was still there. I pushed up another change to hopefully fix that.

LucyGwilliamAdmin (approved):
Fab, works as I'd expect - will send to DAC once merged

PR #970, opened 2020-10-20
Disaggregation status

### LINKED ISSUE #794: Generate disaggregation status
Following on from the addition of the disaggregated report, we want to  automatically generate indicator 17.18.1:
Proportion of sustainable development indicators produced at the national level with full disaggregation when relevant to the target, in accordance with the Fundamental Principles of Official Statistics

brockfanning:
Fixes #794 

This depends on [another PR](https://github.com/open-sdg/sdg-build/pull/174), so requires pointing to a fork of sdg-build in the data repository's requirements.txt:

```
git+git://github.com/brockfanning/sdg-build-1@reporting-status-disaggregation
```

Also enable the behavior, in the site configuration, the `reporting_status` setting should include `disaggregation_tabs: true`. For example:

```
reporting_status:
  title: My reporting status title
  description: My reporting status description
  disaggregation_tabs: true
```

Finally, without any metadata changes all indicators will appear 100% disaggregated. To see more realistic percentages, indicator can have an `expected_disaggregations` metadata field, which lists the expected disaggregation columns. For example:

```
expected_disaggregations:
  - Sex
  - Age
```

In the example above, if the indicator only have a column for Age, then it would not be considered fully disaggregated.


LucyGwilliamAdmin:
Seems like I'm getting something strange - nothing is showing on the disaggregation reporting status tab but the first part of it is showing at the bottom of the reporting status tab:

http://sdgdev-813006012.eu-west-1.elb.amazonaws.com/disagg-rs1/reporting-status/

brockfanning:
Thanks @LucyGwilliamAdmin there were some markup issues. There may still be more, but I've pushed a fix. It also looks like the percentages do not add up to 100...

brockfanning:
@LucyGwilliamAdmin I also just pushed a fix up in the sdg-build branch which should hopefully help with that percentage problem.

brockfanning:
@LucyGwilliamAdmin Sorry for all the notes - you might also need to point to the latest jekyll-open-sdg-plugins. You can use this in the Gemfile, I believe:

```
gem "jekyll-open-sdg-plugins", "1.2.0.pre.beta3
```

LucyGwilliamAdmin:
@brockfanning disaggregation is showing now. It looks like percentages are adding up to 100 for all but goal 12.

1. I have a couple of questions though - what contributes to total number of indicators here (and also for the goals)?
image

    We have identified 50 indicators in total that request disaggregations in the indicator name so (as long as the metadata has been edited correctly) we should have 50 md files that contain the field `expected_disaggregations` so I'm guessing it's not that this is taken into account, as in this case the total is 240.

2. How would we go about adding another status e.g. partially reported (i.e. if 0 < number of expected disaggregations reported < number of expected disaggregations) - is this something that would need to be build into sdg-build? and would it be the three statuses by default with an option of just the two (although I personally think in all/most cases the 3 would be preferable) - apologies as I realise this makes it more complex

brockfanning:
@LucyGwilliamAdmin I suspect that the unexpected total is because it is only considering statistical indicators. Here is the chunk of code that does the looping and counting: https://github.com/brockfanning/sdg-build-1/blob/reporting-status-disaggregation/sdg/DisaggregationStatusService.py#L112-L145

I'm open to any other logic there. Probably the best would be if you could describe the ideal output. From your other question I'm wondering if maybe you'd prefer it to only consider indicators that have `expected_disaggregations` set?

Adding another status would need to be coded into sdg-build. This is different from normal reporting status, where it only needs to count totals according to arbitrary metadata field values. By contrast, here the different statuses will need different counting logic. It certainly could be done though, and I've coded this PR with the expectation that we'll be adding more. But we had discussed an initial attempt with the only the simplest distinction of fully-disaggregated or not fully-disaggregated.

I imagine that if we add more statuses, but want to make them optional, we'll need another data config option that controls that, like "disaggregation_status_types" or something.

LucyGwilliamAdmin:
@AnnCorp @phillipgardiner what are your thoughts on the total number of indicators - should this be just the indicators that actually require some sort of disaggregation e.g. in our case we have identified 50?

@brockfanning depending on Ann and Phill's response, if we did want to show only those that require a disaggregation, then I think yes, for us it would be better if only the indicators which have `expected_disaggregations` are included - I'm wondering what others' thoughts might be though.

Yes, that's true - I remember us discussing starting with an MVP. @AnnCorp in the UK, would we want to start with the three reporting statuses though (fully, partially, none), and if so, is that something we want to bring in now?

AnnCorp:
@LucyGwilliamAdmin I'll fire an email to Jo re this and loop you in - we can factor that in to how we take forward.

LucyGwilliamAdmin:
@brockfanning we've discussed this our side and wondering if the following is possible:

4 statuses:

- Fully disaggregated
- Partially disaggregated
- Not yet disaggregated
- Out of scope (total number of indicators - the number of indicators which have `expected_disaggregations`)

Having out of scope would allow the total to remain the same no matter what is counted as fully disaggregated

brockfanning:
@LucyGwilliamAdmin I've updated both PRs with that, so it's ready for testing.

LucyGwilliamAdmin:
@brockfanning thanks for letting me know - I've just rebuilt the feature branch but the calculations don't seem to be correct - We should have 194 indicators out of scope and then the other statuses would total 50

phillipgardiner:
@LucyGwilliamAdmin @brockfanning @AnnCorp  Sorry only just seen the link to the feature branch in the above conversation. I think some of the labels are potentially confusing (but I don't have alternatives - yet!). For instance on the feature branch we have 'Partially disaggregated' and 'Not fully disaggregated' these could be considered the same thing (e.g. there is a bit of it done)? Also 'out of scope' - what makes it out of scope? too difficult? not available? unbale to be disaggregated? are we planning on discussing this on tech call?

I also notice a few little things to do with capitalisation compared to the Overall reporting status page.

LucyGwilliamAdmin:
@phillipgardiner some good points

I think for the 3rd status, Jo suggested "Not yet disaggregated", which I think works quite well. - I'm not sure if the labels are customisable though?

Re. out of scope, the plan would be to have some text at the top describing what is currently being considered and what's in scope.

brockfanning:
Fix capitalization of "Reporting Status" ("Reporting status") and "Overall Reporting Status".

brockfanning:
Do a quick English-only sdg-translations PR.

brockfanning:
"Not disaggregated" instead of "Not yet disaggregated".

brockfanning:
@LucyGwilliamAdmin The PRs are updated and ready for a feature branch refresh.

LucyGwilliamAdmin:
Everything we discussed this morning looks good:
http://sdgdev-813006012.eu-west-1.elb.amazonaws.com/disagg-rs1/reporting-status/

LucyGwilliamAdmin:
Accidentally submitted comment before finishing typing...

All that we discussed this morning looks good. I have just checked the percentages again and they don't seem to be adding up

brockfanning:
@LucyGwilliamAdmin I guess that it's rounding up in some cases. I can look into that.

brockfanning:
@LucyGwilliamAdmin Yes we do have code that rounds the numbers to the nearest whole number. It actually works the same on the regular reporting status - but maybe it doesn't appear so obvious because it is not divided into four parts.

One option is to show 1 decimal place in the percentages. There would still be some rounding, and technically the numbers would still add up to a bit more than 100%, but most people would probably not notice.

The only gotcha with that approach is that it will always show 1 decimal, even if that decimal is 0. Eg, "50.0%". We may be able to add some custom code to strip trailing zeroes, though that would take some development.

And of course another option is to do nothing and accept that sometimes the numbers will add up to 101%.

LucyGwilliamAdmin:
@AnnCorp what are your thoughts on the rounding issue?

brockfanning:
@LucyGwilliamAdmin @AnnCorp To illustrate the problem, take the numbers for Goal 2:

* Fully disaggregated: 7.6923076923076925
* Partially disaggregated: 7.6923076923076925
* Not disaggregated: 7.6923076923076925
* Out of scope: 76.92307692307692

If I add those together in a calculator, I get 100 exactly. But if I round them all before adding them together, I get 101.

AnnCorp:
@LucyGwilliamAdmin @brockfanning the consensus from our management team is to go with rounded and have some text at the top like '“These percentages have been rounded. As a result, adding them together may not equal exactly 100%.”

phillipgardiner:
The above approach seems sensible at this time. What is the reason for the rounding error? Is this something that can ultimately be fixed at some point in the future? I'm only asking as some of the user testing feedback related to accuracy of data etc. Do other countries suffer from this currently then as outside the UK other countries tend to use more status labels?

LucyGwilliamAdmin:
Adding feedback from DAC:

1. When the page is magnified, some of the content does not reflow properly, I found that on iPhone, and on desktop when the page was magnified to 400%, the tabs sat on top of each other which made it difficult to identify their purpose, especially as the as the light grey background of the ‘Disaggregation status’ tab fails nontext contrast guidelines. This is probably the biggest issue identified.

    image

    Additionally, some of the content slighty overlaps:
    image

2. Something else which I would identify is that structural mark-up may make the information easier to read. This does not have to affect the visual presentation at all.
Example:

    ```
    <dl>
       <dt>
        <span class="status complete">18</span>
         <strong>Fully disaggregated</strong>
       </dt>
       <dd>
          <span class="value">7%</span>
       </dd>
       […]
    </dl>
    ```

3. Another thing our screen reader users do often pick up on is splitting headings across multiple components, as this can be confusing and appear as separate headings.

    image
 
    Current Code Ref(s): #goalsview > div:nth-child(3) > div.details > h3

    ```
    <h3 class="status-goal" data-element-id="headingsMap-3">
       <a href="/disagg-rs1/1">No poverty</a>
       <span class="total">14 indicators</span>
    </h3>
    ```

    I think this would work better as a `<hgroup>` as the text ‘No poverty’ and ‘14 indicators’ are really separate, with ‘14 indicators’ more of a subheading. 

    Example:

    ```
    <hgroup>
       <h3 class="status-goal" data-element-id="headingsMap-3">
          <a href="/disagg-rs1/1">No poverty</a>
       </h3>
       <h4><span class="total">14 indicators</span></h4>
    </hgroup>
    ```

4. For the grey and white striped pattern, the grey (#ccc) fail colour contrast guidelines against the white with a ratio of 1.6:1, so low vision users may struggle to distinguish any colour differences. However, as you have the equivalent information in text above the visualisation, this would probably not fail AA nontext contrast, but I would recommend using a darker grey.
image
 

5. The text ‘194’ over the varying background may be difficult for low vision users. This does not fail colour contrast guidelines, but a bolder font may assist users. Otherwise, perhaps consider a white background under the text ‘194’, on top of the gradient background, so that the text does not sit directly over the gradient background. 
image

6. In Windows High Contrast mode, background colours invert to black, so some visual information is lost. You have text alternatives, but it may assist low vision users in identifying the purpose of some of this content if you made some small fixes. This includes the number of indicators lacking a border and I know in the initial audit, Low Vision users mistook the bar visualisation for a scrollbar.
    image

    This is a quick fix as using a ‘border: 2px solid transparent’ will not affect the visual styling at all except for users in windows high contrast mode, as the borders become white. So it could look more like this:
    image

    I think the borders around the indicators help, and the visualisation isn’t a perfect fix but at least it doesn’t look like a scrollbar!

    I think the borders around the indicators help, and the visualisation isn’t a perfect fix but at least it doesn’t look like a scrollbar!

7. I also noted that the links for alternate colour schemes are indistinguishable in high Contrast Mode:

    image

    I don’t think this is an issue as users would need to be making use of alternate colour schemes already to notice this, but I just thought I would mention it. 

8. The ‘id’ attribute of the same images across both tabs are the same. This shouldn’t really cause any issues when one tab is hidden with ‘display:none;’ but it is something our automated tests bring up. 

9. There is no `<main>` landmark. Landmarks can be really helpful in presenting the page structure to screen reader users:

    Current Code Ref(s): #main-content

    ```
    <div id="main-content" class="container reportingstatus">[…]</div>
    ```

    Example:

    ```
    <main id="main-content" class="container reportingstatus">[…]</main>
    ```

brockfanning:
@phillipgardiner Here's a screenshot of some possible fixes for items 4 and 5 above. This changes the stripes to use that darker grey (the same grey as in "partially disaggregated") and puts a translucent background around the number in "out of scope". Any feedback?

status-contrast-fixes

brockfanning:
@LucyGwilliamAdmin I've gone through to try and get the quick fixes done, and pushed a new commit. Here are thoughts on each item:
1. This is a larger issue and and I think should be tackled separately, when we try to adopt [this approach](https://design-system.service.gov.uk/components/tabs/default/index.html).
2. I tried this, but the styles here are a bit fragile and it did not look good. I don't think this can be done quickly.
3. Same as above.
4. Possible fix in latest commit.
5. Possible fix in latest commit.
6. In the latest commit I've done the recommended fix for the number boxes, but the same fix caused issues in the horizontal bar. That will take a deeper dive to figure out. So this is only partly fixed.
7. Did not tackle here - I think this will take longer.
8. Possible fix in the latest commit.
9. Possible fix in the latest commit (and I also did the same on many other layouts in the platform).

brockfanning:
@phillipgardiner Also to answer your question about the rounding issue: It's not really an error to be fixed - more just a function of rounding in general. When you round numbers, you lose precision, and there's not much of a way around that. So this is more of a decision of preference: do we prefer the simplicity of rounded numbers with a chance of minor inaccuracy, or do we want to show decimal places to avoid the chance of inaccuracy?

phillipgardiner:
@brockfanning Thank you for the clear explanation, makes total sense now 👍 

In regard to the stripy progress bar. I think possibly DAC might still not be happy with the visibility of the number within the box - but can't be sure. @LucyGwilliamAdmin is this something we could quickly check with them?? Otherwise I think its fine (although tend to agree with @AnnCorp that it might be better swapping the white and stripy round in the order they are used, but this can be discussed).

LucyGwilliamAdmin:
@brockfanning I'm just about to go back to DAC with the fixes so far and ask for clarification on which parts are actually AA fails but looking at the original email it looks like it's the reflow issue (point 1) which is defo a fail and the rest aren't:

> I have reviewed the page this morning and it looks good. It seems to be working well. I didn’t identify much which I considered to be a WCAG fail, but I found a AA reflow issue and I do have some additional comments which I feel may make the site easier to use. 

So i'm thinking we may need to get a fix for that in but looking at the rebuild of the feature branch I think you've changed the styling so that it meets high contrast rules right? 
http://sdgdev-813006012.eu-west-1.elb.amazonaws.com/disagg-rs1/reporting-status/
image
Which may be enough in meantime - I'll check

brockfanning:
@LucyGwilliamAdmin No I didn't change anything related to tab styling. Ok I understand - I'll give top priority to implementing a mobile tab solution - hopefully the approach Phil recommends.

Just musing - I wonder why this didn't come up during the last audit. This same issue affects all the other tabs, like Chart/Table/Map, and Global/National/Sources/Edit.

brockfanning:
@LucyGwilliamAdmin I just added a possible shorter-term quick fix - this simply forces the tabs to go full width on smaller screens, and tries to make the active tab more visually obvious. Do you think this satisfies the DAC feedback?

phillipgardiner:
@LucyGwilliamAdmin @brockfanning Just feeding in about the accessibility of tabs. There is an audit trail of the ones used by GDS: https://design-system.service.gov.uk/components/tabs/. In the trail they noticed an accessibility issues with thier first version of the tabs (which looked similar to ours). Its looks like to rectify this they introduced underlines to the text?? You can see the full trail here: https://github.com/alphagov/govuk-design-system-backlog/issues/100.

LucyGwilliamAdmin:
Received the following feedback from DAC re. latest changes:

> The AA Reflow fail is just if content is lost when the page is magnified and related to the two comments with the tab panels and the overlapping content which was (1.) 

> Looking at this now, the tab panels look a lot better but the text does still overlap with the border of the containers for the indicators. I shouldn’t imagine this would cause users too much of an issue, but would always advise that it be clearly spaced:

 
image


> The patterned indicator looks a lot clearer to me now, and I checked with one of our low vision analysts who confirmed it’s much more readable now! 

> The borders on the indicators are working well, and if you want to include them on the bars, you would have to associate them with each of the colour blocks – I think when I tried this it sent some of the styling a bit awry. But this is not a WCAG fail because you have the information there in text anyway, it is just advised so that it would make the purpose of the visualisation a little clearer to those users navigating in this way.

@brockfanning by the looks everything is fine and more clear except the overlapping shown in the image - do you think this is something that could be fixed?

LucyGwilliamAdmin:
@brockfanning I still wasn't clear on whether the overlap was an AA fail so got clarification and it isn't so that's something we could fix in future.

In this case, I think we're all happy with this now? or was there something else?

@phillipgardiner @AnnCorp @jwestw

phillipgardiner:
@LucyGwilliamAdmin @brockfanning looks fine to me, we just need to make sure we capture the bits not tackled somewhere for future work 👍

LucyGwilliamAdmin:
@phillipgardiner okie doke - I'll make some issues then approve 👍

LucyGwilliamAdmin (approved):

PR #971, opened 2020-10-22
Documentation of data repository configuration

brockfanning:
@LucyGwilliamAdmin @jwestw Our documentation of the various data config settings is a bit obscure right now, being buried in the sdg-build library. What do you think about adding it to the Open SDG docs, like in this PR?

The only downside is that if we change the way sdg-build works, we'll need to remember to update these docs. But personally I'm OK with that, if it means we have better documentation overall.

### DIFF
--- a/docs/configuration.md
+++ b/docs/configuration.md
@@ -1,6 +1,8 @@
-<h1>Configuration</h1>
+<h1>Site configuration</h1>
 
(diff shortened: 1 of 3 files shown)


brockfanning:
@jwestw @LucyGwilliamAdmin Probably the best way to review this is on my fork in Github: https://github.com/brockfanning/open-sdg/blob/data-config-docs/docs/data-configuration.md

LucyGwilliamAdmin:
I think this is a really good idea - will open it up to less technical users if they need it

LucyGwilliamAdmin (approved):

PR #974, opened 2020-10-28
Correct metadata field name

LucyGwilliamAdmin:
@brockfanning was just looking at docs and realised that it's tags and not tag


brockfanning:
@LucyGwilliamAdmin It looks like a testing problem has made it into the master branch (unrelated to your change here). I'll do some investigation. I'll go ahead and approve this.

brockfanning (approved):

PR #975, opened 2020-10-29
Mention that remote_data_prefix can actually be a local folder

### LINKED ISSUE #754: Document that remote_data_prefix can be a local folder
There is the option to put the data and site into the same repository - which is what we use in the open-sdg-simple-starter repository. But we've neglected to document that `local_data_folder` option. (Example [here](https://github.com/open-sdg/open-sdg-simple-starter/blob/develop/config_site.yml#L12))

brockfanning:
Fixes #754

### DIFF
--- a/docs/configuration.md
+++ b/docs/configuration.md
@@ -507,6 +507,12 @@ plugins:
 remote_data_prefix: https://my-github-org.github.io/my-data-repository
 ```
 
+Note that this is typically a remote URL, but it also works as a relative path to a local folder on disk. For example:
+
+```nohighlight
+remote_data_prefix: my-data-build-folder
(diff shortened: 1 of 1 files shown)


LucyGwilliamAdmin:
I'm not sure I quite understand this one - is it for local development, or if someone has their data build stored within their site repo?

LucyGwilliamAdmin (approved):

brockfanning:
@LucyGwilliamAdmin Sorry - I forgot to reply - yes that's right.

brockfanning:
@LucyGwilliamAdmin Er, the latter - for if someone has their data stored in the site repository.

PR #980, opened 2020-11-12
1.2.0 release prep

brockfanning:
EDIT: Since I'm adding fixes and changelog updates to this PR, this can be a general 1.2.0 preparation PR.

This is a helper script to test the behavior of the codebase when a single file is reverted to the 1.0.0 version. It's intended to reveal potential breaking changes in the case that an implementation has overridden that file. Using this will be a very manual process, because many times the tests will fail for some reason that does not mean there is a breaking change. The usage would be something like this:

```
node tests/release/back-compat.js _layouts/goal.html
```

The above would look for breaking changes in the goal.html layout. We would need to run this script for every single file that has changed, so it will still be a time-consuming process, just a bit easier.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -8,3 +8,4 @@ make-cache
 venv
 /site
 .vscode
+version-*
--- a/Makefile
+++ b/Makefile
@@ -71,4 +71,8 @@ test.docs: build.docs
 	gem install html-proofer
 	htmlproofer site --disable_external
 
+test.only:
+	cd site-starter && bundle exec htmlproofer --file-ignore '/documentation/' --disable-external ./_site
+	cd tests && npx cucumber-js
+
 test: test.html test.features test.accessibility test.docs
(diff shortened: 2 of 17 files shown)


brockfanning:
I'm finding some breaking changes with this script, so I'll be including those fixes in the PR.

brockfanning:
@jwestw @LucyGwilliamAdmin This is ready for review. It includes a few instances of backwards-compatibility safety code and styles. This is intended to make sure that things generally work, no matter what files a country may have overridden.

There were just two files that I have not yet added backwards-compatibility for:

* _includes/components/charts/chart.html
* _includes/components/indicator/table.html

Adding backwards compatibility for those two files would be a lot of code, so I figured I would hold off until we discuss.

brockfanning:
Go through known implementations and see if any are overriding the chart.html and/or table.html files.

Create an "Upgrading to 1.2.0" document.

LucyGwilliamAdmin (approved):

PR #987, opened 2020-11-27
Use recent translation keys instead of hardcoded English

brockfanning:


brockfanning:
@LucyGwilliamAdmin @jwestw With our new Weblate integration I'm more comfortable with doing an English-only translation in sdg-translations and then using those translation keys right away. So this PR switches several bits of hardcoded English to use some new translation keys that I've added in sdg-translations (English only). My thinking here is that, when we release version 1.2.0, we will go into Weblate and perform machine translations for any untranslated keys.

LucyGwilliamAdmin:
@brockfanning just looked at this and think I'm happy to approve but have a quick question - what's the difference between `translations` and `page.t`?

brockfanning:
@LucyGwilliamAdmin "page.t" is for use in Jekyll templates, whereas "translations" is for use in Javascript.

LucyGwilliamAdmin (approved):
Happy to approve

Issue #892, opened 2020-09-14
Highlight only tab words rather than whole tab - on focus


phillipgardiner:
On the gov.uk implementation of tabs: https://design-system.service.gov.uk/components/tabs/. When the tab is focused, it is only the words that are highlighted rather than the whole tab. SDG currently highlights the whole tab.

This has previously been discussed with @brockfanning as part of accessibility changes but due to time required to implement and it not being critical it was decided to create a separate ticket to implement the above.

phillipgardiner:
Similarity with drop downs (like those used in the graph filters, only the outline of the drop down should be highlighted, not filled with yellow: https://design-system.service.gov.uk/components/select/

brockfanning:
The tab part of this was implemented. Should we close and open a new ticket for the drop-down change?

phillipgardiner:
@brockfanning Yes please

PR #990, opened 2020-12-01
Tab focus design fixes

brockfanning:
@phillipgardiner Can I get your feedback on this? I can't remember if we've discussed this. My thinking is that, since tab focus is activated constantly by all users (just by using the tabs) that it might be a good idea to have it be a bit more subtle - with a yellow outline instead of a full yellow background (which it is inheriting by virtue of being a link). Here's a before/after for this PR:

Before:

screenshot-127 0 0 1_4000-2020 12 01-06_39_02

After:

screenshot-127 0 0 1_4000-2020 12 01-06_31_52

### DIFF
--- a/_includes/components/indicator/data-tabs.html
+++ b/_includes/components/indicator/data-tabs.html
@@ -1,7 +1,7 @@
 <div style="display: none;">
   <span id="table-alternative">{{ page.t.indicator.data_tabular_alternative }}</span>
 </div>
-<ul class="nav nav-tabs data-view" role="tablist">
+<ul class="nav nav-tabs non-stacking-tabs data-view" role="tablist">
   <li role="presentation" class="nav-item active">
     <a class="nav-link" data-toggle="tab" id="tab-chartview" href="#chartview" aria-controls="chartview" role="tab" {% include autotrack.html preset="tab_data_chart" category="Tab change" action="Change data view" label="Change to Chart tab" %} aria-describedby="table-alternative">{{ page.t.indicator.chart }}</a>
   </li>
--- a/_sass/base/_typography.scss
+++ b/_sass/base/_typography.scss
@@ -107,6 +107,7 @@ input[type="radio"]:focus, body.contrast-high input[type="radio"]:focus,
 input[type="checkbox"]:focus, body.contrast-high input[type="checkbox"]:focus {
   box-shadow: 0px 0px 0px 4px $focusOutlineColor !important;
 }
+.nav-tabs .nav-item .nav-link:focus,
 .tabPanel:focus,
 input[type="text"]:focus,
 body.contrast-high input[type="text"]:focus {
--- a/_sass/base/_typography.scss
+++ b/_sass/base/_typography.scss
@@ -107,16 +107,10 @@ input[type="radio"]:focus, body.contrast-high input[type="radio"]:focus,
 input[type="checkbox"]:focus, body.contrast-high input[type="checkbox"]:focus {
   box-shadow: 0px 0px 0px 4px $focusOutlineColor !important;
 }
-.nav-tabs .nav-item .nav-link:focus,
 .tabPanel:focus,
 input[type="text"]:focus,
 body.contrast-high input[type="text"]:focus {
   background-color: $backgroundColor !important;
   box-shadow: none !important;
   outline: 3px solid $focusOutlineColor !important;
-}
-body.contrast-high {
-  .nav-tabs .nav-item .nav-link:focus {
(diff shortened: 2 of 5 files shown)


phillipgardiner:
@brockfanning hi, if we are truly aiming to use GDS as our default then we should probably consider following their lead with this: https://design-system.service.gov.uk/components/tabs/default/index.html. As you can see they highlight the text rather than the whole tab. Is something like this possible at all?

brockfanning:
@phillipgardiner That involves some "fighting" against the Bootstrap styles, and I've been assuming that we should wait until upgrading to Bootstrap 4 before tackling that. But I'm looking at Bootstrap 4's code now, and it's no better. So we may as well try it now - I'll give it a try. I agree that it would be better and more consistent.

brockfanning:
@phillipgardiner I think this might work out, here is a new screenshot:

screenshot-127 0 0 1_4000-2020 12 01-08_43_16

I also added underlines to the non-active tabs.

phillipgardiner:
@brockfanning just found this ticket: https://github.com/open-sdg/open-sdg/issues/892 doesn't say a lot more but we obviously did discuss it in the past.

phillipgardiner:
> @phillipgardiner I think this might work out, here is a new screenshot:
@brockfanning looking good 👍

brockfanning:
@LucyGwilliamAdmin @jwestw Do you think we could try to get this on a feature branch and review/merge this for 1.2.0? Since 1.2.0 is the release that includes our focus style revamp, it would be good to get this in now, so that we're not adding something and then changing it in a subsequent release.

In addition to using the GDS style for tabs, this fixes an unrelated problem that I notice while doing this: In the disaggregation status PR (which has now been merged) we added "stacking" to the tab on smaller screens. The problem is that this was applying to the Chart/Table/Map tabs. Those tabs are so short, that I don't think we want stacking there. So this PR also ensures that the Chart/Table/Map tabs do not stack on smaller screens.

LucyGwilliamAdmin (approved):
Looks good to me:
http://sdgdev-813006012.eu-west-1.elb.amazonaws.com/tab-focus/3-a-1

@phillipgardiner here's it in action if you want to take a look

phillipgardiner:
@LucyGwilliamAdmin @brockfanning Sorry I've spotted there seems to be something a bit weird with the lines around the tab. In the feature branch a line no longer exists on the bottom of inactive tabs, but does appear where is shouldn't' on the active tab. 

Feature branch: 
Screenshot (62)

Currently: 
Screenshot (63)

Hope that makes sense?

brockfanning:
@phillipgardiner @LucyGwilliamAdmin That's my fault - I had thrown in a transparent border to try and suppress the 1 pixel "shifting" effect when a tab becomes active. But that caused this new issue so I've reverted that change. The shifting is back now, but I guess I can live with it.

LucyGwilliamAdmin:
I've just rebuilt:
http://sdgdev-813006012.eu-west-1.elb.amazonaws.com/tab-focus/3-a-1/

brockfanning:
I've been trying different fixes for that shifting issue (it bothers me...) and just pushed up a possible fix. (5 minute too late, very sorry @LucyGwilliamAdmin!)

LucyGwilliamAdmin:
No probs - just rebuilt - I'm not too sure what you mean about shifting so can you check the fix please?

brockfanning:
The "shifting" was a slight 1 pixel shift whenever a tab become active/inactive. The feature branch is looking good for now - tested in Chrome, Firefox, Edge.

brockfanning:
@phillipgardiner How is it looking for you?

phillipgardiner:
@LucyGwilliamAdmin @brockfanning I get a 502 bad gateway error when trying to view the feature branch.

LucyGwilliamAdmin:
@phillipgardiner looks like the feature branch server went down over night - I have just restarted it so should work now (does for me now)

phillipgardiner:
@brockfanning @LucyGwilliamAdmin Looks really good. The only thing I have spotted and it looks like its not correct on the live site either is when you hover over the tab, the cursor should ideally change to a pointer, currently it stays as a line. Don't know if its worth fixing this now, or adding a new card to address this, or add it to the card thats looking at the styling when displayed on mobile?

brockfanning:
The cursor/pointer fix is done, and was very straightforward, so I don't think we need to do another feature branch refresh, and I'll go ahead and merge when the tests pass, if there are no objections before then. :)

LucyGwilliamAdmin:
@brockfanning sounds good with me!

PR #993, opened 2020-12-02
Support non-numeric years in maps

### LINKED ISSUE #108: Mapping: Non-standard/mixed year types
The map currently recognises when the year is in the format '2014'. However, for mixed years ('2014-15' or '2014/15') it becomes confused. Is there a way to have these mixed-year formats display on the site?

https://sustainabledevelopment-rwanda.github.io/1-2-1/

brockfanning:
Fixes #108 

Currently a "Year" in data can be non-numeric, like "2017-2019" for example. This works fine for charts and tables, but it breaks the maps. This attempts to provide better support for maps. The years must still *start* with a year, but can include anything else. For example this code should allow for the following year values:
* 2017-2018
* 2017.4
* 2017 Q4
* 2017/4

Note: For testing this will involve creating a data branch too - will need to change a data file for an indicator that has a map e.g. 3.a.1 so that values in year column have something other than the year (don't need to change all though - just headline)

[Feature Branch - indicator 3-a-1](http://sdgdev-813006012.eu-west-1.elb.amazonaws.com/non-numeric-years/3-a-1/)

### DIFF
--- a/_includes/assets/js/plugins/jquery.sdgMap.js
+++ b/_includes/assets/js/plugins/jquery.sdgMap.js
@@ -368,7 +368,7 @@
         plugin.map.addControl(L.Control.yearSlider({
           years: plugin.years,
           yearChangeCallback: function(e) {
-            plugin.currentYear = new Date(e.time).getFullYear();
+            plugin.currentYear = plugin.years[e.target._currentTimeIndex];
             plugin.updateColors();
             plugin.updateTooltips();
             plugin.selectionLegend.update();
--- a/_includes/assets/js/plugins/leaflet.yearSlider.js
+++ b/_includes/assets/js/plugins/leaflet.yearSlider.js
@@ -31,7 +31,14 @@
 
     // Hijack the displayed date format.
     _getDisplayDateFormat: function(date){
-      return date.getFullYear();
+      var time = date.toISOString().slice(0, 10);
+      var match = this.options.years.find(function(y) { return y.time == time; });
+      if (match) {
+        return match.display;
+      }
+      else {
+        return date.getFullYear();
+      }
     },
 
     // Override the _createButton method to prevent the date from being a link.
@@ -106,6 +113,7 @@
 
   // Helper function to compose the full widget.
   L.Control.yearSlider = function(options) {
+    var years = getYears(options.years);
     // Extend the defaults.
     options = L.Util.extend(defaultOptions, options);
     // Hardcode the timeDimension to year intervals.
@@ -113,8 +121,8 @@
       // We pad our years to at least January 2nd, so that timezone issues don't
       // cause any problems. This converts the array of years into a comma-
       // delimited string of YYYY-MM-DD dates.
-      times: options.years.join('-01-02,') + '-01-02',
-      currentTime: new Date(options.years[0] + '-01-02').getTime(),
+      times: years.map(function(y) { return y.time }).join(','),
+      currentTime: new Date(years[0].time).getTime(),
     });
     // Create the player.
     options.player = new L.TimeDimension.Player(options.playerOptions, options.timeDimension);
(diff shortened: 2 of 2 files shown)


jwestw:
This is the feature branch: http://sdgdev-813006012.eu-west-1.elb.amazonaws.com/non-numeric-years

I am working on changing the data now.

jwestw:
For some reason I can't get the data I have changed to show up in the feature branch I have made althoguht I have pointed the remote data prefix to: 
```remote_data_prefix: "http://sdgdev-813006012.eu-west-1.elb.amazonaws.com/non-numeric-years-data"```

But at least we can see that this is somewhat working in the data preview: 
http://sdgdev-813006012.eu-west-1.elb.amazonaws.com/datapreview_non-numeric-years-data/3-a-1/
i.e. it hasn't broken the maps.

brockfanning:
@jwestw My impression was that the data branches get "data/" added before the branch, so the remote_data_prefix should be: http://sdgdev-813006012.eu-west-1.elb.amazonaws.com/data/non-numeric-years-data/

jwestw:
Thanks for your help Brock. This is working in the [feature branch](http://sdgdev-813006012.eu-west-1.elb.amazonaws.com/non-numeric-years/3-a-1/) now. 

I am happy that the maps are functioning as expected with this change. 

Is there anything else you'd like me to do with this data or are you happy with that?

brockfanning:
@jwestw I think I misspoke about the "just headline" part. In order to see the non-numeric values on the map, they need to be on rows that have Countries. Like in [this file](https://github.com/ONSdigital/sdg-data/blob/non-numeric-years-data/data/indicator_3-a-1.csv), lines 8, 15, 22, 29, 113, 120, 127, and 124. (You could keep going but that should be enough, maybe changing 2011 to "2011 Q1" and 2012 to "2012 Q1"?

jwestw:
I have altered all of the [data for 3-a-1](https://github.com/ONSdigital/sdg-data/blob/non-numeric-years-data/data/indicator_3-a-1.csv) in the year column so as to be non-numeric. 

While this works well in the maps (which is what [the ticket ](https://github.com/open-sdg/open-sdg/issues/108)is about) it causes some slightly odd behaviour with the graphs and the tables as both are showing not only the non-numeric years but also a numeric year (presumably parsed from the non-numeric). 

**Graphs**
image
The numeric years are on the x axis with no data points plotted. 

**Tables**
image
The numeric years are in the table with no values. 

For the behaviour with the maps it is a pass. But with regards to the tables and graphs, is this what is expected?

brockfanning:
@jwestw I definitely would not expect years to appear on the site when they are not in the data. Is the feature branch still http://sdgdev-813006012.eu-west-1.elb.amazonaws.com/non-numeric-years/3-a-1/ because I'm not seeing this behavior there.

LucyGwilliamAdmin:
I just had quick check and I'm also not seeing that behaviour on the branch Brock linked to.

However even though the year slider seems to be working, data doesn't seem to be showing on the map:
image

LucyGwilliamAdmin:
I take that back - I've just refreshed and data is showing on map now

jwestw:
Yes, the page I am looking at is [here](http://sdgdev-813006012.eu-west-1.elb.amazonaws.com/non-numeric-years/3-a-1/). I was seeing the behaviour I reported until I did a hard refresh and emptied the cache. 

I think what I was seeing may have been due to the data before the most recent commit having some inconsistent years.  The easiest way to see those is to see what I replaced in the [recent commit](https://github.com/ONSdigital/sdg-data/commit/d160d696b5cca50acf5e191bbb48a138358fdc8f). Still, none of the years were purely numeric (e.g. 2012) like those I was seeing in the tables and charts, so it's still a bit unexplained. 

Now it has been hard refreshed it's behaving as we would expect anyway.

brockfanning:
@jwestw Thanks, I'll revisit this and investigate.

brockfanning:
@jwestw I just pushed a change that I think should fix it.

jwestw:
Excellent. The maps are working perfectly now. 

I still notice though that some of the data points are not being data points are not being displayed on tables and charts. 

In the data you will see values for the Year "2012-Q1".

image

However, this is not showing on the graph: 

image

Nor is it showing on the table:

image

(Other data is missing too, clearly). 

Sorry, I feel like I've really minced the values in the Year column so this has really thrown something unusual and possibly not realistic at the code, but it does seem to have broken it. 

By the way, I have emptied the cache and hard reloaded just in case it was my browser again - but it's still displaying like this.

brockfanning:
@jwestw I haven't gone through every row, but at first glance I don't see any rows that are "2012-Q1" without any other disaggregations. To use the terminology we use in the code, there is no "headline" row for 2012-Q1. So it would be expected for "2012-Q1" to be missing from the charts/tables until a disaggregation is selected.

Regardless though, since this issue is only about maps, and the files changed are only map-related, it is safe to say that this PR is not causing any new problems with the charts/tables. So I think even if there are problems with non-numeric years in the charts/tables, that could be a new issue.

jwestw:
Yes I agree. The maps are supporting non-numeric years.

jwestw (approved):
Approving according to last comment.

PR #995, opened 2020-12-03
Allow alteration of DataTable configuration and fix tables for "binary" indicators

### LINKED ISSUE #1013: Table for "binary" indicators shows 1/-1 instead of Yes/No
**Describe the bug**

We have special treatment of the "binary" graph type, which converts 1 to "Yes" and -1 to "No". However this only happens on the charts. On the table, the 1/-1 still displays.

**To Reproduce**

Here is an example: https://sdg.data.gov/17-18-3/
Click on "Table" to see the issue.

**Expected behavior**

The values should be "Yes" instead of 1 and "No" instead of -1.

brockfanning:
Fixes #1013 

We already allow alteration of Chart.js configuration through the `opensdg.chartConfigAlter` function. Along the same lines, this allows alteration of the DataTables configuration through `opensdg.tableConfigAlter`. A possible use-case might be changing the display of some data value to a more human-friendly version.

To demonstrate this also fixes issue #1013 by converting "1" into "Yes" and "-1" into "No" in the "binary" graph type.

Note: This requires a data branch - will need to change graph_type to binary and make sure data complies to requirements set out here: https://open-sdg.readthedocs.io/en/latest/charts/#binary

### DIFF
--- a/_includes/assets/js/indicatorView.js
+++ b/_includes/assets/js/indicatorView.js
@@ -323,6 +323,12 @@ var indicatorView = function (model, options) {
     });
   };
 
+  this.alterTableConfig = function(config, info) {
+    opensdg.tableConfigAlterations.forEach(function(callback) {
+      callback(config, info);
+    });
+  };
+
   this.updateChartTitle = function(chartTitle) {
     if (typeof chartTitle !== 'undefined') {
       $('.chart-title').text(chartTitle);
@@ -597,7 +603,7 @@ var indicatorView = function (model, options) {
     }
   };
 
-  var initialiseDataTable = function(el) {
+  var initialiseDataTable = function(el, info) {
     var datatables_options = options.datatables_options || {
       paging: false,
       bInfo: false,
@@ -609,6 +615,7 @@ var indicatorView = function (model, options) {
 
     datatables_options.aaSorting = [];
 
+    view_obj.alterTableConfig(datatables_options, info);
     table.DataTable(datatables_options);
     table.removeAttr('role');
     table.find('thead th').removeAttr('rowspan').removeAttr('colspan').removeAttr('aria-label');
@@ -829,8 +836,12 @@ var indicatorView = function (model, options) {
 
       $(el).append(currentTable);
 
-      // initialise data table
-      initialiseDataTable(el);
+      // initialise data table and provide some info for alterations.
+      var alterationInfo = {
+        table: table,
+        indicatorId: indicatorId,
+      };
+      initialiseDataTable(el, alterationInfo);
 
       $(el).removeClass('table-has-no-data');
       $('#selectionTableFooter').show();
--- a/_includes/components/charts/binary.html
+++ b/_includes/components/charts/binary.html
@@ -53,10 +53,14 @@ opensdg.chartConfigAlter(function(config) {
 });
 
 opensdg.tableConfigAlter(function(config, info) {
+  var targetColumns = [];
+  for (var i = 1; i < info.table.headings.length; i++) {
+    targetColumns.push(i);
+  }
   var overrides = {
     columnDefs: [
       {
-        targets: -1,
+        targets: targetColumns,
         createdCell: function(td, cellData, rowData, row, col) {
           $(td).text(opensdg.convertBinaryValue(cellData));
         },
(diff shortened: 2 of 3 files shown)


jwestw:
Feature branch:  http://sdgdev-813006012.eu-west-1.elb.amazonaws.com/table-config-alter

brockfanning:
@jwestw This may unfortunately be a pain to test with feature branches, because of the fact that the UK is not using the binary graph type. The only way I can imagine would be:
1. Create a data branch where one indicator is switched to the "binary" graph type, and its data values to only "1" and "-1".
2. Create a feature branch using open-sdg/open-sdg@master, pointing to the data branch in remote_data_prefix.
3. Create a feature branch using brockfanning/open-sdg@table-config-alter, pointing to the data branch in remote_data_prefix.

Then we could compare the behavior of the two feature branches to see how this PR performs.

jwestw:
Thanks for the pointers @brockfanning

jwestw:
The feature branch pointing at brockfanning/open-sdg@table-config-alter built successfully and the binary graph is working here: http://sdgdev-813006012.eu-west-1.elb.amazonaws.com/datapreview_table-config-alter-data/15-3-1/

jwestw:
Currently the data has two dimensions (Year, Type) and a value. 
On each line both Year and Type are changing. 
The data table shows yes or no as we would expect when one Type is selected. However, when multiple types are selected -1 and 1 appear for those additional selections.

brockfanning:
@jwestw Thanks, I've pushed a commit to try and fix that. Also I think the feature branch with this PR is [here](http://sdgdev-813006012.eu-west-1.elb.amazonaws.com/table-config-alter/15-3-1/) right?

jwestw:
Yes. And that's fixed that. 

However I tried to upload a different data set https://github.com/ONSdigital/sdg-data/commit/03e2d60832dae74812ad8169efb7e3d6bf87c589

But the build keeps failing with errors. "- Graph type (graph_type/anyOf): 'binary' is not valid under any of the given schemas
"

image


Do you have any idea why binary is not accepted?

brockfanning:
@jwestw Yes I think so - I think it needs to be added in _prose.yml. I've just done this on the data starter, so here is an example: https://github.com/open-sdg/open-sdg-data-starter/commit/ce293c07131268dc9a4463f1283113c9562a8667

jwestw:
I can't seem to get this indicator working. 
http://sdgdev-813006012.eu-west-1.elb.amazonaws.com/datapreview_table-config-alter-data/15-4-1/

The data for it is here: https://github.com/ONSdigital/sdg-data/blob/table-config-alter-data/data/indicator_15-4-1.csv
And the meta data here: https://github.com/ONSdigital/sdg-data/blob/table-config-alter-data/meta/15-4-1.md

They both seem correct to me and they are on the same branch as the other indicator above, so the remote theme is the same. Not sure why this is not working though

brockfanning:
@jwestw I'm not sure what that "datapreview_table-config-alter-data" folder is coming from. From what I can tell the folder should be "table-config-alter", which would make the link this: http://sdgdev-813006012.eu-west-1.elb.amazonaws.com/table-config-alter/15-4-1/

I thought the folder corresponds to a branch in the onsdigital/sdg-indicators repo, and there is not a "datapreview_table-config-alter-data" branch - only a "table-config-alter" branch.

james-westwood:
Hey @brockfanning I am on a different computer now and the link I posted before is working (as is the one you posted). Seems like this is all good to go. 

About my previous comment and to answer your query (sorry if I am telling you what you already know here, but just for clarity). I should *not* have linked to the data feature branch in [that comment](https://github.com/open-sdg/open-sdg/pull/995#issuecomment-750141983). 
When I make feature branch of the indicators which requires data or meta data being changed for testing purposes then I also have to create a data repo feature branch. I typically call that something slightly different (in this case adding "-data" as a suffix) which just helps me differentiate as I am going between tabs. That data folder is [on a branch of our data repo](https://github.com/ONSdigital/sdg-data/tree/table-config-alter-data).

LucyGwilliamAdmin:
@jwestw looks like this one is ready? if so, can you approve please?

jwestw:
Okay [15.4.1](http://sdgdev-813006012.eu-west-1.elb.amazonaws.com/datapreview_table-config-alter-data/15-4-1/) which uses a different yes/no, 1/-1  value for each year is working well. 

[15.3.1](http://sdgdev-813006012.eu-west-1.elb.amazonaws.com/datapreview_table-config-alter-data/15-3-1/) which has data like this 
image

where both a category and a year changes doesn't display so well - but maybe this does not reflect real data.

jwestw (approved):
Approving as per last comment.

PR #1002, opened 2020-12-04
Left aligned header, search box awareness, search results-page tweaks

### LINKED ISSUE #666: Remove navigation search when on search results page
When a user is on the search results page. The search box in the navigation bar should NOT be visible. This will stop confusion about which search box to use when they wish to undertake a second subsequent search. Therefore the top search is visible on all pages apart from search results.

### LINKED ISSUE #782: Allow for a configurable header with the menu aligned left

Some aspects of the current header are problematic and would benefit from changes. For example the menu items could be more clear if they were aligned to the left, separate from the language/contrast toggles.

However we do not want to change this for all existing implementations of Open SDG. So we would like to create this as a new optional header.

Solution to be discussed in this ticket, but mainly the goal is to get the menu items aligned to the left, and do something else with the language/contrast toggles.

To improve UX.

We'll discuss alternatives in this ticket.

### LINKED ISSUE #996: Increase search box awareness
Some people have difficulty knowing which goals the data they are looking for would come under (but once search was used, this task becomes a lot easier)

Consider:
1) making the search box outline in the top right of the page a deeper shade of black.
2) increase the search box size to allow 27 visible characters to be displayed (accommodates 90% of queries)
3) potentially change the magnifying glass to a button labelled search (separated from the search field)
4) change the colour of the search button to something that highlights it more (GDS green)

To make the search box more obvious to ensure users can find what they're looking for

brockfanning:
Fixes #782
Fixes #996 
Fixes #666 

I combined multiple tickets into this PR because they all affect the same code and it would have been inefficient to track them separately. Seeing #996 and #666 does not require any steps other than using this branch. The rest of this PR description is in reference to #782.

## Usage

To try switching to the include "menu left aligned" header, use this site config:

```
header:
    include: header-menu-left-aligned.html
```

Not technically a requirement for testing, but this PR corresponds with this config form update: https://github.com/open-sdg/jekyll-open-sdg-plugins/pull/63

Feature branch: http://sdgdev-813006012.eu-west-1.elb.amazonaws.com/left-aligned-header/


brockfanning:
@LucyGwilliamAdmin @jwestw This is ready for @phillipgardiner to take a look at, so a feature branch would be much appreciated.

phillipgardiner:
@brockfanning Posting here as the original issue #1004 has been closed. They look good. My question (as not able to currently see it it action) is how the yellow boxing that would appear around the search would work? As the form filed and button are all in one, would it highlight the whole thing as a single element? My expectation would be that you tab onto the form field (and that highlighted) and then tab again on to the magnifying glass button (which is then high lighted). Like this example: https://www.gov.uk/.

jwestw:
The feature branch is here: http://sdgdev-813006012.eu-west-1.elb.amazonaws.com/left-aligned-header/

jwestw:
Compare

Current live site: 

image

New:

image

jwestw:
Is this still to be discussed? The changes look good code/functionality wise but I am not sure if this is what has been agreed.

phillipgardiner:
@brockfanning do you happen to have a screen shot of the menu left aligned but with the language selector present. I was interested just to see how this looks along side the 'A' icons?

brockfanning:
@phillipgardiner Sure, I can make some now. In the meantime I forgot to mention something on the call - do you think we need to also change the button to the word "Search" or is it OK to leave it as-is?

brockfanning:
Here are those screenshots:

left-aligned-language-dropdown-2

left-aligned-language-dropdown

left-aligned-language-links

brockfanning:
Oops, I missed the menu in those. One sec.

phillipgardiner:
> @phillipgardiner Sure, I can make some now. In the meantime I forgot to mention something on the call - do you think we need to also change the button to the word "Search" or is it OK to leave it as-is?

@brockfanning Leave as it is I say and use the magnifier glass

phillipgardiner:
@brockfanning @LucyGwilliamAdmin @AnnCorp sorry about the large number of comments, there is a lot to get out of my head!

1) On a bit of a separate note (not sure if separate ticket is required). I believe we need a black circle around the 'disable high contrast' A option. It now looks a bit odd floating on its own. Interestingly I notice when the window is sized for mobile, only one 'A' is visible and acts as a toggle on / off - which does raise the question is a  'disable high contrast' option required on desktop view when you are not on the high contrast version (and visa versa) - to be discussed?
2) I'm thinking that maybe we should make the menu navigation options black text - to be discussed?
3) The search box now definitely jumps out but have we overdone it? as it now grabs your eye over everything else. Is it possible to try making the black border slightly less thick?
4) When the yellow border is displayed around the search box, it should ideally only highlight the text field first and then the button. See example: https://www.gov.uk/ It looks like the difference with our version could be being caused by the button sitting inside the text field rather than outside it? Can the button be moved to outside the text field? 
5) On the search results page it looks as if the search button is not correctly placed (there is a gap to the right). Again could this be addressed by moving the search button outside the text field?
6) On the search results page I think we can make the search box wider, to give it more dominance on the page. I suggest making it as wide as would fit on a mobile view.

That's it!

brockfanning:
@phillipgardiner I think that numbers 1 and 2 might make sense as separately tickets, since I'm assuming they are good ideas even with the normal right-aligned header.

I'll try to implement the rest of the items in this ticket.

brockfanning:
@phillipgardiner @jwestw The items 3-6 are ready for an updated feature branch and review.

jwestw:
Feature branch re-built.

phillipgardiner:
@brockfanning Thanks for doing this. my comments are below:

1) I think the search box outline is now too subtle. I have been playing around with 'inspect' on the web browser and if I can the box shadow from 1px to 1.25px I think this has the desired effect.
2) The only issue I have noticed is that the search button doesnt seem to quite line up with the search box. I have provided some screen shots. The first screen shot is the homepage. The seconds is when I make the page smaller. And interestingly the minute the mobile navigation appears it then aligns (third screen shot).

Screenshot (72)
Screenshot (70)
Screenshot (71)

Thats it,

brockfanning:
@phillipgardiner Do you remember which browser you were using for these screenshots?

phillipgardiner:
@brockfanning google chrome.

brockfanning:
I know the feature-branch server is down at the moment, but this is ready for another feature branch rebuild whenever it's back up.

LucyGwilliamAdmin:
@brockfanning @phillipgardiner feature branch rebuilt

phillipgardiner:
@brockfanning @LucyGwilliamAdmin Good work brock, All looks good from my perspective 👍

LucyGwilliamAdmin (approved):

Issue #1000, opened 2020-12-04
Accessibility of config forms


LucyGwilliamAdmin:
Feedback from DAC re. site and indicator config forms:
Note: "I" is from tester not from me

I have had a look at the first URL (http://sdgdev-813006012.eu-west-1.elb.amazonaws.com/release-prep-and-config-forms/config/) this morning which I have used one hour for. I imagine that issues will be consistent across the two pages, but please let me know if you would like me to use another hour for the second page. I have recorded some comments which would likely be accessibility barriers and that I would expect to be identified if you were to have an expert review or a full audit. I imagine that fixing these would go a substantial way to making the pages fully accessible, and I didn’t notice anything else immediately obvious to myself as an issue within the high level look I was able to give it in the hours consultancy slot. However please do note that this has not been fully tested with all the technologies we would use for a full audit, and that other issues may still be present. I have recorded some issues below that I would flag.

1. Unlabelled form fields (A issue – name, role and value)
  
    ![image](https://user-images.githubusercontent.com/43404358/101171139-45f94f80-3637-11eb-8a7e-02e0c772bf62.png)

    Please note, I expanded some of the sections and ran axe on this which noted 128 form fields without programmatically associated labels. This means they have no accessible name to present their purpose to screen reader users or to allow voice activation users to reference them to their assistive technology by the visible label. 

2. Headings don’t reflect hierarchy or indicate sections or regions on the page (A issue – info and relationships)

    ![image](https://user-images.githubusercontent.com/43404358/101171252-67f2d200-3637-11eb-9d0d-e8a7deb160cf.png)

    This may be particularly confusing as there is a button which expands the ‘breadcrumbs section’. This triggers content to appear further down the page, but when this content is appears, because the headings are not subheadings of ‘breadcrumbs’ there is nothing to indicate to screen reader users that this is the newly triggered content which belongs to the visually indicated (by a grey background) ‘breadcrumbs section’. 

3. When the page is magnified, horizontal scrolling is induced due to the use of tables
    
    ![image](https://user-images.githubusercontent.com/43404358/101171357-89ec5480-3637-11eb-8ebc-21fe860d0e7b.png)

4. JAWS was not triggering the input type colour when I tested in Chrome – it seemed to only activate the link above. I tested JAWS with the same type of field on another site and it worked with Chrome, so I don’t believe this is a JAWS issue, but I would need to have a bit of a dig to try and work out why this isn’t recognised. The page wouldn’t load in IE for me. 

[...]

PR #1003, opened 2020-12-05
Configuration form accessibility fixes

brockfanning:
This attempts to address the items in #1000. Changes include:

* Remove all "table formatting" (ie, fields in rows) and instead display all fields vertically
* Remove the Markdown editor and show plain text areas instead
* Labels for integer fields (eg "Goal color number")
* Adds a message to links about opening in a new tab (screenreaders only)
* Use a button for the search filter
* Adds screenreader status announcement of search filter results

This also fixes IE support in general, which had been broken by recent changes.

A couple of items from #1000 are not fixed here:
* Item 4 - related to color fields: Since we're only using the standard HTML5 input for color, I would want to get confirmation that we actually need to remove those.
* Items 2 and 9 (I think these are the same right?): This is going to be a more difficult fix, and I thought I would submit this PR with the easier items.

### DIFF
--- a/_includes/assets/js/indicatorView.js
+++ b/_includes/assets/js/indicatorView.js
@@ -754,7 +754,9 @@ var indicatorView = function (model, options) {
 
   this.createIndicatorDownloadButtons = function(indicatorDownloads, indicatorId, el) {
     if (indicatorDownloads) {
-      for (var buttonLabel of Object.keys(indicatorDownloads)) {
+      var buttonLabels = Object.keys(indicatorDownloads);
+      for (var i = 0; i < buttonLabels.length; i++) {
+        var buttonLabel = buttonLabels[i];
         var href = indicatorDownloads[buttonLabel].href;
         var buttonLabelTranslated = translations.t(buttonLabel);
         var gaLabel = buttonLabel + ': ' + indicatorId;
--- a/_includes/assets/js/plugins/jquery.sdgMap.js
+++ b/_includes/assets/js/plugins/jquery.sdgMap.js
@@ -63,7 +63,8 @@
       var colorRangeParts = options.mapOptions.colorRange.split('.'),
           colorRange = window,
           overrideColorRange = true;
-      for (var colorRangePart of colorRangeParts) {
+      for (var i = 0; i < colorRangeParts.length; i++) {
+        var colorRangePart = colorRangeParts[i];
         if (typeof colorRange[colorRangePart] !== 'undefined') {
           colorRange = colorRange[colorRangePart];
         }
--- a/_layouts/config-builder.html
+++ b/_layouts/config-builder.html
@@ -132,7 +132,9 @@
               }
             }
             if (typeof propertyEditor.editors !== 'undefined') {
-              for (var editorKey of Object.keys(propertyEditor.editors)) {
+              var editorKeys = Object.keys(propertyEditor.editors);
+              for (var i = 0; i < editorKeys.length; i++) {
+                var editorKey = editorKeys[i];
                 propertyEditorAccessibilityFixes(propertyEditor.editors[editorKey]);
               }
             }
@@ -200,7 +202,8 @@
       if (errors.length) {
         var alertText = 'Please correct the errors listed below';
         var fieldNames = []
-        for (var error of errors) {
+        for (i = 0; i < errors.length; i++) {
+          var error = errors[i];
           fieldNames.push(error.path.replace('root.', ''));
         }
         alert(alertText + ' (' + fieldNames.join(', ') + ').');
--- a/_layouts/config-builder.html
+++ b/_layouts/config-builder.html
@@ -159,6 +159,7 @@
           case 'string':
           case 'integer':
           case 'boolean':
+          case 'number':
             propertyEditor.input.setAttribute('id', propertyEditor.formname);
             propertyEditor.label.setAttribute('for', propertyEditor.formname);
             if (typeof propertyEditor.description !== 'undefined') {
--- a/_layouts/config-builder.html
+++ b/_layouts/config-builder.html
@@ -81,6 +81,15 @@
       {% endif %}
     {% endif %}
 
+    function normalizeCurrentValue(key, value) {
+      // @deprecated start
+      if (key == 'create_pages' && value.pages) {
+        return value.pages;
+      }
+      // @deprecated end
+      return value;
+    }
+
     // Change the theme to create div instead of H3 elements for headers.
     JSONEditor.defaults.themes.bootstrap3.prototype.getHeader = function(text) {
       var el = document.createElement('div')
@@ -179,7 +188,7 @@
     {%- endif -%}
     {% if current_value %}
       try {
-        var currentValue = {{ current_value | jsonify }};
+        var currentValue = normalizeCurrentValue('{{ config_key }}', {{ current_value | jsonify }});
         var propertyEditor = editor.getEditor('root.{{ config_key }}');
         propertyEditor.setValue(null);
         propertyEditor.setValue(currentValue);
--- a/_layouts/config-builder.html
+++ b/_layouts/config-builder.html
@@ -215,6 +215,16 @@
         SaveAsFile(yaml, '{{ page.config_filename }}', 'text/plain;charset=utf-8');
       }
     });
+
+    function addNewTabMessage(link) {
+      if (!link.querySelector('.sr-only')) {
+        link.insertAdjacentHTML('beforeend', '<span class="sr-only">(opens in a new tab)</span>');
+      }
+    }
+
+    document.querySelectorAll('a[target="_blank"]').forEach(function(link) {
+      addNewTabMessage(link);
+    });
   </script>
 </div>
 {% include footer.html %}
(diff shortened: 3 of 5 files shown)


brockfanning:
@LucyGwilliamAdmin @jwestw This PR includes some accessibility fixes for the config forms, but it also includes a general platform-wide fix for IE support. **That IE issue definitely should be fixed before we launch 1.2.0.** Let me know if I should split it out into a separate PR.

brockfanning:
@jwestw @LucyGwilliamAdmin I have pushed some code that changes the H3 elements to divs. I think this should help with items 2 and 9 from the DAC feedback. This is now ready to be reviewed and/or sent back to DAC (with the only unaddressed item being #4 about the colors).

LucyGwilliamAdmin:
@brockfanning Thanks for letting me know - I will send back to DAC today

LucyGwilliamAdmin:
@brockfanning before I send - is Create pages table what is expected now?
http://sdgdev-813006012.eu-west-1.elb.amazonaws.com/config-form-accessibility-fixes/config/

brockfanning:
@LucyGwilliamAdmin Hmm, no it's not. Can you try pointing to the master branch of jekyll-open-sdg-plugins?

LucyGwilliamAdmin:
@brockfanning done and rebuilt - I think that's more what I was expecting - what about you?

brockfanning:
@LucyGwilliamAdmin Sorry, still some issues. Please hold off, I'll investigate more.

brockfanning:
@LucyGwilliamAdmin I've added a bit of backwards compatibility for the create_pages setting. I think this should be ready again for review.

LucyGwilliamAdmin:
I have just rebuilt: http://sdgdev-813006012.eu-west-1.elb.amazonaws.com/config-form-accessibility-fixes/config/

brockfanning:
@LucyGwilliamAdmin There's still an issue with the date formats setting. I'll work on that. It might be OK to send to DAC anyway if you want to avoid delays (and if you're tired of making feature branches :).

LucyGwilliamAdmin:
@brockfanning I drafted the email ready to send earlier but I ended up sending by accident so I've already got some feedback from them...

These comments are from having a brief look at page - I think once the additional feedback has been looked at/resolved, we will go back for a fuller review. However I will book in that time now as they like to know in advance - do you have any thoughts on when maybe it should be?

re. 4 from DAC: 

> No I wouldn’t remove them, but without a proper look I am not sure why they weren’t working with screen reading software to offer a fuller suggestion.  I tested the standard components on a different site and they did seem to work with JAWS so I would imagine that there must be something within the mark-up causing this issue. Apologies that was all the digging I was able to do within that space of time. If you would like me to book some time in to take a more thorough look, please let me know. If these don’t work with screen reading software and there is no alternative, then this could be a high level fail. I think I would probably advise including an additional alternative way for users to enter the hexadecimal colour code manually via a text input component with a clear label and instructions. This would mean that users could quickly enter the code without using this component if they knew it offhand. 

Other feedback from DAC:
> You have multiple links with the same link text ‘more information’ but different purposes. I would advise you to keep the link text fully descriptive, for example, ‘More information on Goals page (opens in a new tab)’. At the moment the text ‘(opens in a new tab)’ is visually hidden and only presented to screen reader users, but this is actually something that a wide range of users bring up as an issue so I would suggest this information be presented visually on the page to all users. I would consider this a AAA non-conformance.

> It appears that headings are not marked-up on this page either now. This would be a A fail (information and relationships). 

> One issue I would also identify which is more of a usability issue is that instructional text is included under the fields to which they correspond, and they are not referenced with an ‘aria-describedby’ as recommended by the GOV.UK Design System. 

> One thing that I would probably also want to look further at would be how required fields are indicated for form groups within tables, such as ‘Main menu’, as the text ‘required’ is not programmatically associated with the fields, users may not be able to identify which fields are required. I believe the GOV.UK Design System recommends indicating instead when fields are optional, and this being a part of the labels or legend for form groups.

brockfanning:
That was fast!

Regarding the color fields, I really don't want to complicate things with duplicating them. I would vote that we either leave it as is, or turn them all into plain text fields. Do you agree and/or have a preference? Personally I'm fine either way. I'd be a bit surprised if screenreaders don't support this HTML5 field, but stranger things have happened, and these color widgets are not critical to the overall feature.

Regarding this feedback -- "You have multiple links with the same link text ‘more information’ but different purposes." -- I thought I had already added specifics on each link. I may well have missed some though. A specific example would be helpful.

Regarding this feedback -- "I would suggest this information be presented visually on the page to all users" -- this should now be fixed.

Regarding this feedback -- "It appears that headings are not marked-up on this page either now." -- I will need to understand this better. Is the problem that the labels of the fields are large font and appear to be headings - and thus they need to actually be headings? Is there any way to satisfy this without making them headings? They were H3's before but I changed them to DIV to solve the hierarchy problem. Because of the recursive nature of this javascript library ("json-editor"), whatever element we choose will be repeated in a nested way.

Regarding this feedback -- "they are not referenced with an ‘aria-describedby’" -- In some cases they actually are. But I am not surprised if some are not. If it's easy to provide, a specific example would help me fix this.

The issues with "required" is certainly a complex issue, because of the way this javascript library works. The comment above about required fields seems a bit too vague for me to get started. I would prefer to have a clear description of the AA requirements there.

Stepping back a bit for an overall comment though: I'm not clear how much we need to fix before 1.2.0. I believe that every time we send this page to DAC, they are going to find more problems...

LucyGwilliamAdmin:
@brockfanning 

I was thinking same for colour fields - doesn't make sense to duplicate - I do prefer it as it is now as users can preview and choose colour in one place. We are going to ask them to spend some more time on the forms so they might be able to suggest how to fix.

From what I can see all "more information..." have been updated

However there does seem to be a space missing between "more information..." and "(opens in a new tab)" 

I will clarify about hierarchies and headings

Just checked out the aria-described by issue. It looks like some of the nested item input fields are missing it:

- Create pages items
- Date formats items
- Footer menu items
- Frontpage cards items
- Map layers items
- Map options: weight, opacity and fill opacity
- Main menu items

brockfanning:
@LucyGwilliamAdmin My latest commits here should fix the aria-describedby issues you listed, and also the missing space.

LucyGwilliamAdmin:
@brockfanning great 👍 

Here's some more info about the headings/hierarchies from DAC:

A good resource on sematic mark-up is this one WebAIM: Semantic Structure: Regions, Headings, and Lists. Screen reader users rely on semantic mark-up to present the relationships and purpose of the information on the page. For example, heading mark-up allows screen reader users to identify the purpose of text as a heading, it also contextualises the section of content and is also a useful navigational tool as users are able to filter the page by headings and then navigate straight to the section they require. Heading levels shouldn’t be used for visual styling – they should represent hierarchical and structural relationships on the page.

In terms of this page: http://sdgdev-813006012.eu-west-1.elb.amazonaws.com/release-prep-and-config-forms/config/
image

It doesn’t make sense to jump from a `<h1>` to a `<h3>` because a `<h3>` is semantically a second level subheading and should belong under a `<h2>`. All the current `<h3>` introducing a section with a grey border should therefore be `<h2>` as they belong directly ‘under’ the main page `<h1>` heading.   

However, the headings which sit underneath these sections, should be a `<h3>` as this demonstrates that these belong to the greater section. For example:

image

brockfanning:
@LucyGwilliamAdmin My response regarding the headers: I understand that an H1->H2->H3-etc hierarchy is ideal for headings. However my question is, what if the page does not use headings at all? Does that make it an accessibility failure? If so, AAA, AA, or A?

brockfanning:
@LucyGwilliamAdmin Actually if you haven't already sent that to DAC, I've got a lead on a possible way to get the hierarchical headings working. I'm looking into that again.

brockfanning:
@LucyGwilliamAdmin An update - we can definitely get hierarchical headings... eventually. However it depends on a change to a third-party library ([here is my PR in that project](https://github.com/json-editor/json-editor/pull/882)). If it is urgent, we can always fork that project and get the feature immediately. Otherwise we could wait and see if the feature gets accepted. So the options are:
1. Merge this PR as-is, and there will be no headings. When/if the json-editor project accepts my other PR, update this code to get hierarchical headings.
2. Fork the json-editor project and add that feature to our fork, then update this code to use our fork and get hierarchical headings.

I lean towards the first option for now, if that is OK with you, because it would be good to avoid forking another project if we can avoid it.

LucyGwilliamAdmin:
@brockfanning I agree - I think let's do first option for now and I think we can get this in to 1.2 as is for now (subject to testing by us, not DAC - think we can leave going back to DAC on this one for now)

LucyGwilliamAdmin:
@brockfanning I was just about to approve but have just checked and your PR in json-editor has been approved.

Do you want to get this into 1.2-beta as is or update the code to get hierarchal headings?

brockfanning:
@LucyGwilliamAdmin Well, the PR has been approved by 1 reviewer, but they require 2 reviews, and then it will need to be merged and included in a release. So it still may be some time before it's ready to use.

LucyGwilliamAdmin:
Ah - I see, will approve now

LucyGwilliamAdmin (approved):

PR #1004, opened 2020-12-05
Search box awareness

### LINKED ISSUE #996: Increase search box awareness

brockfanning:
Fixes #996

### DIFF
--- a/_sass/layouts/_header.scss
+++ b/_sass/layouts/_header.scss
@@ -177,16 +177,19 @@ body.contrast-high {
       transition: width 0.4s;
       margin-top: 5px;
       font-size: $searchBar-input-fontSize;
+      box-shadow: $text-color 0px 0px 0px 1px;
     }
 
     button {
       height: 34px;
-      color: $color-dark;
+      color: $button-foregroundColor;
       position: absolute;
       right: 0px;
       bottom: 0px;
       font-size: 1.40em;
       line-height: 15px;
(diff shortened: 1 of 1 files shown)


brockfanning:
@phillipgardiner Since the recommendation is to "potentially" change the magnifying glass to the word "Search" I will leave that for last. So here are some screenshots of everything else:

search-box-awareness-1
search-box-awareness-2
search-box-awareness-3
search-box-awareness-4

brockfanning:
I'm going to move these changes in with another related PR.

phillipgardiner:
@brockfanning Thanks for these. They look good. My question (as not able to currently see it it action) is how the yellow boxing that would appear around the search would work? As the form filed and button are all in one, would it highlight the whole thing as a single element? My expectation would be that you tab onto the form field (and that highlighted) and then tab again on to the magnifying glass button (which is then high lighted). Like this example: https://www.gov.uk/.

PR #1005, opened 2020-12-06
Support for goal content

### LINKED ISSUE #631: Add configurable text for goal page

brockfanning:
Fixes #631 

## Dependencies

This depends on a PR on jekyll-open-sdg-plugins, so needs this in the Gemfile for the site repository:

```
gem 'jekyll-open-sdg-plugins', git: 'https://github.com/brockfanning/jekyll-open-sdg-plugins.git', branch: 'goal-content-configuration'
```

## Usage

Add to the "create_goals" site configuration, like so:

```
create_goals:
    layout: goal
    goals:
      - content: My goal content for goal 1
      - content: "My goal content for goal 2, with [a link](https://example.com)"
      - content: custom.my_translation_key
```

Feature branch: http://sdgdev-813006012.eu-west-1.elb.amazonaws.com/goal-content

### DIFF
--- a/_includes/components/goal/goal-content.html
+++ b/_includes/components/goal/goal-content.html
@@ -1,0 +1,8 @@
+{% assign goal_page_content_raw = include.content | strip %}
+{% if goal_page_content_raw != '' %}
+<div class="row" id="page-content">
+  <div class="col-xs-12">
+    {{ goal_page_content_raw | t | markdownify }}
+  </div>
+</div>
+{% endif %}
--- a/_layouts/goal-by-target-vertical.html
+++ b/_layouts/goal-by-target-vertical.html
@@ -22,7 +22,7 @@
 
 <div id="main-content" class="container" role="main">
 
-  {{ content }}
+  {% include components/goal/goal-content.html content=content %}
 
   <div class="container">
     <h2>{{ page.t.general.targets_and_indicators }}</h2>
--- a/_layouts/goal-by-target.html
+++ b/_layouts/goal-by-target.html
@@ -22,7 +22,7 @@
 
 <div id="main-content" class="container goal-indicators goal-{{ page.goal.number }} goal-by-target" role="main">
 
-  {{ content }}
+  {% include components/goal/goal-content.html content=content %}
 
   <div class="visible-md-block visible-lg-block">
     <div class="col-md-6">
--- a/_layouts/goal.html
+++ b/_layouts/goal.html
@@ -22,7 +22,7 @@
 
 <div id="main-content" class="container goal-indicators goal-{{ page.goal.number }}" role="main">
 
-  {{ content }}
+  {% include components/goal/goal-content.html content=content %}
 
   {% assign goal_indicators = page.indicators | where: 'goal_number', page.goal.number %}
   {% for indicator in goal_indicators %}
--- a/docs/configuration.md
+++ b/docs/configuration.md
@@ -109,11 +109,16 @@ create_config_forms:
 
 ### create_goals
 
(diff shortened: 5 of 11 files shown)


LucyGwilliamAdmin:
I tried making a feature branch for this:
http://sdgdev-813006012.eu-west-1.elb.amazonaws.com/goal-content/1/

However I'm having an issue getting the feature branch to rebuild now that I have tried to add custom translation for goal 3:
https://github.com/ONSdigital/sdg-indicators/runs/1558509299?check_suite_focus=true#step:4:124

brockfanning:
@LucyGwilliamAdmin Oops, I suspect the one with the Markdown link will need to be in quotes, because it has a colon. I'll update the description of the PR.

LucyGwilliamAdmin:
@brockfanning just changed that but still getting same error

It seemed fine until I tried to use feature branch with translation for goal 3 text - i'll revisit in case I've made a mistake along the way

brockfanning:
I'm not sure what's going on - I'm able to build and run your branches locally without errors, and the goal content is showing up.

screenshot-127 0 0 1_4000-2020 12 15-16_23_05

LucyGwilliamAdmin:
@brockfanning turned out my feature branch just wasn't updating because feature branch server was down. It's back up now and works fine:
http://sdgdev-813006012.eu-west-1.elb.amazonaws.com/goal-content/3/

LucyGwilliamAdmin (approved):

phillipgardiner:
@brockfanning @LucyGwilliamAdmin Sorry just butting in on this one. But we don't need the yellow lines across the page. We can achieve separation through the use of appropriate white space (see mock up below). The start of the content can move up closer towards the goal icon and then a gap of white space before the 'Targets and Indicators' title should suffice. Thanks

Screenshot (73)

brockfanning:
@LucyGwilliamAdmin @phillipgardiner I agree with the suggestion but I'm wondering if this means we should also change indicator content. Those yellow lines were copied over from those styles (or maybe they just inherited them). For example: https://onsdigital.github.io/sdg-indicators/1-1-1/

I guess our options are:
1. Goal content and indicator content can be styled differently - so we make the suggested changes on goal content only
2. Goal content and indicator content should be styled the same - so we make the suggested changes on goal content and also re-visit indicator content
3. Goal content and indicator content should be styled the same - so we leave it as-is

What are your thoughts? I'm thinking that if we decide on the first option, we should make this change before 1.2.0. But if we decide on the second option, we could conceivably wait until 1.3.0.

phillipgardiner:
@brockfanning @LucyGwilliamAdmin Wow I can't believe I've never noticed those lines before,  but now i'm aware of them!! My view is although the goal and indicator content do not have to be the same, in this instance I would be looking to remove the lines from the indicator page also (because it would just look better).

We should therefore be aiming for option 2. However we work in an iterative world, so I personally would not have a problem making the goal page correct now (e.g. 1.2)  and then updating the indicator page later (1.3). Ultimately though its not my shout how its delivered and when but option 2 would be the vision we would be looking for at some point in the future.

phillipgardiner:
@brockfanning @LucyGwilliamAdmin umm on further consideration may be option 1 is best? I firmly believe that is how the goal page should look. However I have not had time to consider the effect this may have on the indicator page and after just having a look at a number of indicators removing the lines could have a knock on effect to other things (as just noticed these parapgraphs also have titles). So I would say get goals page correct (in 1.2) and look to improve indicator page in the future (once had time to consider all the impacts removing the lines might have). They do not have to look the same.

PR #1007, opened 2020-12-06
Pin to earlier version of Cucumber for now

brockfanning:
All new PRs are failing because of something that's going on in the latest version of Cucumber, so we need to pin to an old version for now and upgrade when possible.

### DIFF
--- a/tests/package.json
+++ b/tests/package.json
@@ -3,7 +3,7 @@
     "version": "1.0.0",
     "description": "Integration testing for Open SDG",
     "dependencies": {
-        "cucumber": "latest",
+        "cucumber": "5.0.2",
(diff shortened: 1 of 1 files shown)


brockfanning:
@jwestw @LucyGwilliamAdmin This is apparently needed because of some recent developments in the Cucumber project, which are causing all new PRs to file testing. Please approve when able. This is separate from the functioning of the platform, so there is no need to do a feature branch or any manual testing.

brockfanning:
Actually this appears to no longer be necessary - I guess the Cucumber folks fixed whatever it was. This is probably still a good idea though, just to avoid future issues like this.

LucyGwilliamAdmin (approved):

PR #1009, opened 2020-12-07
Bottom margin on frontpage goal tiles

brockfanning:
This fixes a slight regression bug, and would be good to get merged before 1.2.0. The bug only affects the "frontpage" layout (no longer used by the starter repos which are now using the "frontpage-alt" layout, but still used by multiple countries). The margin beneath the goal tiles on the homepage was missing. Here are before and after screenshots:

Before:

frontpage-goal-tile-margin-bottom-before

After:

frontpage-goal-tile-margin-bottom-after

### DIFF
--- a/_sass/layouts/_frontpage.scss
+++ b/_sass/layouts/_frontpage.scss
@@ -1,0 +1,9 @@
+.layout-frontpage {
+    .goal-tiles {
+        .row {
+            & > div {
+                margin-bottom: 10px;
+            }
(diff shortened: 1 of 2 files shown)


LucyGwilliamAdmin (approved):
This is fixed:
http://sdgdev-813006012.eu-west-1.elb.amazonaws.com/frontpage-goal-tiles-margin-bottom/

PR #1012, opened 2020-12-08
Allow click around tab link to trigger tab

brockfanning:
We recently changed the styling of tabs so that only the text gets highlighted. Unfortunately this means that the rest of the tab no longer has the click behavior. This tweak is intended to fix that.

We already had a "tabs.js" file but it was only being loaded if "accessible_tabs" was on. So this PR renames that file to accessibleTabs.js and puts the new behavior in tabs.js.

### DIFF
--- a/_includes/assets/js/accessibleTabs.js
+++ b/_includes/assets/js/accessibleTabs.js
@@ -1,0 +1,97 @@
+$(document).ready(function() {
+    $('.nav-tabs').each(function() {
+        var tabsList = $(this);
+        var tabs = tabsList.find('li > a');
+        var panes = tabsList.parent().find('.tab-pane');
+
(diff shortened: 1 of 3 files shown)


brockfanning:
@jwestw @LucyGwilliamAdmin This is another that I think we need to do before 1.2.0.

jwestw:
Feature branch: http://sdgdev-813006012.eu-west-1.elb.amazonaws.com/mobile-tab-click-area/

jwestw:
It's very difficult to show screenshots of this but the change I am noticing is that on the original/live site only the text in the tab is clickable, but in the new version the whole grey area of the tab is clickable.

jwestw (approved):
Happy to approve according to the previous comments

PR #1015, opened 2020-12-09
Fix tab clicking when accessible_tabs is false

brockfanning:
The recently-merged tab fix (#1012) unfortunately broke tab functionality for sites with accessible_tabs set to false. This should fix that.

### DIFF
--- a/_includes/assets/js/tabs.js
+++ b/_includes/assets/js/tabs.js
@@ -4,8 +4,9 @@ $(document).ready(function() {
 
         // Allow clicking on the <li> to trigger tab click.
         tabsList.find('li').click(function(event) {
-            $(event.target).find('> a').click();
-            event.stopPropagation();
+            if (event.target.tagName === 'LI') {
+                $(event.target).find('> a').click();
+            }
         });
     });
 });


brockfanning:
@jwestw @LucyGwilliamAdmin Apologies, my last fix actually broke stuff. For this maybe it would be helpful to have two feature branches - one with "accessible_tabs" set to true, and one with it set to false.

jwestw:
Feature branch: http://sdgdev-813006012.eu-west-1.elb.amazonaws.com/mobile-tab-click-area-fix

jwestw:
I have checked this on Chrome and it behaves as expected. 

I also checked on IE. It also works on both branches. But I also notice that the graph is not displaying. Does anybody know why that is?

image

This could be due to any combination of Windows OS, IE, and ONS security restrictions, or it could be a problem with the platform.

jwestw (approved):
I am happy to approve according to the previous comments. Seems to behave as expected.

brockfanning:
@jwestw Thanks. We did break IE support in recent code, but there is a fix for that in #1003.

PR #1021, opened 2020-12-12
Series helper function bug fix

brockfanning:
This helper function is used in indicatorModel.js, but is not actually present in the "helpers" object. So this is a bug - a bit hard to test because of all the conditions to satisfy - but clearly in need of fixing.

### DIFF
--- a/_includes/assets/js/model/helpers.js
+++ b/_includes/assets/js/model/helpers.js
@@ -66,6 +66,7 @@
     dataHasGeoCodes: dataHasGeoCodes,
     dataHasSerieses: dataHasSerieses,
     getFirstUnitInData: getFirstUnitInData,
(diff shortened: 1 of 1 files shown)


brockfanning:
@jwestw @LucyGwilliamAdmin This bug is easy to see in the code, but difficult to test. I'd recommend trying to get this merged before 1.2.0, if possible. One way to test would be to try with a site that does not use Series, and then try with a site that does use Series. However it's very straightforward, so we could also just merge without testing.

jwestw:
Hey @brockfanning , since this is so difficult to test I will not make a feature branch or go through any testing. I see your change in the code and unfortunately I do not understand the implications of the change so this will be merged purely on your recommendation.

jwestw (approved):
Approving based on last comment.

PR #1025, opened 2020-12-14
Changing series should affect the visible units/fields

brockfanning:
## Background for this PR

The intended use of Series and Units together, I think, was to have the same Units available for each Series. For example something like this:

Each series has each unit:

| Year | Series | Units | Value |
| --- | --- | --- | --- |
| 2008 | SeriesA | UnitA | 10 |
| 2008 | SeriesA | UnitB | 100 |
| 2008 | SeriesB | UnitA | 20 |
| 2008 | SeriesB | UnitB | 200 |

However, in practice, I have seen that the SDG data from the UN often has the Series and Unit together. For example 2.1.1 has two series: "Prevalence of undernourishment" (a percentage) and "Number of undernourished people" (a total). This kind of scenario, using the same dummy data as above, would look more like:

Each series has its own unit:

| Year | Series | Units | Value |
| --- | --- | --- | --- |
| 2008 | SeriesA | UnitA | 10 |
| 2008 | SeriesB | UnitB | 200 |

In this case, it doesn't make sense to display UnitB when the user is on SeriesA, and likewise it doesn't make sense to display UnitA when the user is on SeriesB.

The same can be true of disaggregations. One series may have disaggregations that the other series don't.

So far we have a solved this with our "No data available" message. With this, the user makes a selection change -- for example changing the unit -- and then sees that message. We have been assuming that the user would then revert that change -- ie, go back to the original unit. But this doesn't help with the "each series has its own unit" case above. Here is what the user might do:

1. Load the page and see data for SeriesA / UnitA.
2. Try changing to UnitB - see the "No data message"
3. Go back to UnitA to get data back.
4. Try changing to SeriesB - see the "No data message"
5. Go back to SeriesA to get data back.
6. Assume there is no SeriesB / UnitB data, move on to another indicator.

Here's an example of what that would look like on indicator 2.1.1:

chrome-capture (1)

Meanwhile what is really required is for the user to change both:

1. Load the page and see data for SeriesA / UnitA.
2. Try changing to UnitB - see the "No data message"
3. Try changing to SeriesB - see the SeriesB / UnitB data.

But this kind-of expects the user to predict the future, or at least to experiment with the selections.

Here's an example of the user doing that, on indicator 2.1.1:

chrome-capture (2)

## Purpose of this PR

With this PR, changing to a new Series becomes similar to loading a completely new indicator. All selections are reset and visible units and disaggregation are reset. To again look at the "each series has its own unit" scenario above, here is what the user would do/see:

1. Load the page and see data for SeriesA / UnitA.
2. There is a SeriesB option, but there is NOT a UnitB option.
3. The user naturally changes to SeriesB.
4. The user now sees data for SeriesB / UnitB.
5. Now there is NOT a UnitA option anymore, instead there is a now only a UnitB option (which is already selected).

Here's an example of what that looks like, on indicator 2.1.1:

chrome-capture

## Important to discuss

The most impactful part of this PR is the way that the series column is treated. With this PR, changing to a new series is similar to loading a completely new indicator page, in that all current disaggregation/unit selection are cleared and reset.

This may seem objectionable and should be discussed. Obviously since I'm submitting the PR I'm of the mind that it is correct, so here is my reasoning:

In the SDGs the existence of the "series" concept is problematic. It would have been better if the SDGs only had goals, targets, and indicators - but instead they have goals, targets, indicators, and series. In order to handle the series concept in Open SDG, we had 2 choices:
1. Add a series toggle on the indicator page
2. Create a new type of page: the "series" page

We went with the first option -- the series toggle -- which I think was the correct choice. However we didn't take the next step of making the series toggle behave like reloading the page. Instead we have a series toggle that maintains the current unit/disaggregation selections, even if they don't exist in the new series. I have been noticing that, in practice, this often means the user sees the "no data" message.

So my opinion is that it would be best to always reset everything when the series is changed. This gives the page a consistency of behavior, and avoids the frustration of seeing the "no data" message.

## Still to do

* If it's agreed on to continue with this, we should add a screenreader announcement that the chart/table selections have been reset whenever the series changes.

## Feature branches
UK: http://sdgdev-813006012.eu-west-1.elb.amazonaws.com/series-unit-field-revamp/15-2-1/
Kyrgyzstan/SDMX sandpit: https://open-sdg-sdmx-input-sandpit.github.io/site/9-1-2/


LucyGwilliamAdmin:
@brockfanning I'm just looking through PRs to see who will test out of me and James - can you let me know what we're looking for with this one please?

brockfanning:
@LucyGwilliamAdmin Just added some notes to the PR description.

brockfanning:
@LucyGwilliamAdmin In case you were already testing - I noticed a bug in the code and fixed. Also I've added more discussion in the notes. This seems like a good one to go over during a call.

LucyGwilliamAdmin:
@brockfanning Thanks for the notes - I agree this is one to discuss. However most of the team are now off until beginning of next year so think one to bring up then.

I can already think of where this could be tested though - Kyrgyzstan have a couple of indicators which have the issue of Series and Units being joined

LucyGwilliamAdmin:
I've began looking at this and have added some feature branches to description. Uk I think looks much better and is much easier to navigate. I'm also speaking to someone on data team who looked at other version.

I'm not too sure what's happening with Kyrgzystan/SDMX sandpit as functionality doesn't seem to be working in same way - just going to check I've updates everything correctly

LucyGwilliamAdmin:
@brockfanning I've just checked on this one and the Kyrgyzstan/SDMX sandpit is working as expected:
https://open-sdg-sdmx-input-sandpit.github.io/site/9-1-2/

@phillipgardiner is this one you're interested in? There's a link to a UK feature branch in the description above

LucyGwilliamAdmin:
@phillipgardiner sorry not too sure what the thumbs up means - did you want to take a look before it's approved?

phillipgardiner:
@LucyGwilliamAdmin I can't remember now either!! Think I was just saying I had a look and it looked good. No need for me to review.

LucyGwilliamAdmin (approved):
Looks good!

PR #1027, opened 2020-12-15
Alternate unit and series columns

brockfanning:
This allows for the unit and/or series columns to be customized instead of the usual "Units" and "Series".

This is mainly useful for translation purposes. When using an SDMX DSD for translation, the translations of units of measurement are in the UNIT_MEASURE group, and the translations of series are in the SERIES group. Having columns with the same name means that they will automatically be translated.

Historically the the unit column as been "Units" but that was arbitrarily chosen. The same for "Series", which was recently added but also arbitrarily chosen. In hindsight, we should have been using "UNIT_MEASURE" and "SERIES" just to match the global standard (SDMX). This PR makes this possible.

To use this, this would be the config:

```
data_fields:
  units: UNIT_MEASURE
  series: SERIES
```


LucyGwilliamAdmin (approved):
Looks good to me - used SDMX file where UNIT_MEASURE and SERIES are used and then added the `data_fields` config and the Units and Series are showing correctly.

PR #1036, opened 2020-12-26
Optimization of validParentsByChild

brockfanning:
The `validParentsByChild` function can be a bottleneck in large datasets with a lot of disaggregation values, because of all the loops and filtering. This is a minor optimization of it that should help some. This moves some calculations outside of a loop to reduce the number of times it has to run.

Here is a before/after comparison using UK's 4.2.1:

## Running the code from 1.3.0-dev

valid-parents-by-child-current

Total: 959.3ms

## Running the code from this PR

valid-parents-by-child-optimized

Total: 521.9ms

So this optimization saves a little less than half a second. Not a huge improvement, but at least something. It should be noted that 521.9ms is still a _lot_ for a javascript function, so there is still room for improvement (like seeing if this function can be avoided altogether in some way).

## Testing advice

Testing should focus on making sure that:
1. Indicators with lots of data and disaggregation load a bit faster
2. Indicators with normal amounts of data/disaggregation load the same, if not faster
3. In all cases, the parent/child functionality works as expected ("Available when 'Region' is selected", etc.)

JW edit: [Feature branch](http://sdgdev-813006012.eu-west-1.elb.amazonaws.com/valid-parents-by-child-optimization)

### DIFF
--- a/_includes/assets/js/model/fieldHelpers.js
+++ b/_includes/assets/js/model/fieldHelpers.js
@@ -343,6 +343,9 @@ function selectMinimumStartingFields(rows, selectableFieldNames, selectedUnit) {
  * @param {Array} fieldItemStates
  * @param {Array} rows
  * @return {Object} Arrays of parents keyed to children
+ *
+ * @TODO: This function can be a bottleneck in large datasets with a lot of
+ * disaggregation values. Can this be further optimized?
  */
 function validParentsByChild(edges, fieldItemStates, rows) {
   var parentFields = getParentFieldNames(edges);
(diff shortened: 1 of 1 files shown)


jwestw:
@brockfanning and I have spoken about how this could be tested. 

I am creating a feature branch and will try to test the loading time for 4.2.1 which is the indicator with the biggest dataset.

Feature branch:

brockfanning:
@jwestw Here are the steps I take to get those numbers:

1. Open Chrome and go to an indicator with a lot of disaggregation, like: https://onsdigital.github.io/sdg-indicators/4-2-1/
2. Press CTRL-SHIFT-j to open the console
3. Where it says Elements, Console, Sources, etc. select "Performance"
4. Press CTRL-SHIFT-e to run a performance profile
5. Where it says Summary, Bottom-Up, etc. select "Bottom-Up"
6. In the filter type: validParents
7. You should see one result
8. Click on the link that starts sdg.js?v=etc...
9. You should now see the code and numbers next to lines

jwestw:
**Performance results from https://sdgdata.gov.uk/4-2-1/**

image

validparentsByChild took 2.32s to execute. 


**Performance results from [feature branch](url)**
image

jwestw:
**Results from 3 page loads on each branch:**


image

jwestw (approved):
Approving according to the speed tests (and the fact the platform is still running smoothly)

PR #1037, opened 2020-12-26
Different approach for sorting field hierarchy

### LINKED ISSUE #390: Edges issue
Looking at this [feature branch](https://sdg.mango-solutions.com/3-3-2-deploy/3-3-2/), you can see that there's still a problem with the edges.

First of all it says Ethnic Group is 'Available when 'Country' is selected' but looks as if it actually dependent on 'Country of birth'

image

image

image

image

brockfanning:
Fixes #390 

This is a different approach to a recent attempt at sorting the fields with hierarchy. This seems to work well for me, but testing is needed across a wide variety of indicators, with and without disaggregation hierarchy.

The UK indicator that should be improved by this is 4.2.1. However as mentioned above it should be tested on other indicators also.

JW EDIT: [Feature Branch](http://sdgdev-813006012.eu-west-1.elb.amazonaws.com/fix-edges-hierarchy-display/)


jwestw:
This seems to be working for 4.2.1. The disagregations are in the right order and are not clickable until their parent is clicked. 

image

Lucy and I think it would be best if the data team could have a look at this one. 

Here is the [feature branch ](http://sdgdev-813006012.eu-west-1.elb.amazonaws.com/fix-edges-hierarchy-display/)for further testing.

jwestw:
@brockfanning @LucyGwilliamAdmin Our data team have agreed to do some testing on this, which will be more comprehensive because they know the indicators they look after. Should be finished by next week (~28th Jan).

ANikolova22:
Hey all,
I've had a check through Goals 17 and 4 on the feature branch where the fix has been applied. Generally the display (where relevant) seems to be working fine, but I noticed an issue with indicator 4-1-1. The options appear correctly arranged, but there seems to be a problem when choosing different time series options (the radio buttons at the top). For example option “Attainment at age 16” doesn’t work at all, may be to do with the fact that that one has fewer disaggregation options? 

4-5-1 seems to have an issue with displaying data for some of the time series options, but I think this is a separate issue not related to the current matter.

Atanaska

LucyGwilliamAdmin:
@ANikolova22 I have a feeling that the issue with 4.1.1 may be to do with that the Units options are actually Series. I have quickly put together this feature branch where I have rename the Units column as Series, and 4.1.1 now behaves quite differently:

http://sdgdev-813006012.eu-west-1.elb.amazonaws.com/fix-edges-hierarchy-display-with-data-changes/4-1-1/

Although looks like this brings its own issues...

Could try something similar with 4.5.1 to see what effect that has - the data feature branch i used is called `series-test-1` if you want to change column name there.

@brockfanning @jwestw I'm also having some other thoughts:
- do data team first need to make sure Units and Series are being used in way that is expected by platform?
- will #1025 affect results of this PR?

brockfanning:
@LucyGwilliamAdmin I'd like to first resolve the issue that @ANikolova22 caught on 4.1.1. That seems like it's likely a direct result of this PR. I'll look into that now and then respond to your other questions based on what I find.

ANikolova22:
Hmm, renaming the column seems to have fixed that one issue, but on series-test-1 branch the hierarchy order is not right, and for both series-test-1 and fix-edges-hierarchy-display-with-data-changes the option where Maths from 'Subjects' category is selected by default when you land on the page doesn't seem to be enforced...

brockfanning:
@jwestw I've just pushed some commits that I think will fix the issue with 4.1.1 that @ANikolova22 discovered. Would you be able to refresh the feature branch?

@LucyGwilliamAdmin My first instinct is that this PR can be independent of the other considerations - this PR is mainly about sorting, so I think it can be looked at separately.

jwestw:
I am rebuilding the [feature branch](http://sdgdev-813006012.eu-west-1.elb.amazonaws.com/fix-edges-hierarchy-display/) so it includes the latest code that @brock has pushed.

LucyGwilliamAdmin:
Also just updated this branch: http://sdgdev-813006012.eu-west-1.elb.amazonaws.com/fix-edges-hierarchy-display-with-data-changes/4-1-1/

There are still changes between whether Units or Series is used so might be good to look at both

brockfanning:
Thanks all! @ANikolova22 I believe that issue on 4.1.1 should be fixed now.

ANikolova22:
Thanks all! James' rebuilt feature branch seems to have solved the issue, even though that column is still called Units. Lucy's version where the Units are renamed to Series would also work fine if the Subject default option is set to "Maths", rather than "English" , but that's an aesthetic matter I think. Still, as Lucy mentioned earlier we (the data team) need to be aware if in this context that column should indeed be called "Series" rather than "Units" and make necessary checks/corrections for other indicators.

LucyGwilliamAdmin:
@ANikolova22 ok, lets leave my feature branch for now - there's a separate PR for series stuff

LucyGwilliamAdmin:
Just noticed that 2-1-1 isn't showing any data on this branch:
http://sdgdev-813006012.eu-west-1.elb.amazonaws.com/fix-edges-hierarchy-display/2-1-1/

ANikolova22:
Indeed! Sorry, I haven't checked through Goal 2. Not sure what happened there!

brockfanning:
Good catch @LucyGwilliamAdmin - apologies all for the trial and error. @jwestw I've made another fix so this is ready for a refresh.

jwestw:
The feature branch has been refreshed.

varunjakki:
No issues with Goal 3 and 8

jwestw:
According to feedback from the data team, there are no hierarchy issues in any of the indicators.

jwestw (approved):
Approving according to last comment.

ANikolova22:
Hi all, given the fix was merged, is it expected to have taken effect on the live NRP by now? Just noticed that the disaggregation hierarchy for 8.5.1 doesn't seem quite right (on the live NRP, it was fine on the feature branch we reviewed).

PR #1038, opened 2020-12-27
Show units in chart tooltips

### LINKED ISSUE #997: Add unit of measure to chart tooltip
Some users are uncertain of the relevance of the number displayed in tooltip

Add unit of measure to chart tooltip

To allow users to be sure of the unit of measure of number shown in tooltip

brockfanning:
Fixes #997 

Testing advice: no configuration needed, this should automatically show the unit of measurement in the tooltips.


jwestw:
The current SDG UK website:

image


The[ Feature Branch](http://sdgdev-813006012.eu-west-1.elb.amazonaws.com/):

image

Looks good to me.

LucyGwilliamAdmin:
I kind of agree with James about the Units type but also aware of accessibility issues and what DAC might say - think they would def want Units type at least read out by screenreader to give context. That being said if screen reader users would want the context then maybe so would visual users. 

I think my main reasoning behind agreeing with James is that it looks kind of bulky so maybe Unit: would be better but can see advantage of using Units type as it's already been translated.

I know I've given no definite thoughts here but want to do a little digging to see what, if anything, the ONS do in these cases, or even other sites.

LucyGwilliamAdmin:
There's this approach:
https://www.ons.gov.uk/employmentandlabourmarket/peopleinwork/employmentandemployeetypes/bulletins/uklabourmarket/latest#earnings-growth

i.e. [DISAGG] ([UNIT TYPE]): [VALUE] but this seem like a lot of unnecessary repetitiveness as would only ever have one unit type on graph at one time

LucyGwilliamAdmin:
Also here's another screenshot with a longer units type just in case of interest: http://sdgdev-813006012.eu-west-1.elb.amazonaws.com/units-in-tooltips/11-3-1/

image

I'm thinking let's go with either with or without `Units type:`

@phillipgardiner do you have any opinion from user design pov?

phillipgardiner:
@jwestw @LucyGwilliamAdmin @brockfanning This is not based on any hard evidence just my UX view. I would go without the units type. My reasoning for this is that the user is selecting the units type before hand so this coupled with the fact the units type is displayed on the 'x' axis of the graph, means it should be clear what the unit type is to them.

phillipgardiner:
@LucyGwilliamAdmin @jwestw @brockfanning I guess my original expectation is that it would look a little different. I was expecting the units to appear after the number (e.g. UK, 10.9 (%), or UK, 10 (£) but I suppose this format does work for graphs with long unit types. 

I think I have a couple of questions:
1) Instead of saying 'Units type' could we just say 'Unit'. For example Unit: £'s - Would this work for all unit types?
2) Would it be possible to wrap text on chart tips? This would stop those with long unit types from creating a very large tool tip and this blocking a large part of the graph
3) Could we adjust the transparency of the tool tip. That way, the chart will still be visible underneath - Obviously staying within accessibility guidelines

Depending on the answers above, my view would be to keep the word unit. As it will make more sense for tool tips where the unit is a simple one (e.g. Unit: £'s) just having £'s floating about near the bottom would be a bit weird.

brockfanning:
@phillipgardiner 
1. I think "Unit" should work fine for all unit types. It does need to be translated but that's OK.
2. Wrapping text is possible, though it's a bit complicated because of the way those tooltips are "canvas" instead of HTML. But I think it should be feasible if we pick a maximum number of pixels - something that would work for both mobile and desktop.
3. The background color can be controlled, yes. Right now it's set at 0.8 opacity.

phillipgardiner:
@brockfanning happy to feed into points 2 and 3 but I think there ones where we'll need to take a look and tweak in a live environment kind of thing until it looks rights

jwestw:
@brockfanning please tell me when you need me to refresh the feature branch.

brockfanning:
Let's try to do the opacity change with screenshots, if we're going to be "turning the dials".

But first a question: my impression is that item 2 and 3 are two different ways to deal with the same problem - which is that sometimes the tooltip can obscure large parts of the page. I do have one reservation about the line-wrapping: even if we do line-wrapping for units, there will still be a chance that long disaggregation names will cause the same issue. So I feel like we would eventually need to do the same for disaggregation names. And given that the line-wrapping here is technically complicated, I would lean towards option 3 only, if that resolves the problem, since that is a simple changing of a number. Would it be feasible to do option 3 and not do option 2?

Finally a last note - changing the opacity of the tooltips background is really a separate thing from adding units to the tooltips, so it should probably be a separate ticket. There is a benefit to having them in separate tickets, which is that the "release notes" become more clear. We generally make the release notes from the titles of these pull-requests. So if we have them in separate tickets, the release notes could be something like:

* Show units in chart tooltips
* Transparent chart tooltip background

Whereas if we bundle the opacity change in this PR, then it would just say:

* Show units in chart tooltips

And there would be an undocumented opacity change in the release, that users would not know about unless they read through this thread.

phillipgardiner:
@brockfanning Yep fair point. Lets just do the transparency for now and then take a view on it. I guess by wrapping the text we would just be making the tool tip longer, thus in turn still blocking the graph more just downwards instead of sideways. So ultimately the transparency is more key. Totally agree about a separate ticket, I will create it :)

brockfanning:
@phillipgardiner @AnnCorp I meant to make a note here earlier, and now I'm not sure if I remember right. Did we decide on "Unit of measurement" in the disaggregation controls and in the chart footer, and "Unit" here in the chart tooltip? Or was it "Unit of measurement" across the board?

phillipgardiner:
@brockfanning We decided on "Unit of measurement" in the disaggregation controls and in the chart footer, and "Unit" in the chart tooltip.

brockfanning:
I merged in a translation of "Unit" for English: https://github.com/open-sdg/sdg-translations/pull/179

And I've updated this PR to use this new translation in tooltips, and also to switch to "Unit of Measurement" in the disaggregation controls.

@jwestw This should be ready for a refresh now. Though to see the new "Unit" translation, you'll need to be pointing at the 1.3.0-dev branch for sdg-translations (in the data config).

_includes/components/indicator/units-template.html
```diff
@@ -2,7 +2,7 @@
   <% if(units.length) { %>
     <h4>{{ page.t.indicator.units_type }}</h4>
```
LucyGwilliamAdmin, line 3: @brockfanning I'm just looking at a feature branch for Phill and think this needs to be changed from `units_type` to `unit _of_measurement`

jwestw:
Notes on what was done here. 

- Created a [data branch](http://sdgdev-813006012.eu-west-1.elb.amazonaws.com/data/units-in-tooltips-data/) ([repo](https://github.com/ONSdigital/sdg-data/tree/units-in-tooltips-data)) and a [feature branch](http://sdgdev-813006012.eu-west-1.elb.amazonaws.com/units-in-tooltips/)
- Updated the data_config so that the translations point to '1.3.0-dev' tag. (Though I couldn't find this in [tags](https://github.com/open-sdg/sdg-translations/tags).
- Tested on the feature branch and it looks like everything is working and the fixes are good.. 

image

jwestw (approved):
According to what I checked/tested detailed in the last comment, I am approving these changes.

phillipgardiner:
@brockfanning @jwestw looks good to me.

PR #1039, opened 2020-12-27
Do not change SVG images in high-contrast mode

brockfanning:
Because SVG images can use CSS styling, there is no need to change them for high-contrast. So this PR skips the image-changing behavior for SVG icons.

Also, this PR adds some useful HTML classes to all goal icons throughout the site. They aren't used anywhere, and won't affect anything, but may be useful in the future, or to any countries that would like to add special styling to goal images.

EDIT: [Feature branch](http://sdgdev-813006012.eu-west-1.elb.amazonaws.com/svg-icon-high-contrast/)

### DIFF
--- a/_includes/assets/js/accessibility.js
+++ b/_includes/assets/js/accessibility.js
@@ -107,6 +107,7 @@ var accessibilitySwitcher = function() {
 
 
   function imageFix(contrast) {
+    {% unless site.goal_image_extension == 'svg' %}
     if (contrast == 'high')  {
       _.each($('img:not([src*=high-contrast])'), function(goalImage){
         if ($(goalImage).attr('src').slice(0, 35) != "https://platform-cdn.sharethis.com/") {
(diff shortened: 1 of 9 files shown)


LucyGwilliamAdmin (approved):
Looks good to me - image source is same when in default contrast mode and high contrast mode:

http://sdgdev-813006012.eu-west-1.elb.amazonaws.com/svg-icon-high-contrast/

LucyGwilliamAdmin:
@brockfanning I've just realised this might have had some unintended side effects, when looking at the site in high contrast mode: if local images are in PNG and not SVG then they're also not swapped for high contrast version.

For example, the arrow on [this page](http://sdgdev-813006012.eu-west-1.elb.amazonaws.com/svg-icon-high-contrast/user-personas/) is now not being swapped for high contrast version and neither is OGL logo in footer

brockfanning:
@LucyGwilliamAdmin Understood - ok - do you mind creating a new issue for that and adding it to the 1.3 project?

LucyGwilliamAdmin:
Yes no probs- will do that now

PR #1042, opened 2020-12-28
HTML 'lang' attribute should depend on current language

brockfanning:
Right now our html `lang` attribute is hardcoded as "en". I'm not aware of any reason we would not want it to be dynamic, based on the current language.

[Feature Branch](http://sdgdev-813006012.eu-west-1.elb.amazonaws.com/dynamic-lang-att2)

### DIFF
--- a/_includes/head.html
+++ b/_includes/head.html
@@ -1,5 +1,5 @@
 <!DOCTYPE html>
-<html lang="en" class="no-js">
+<html lang="{{ page.language }}" class="no-js">
     <head>
         <meta charset="utf-8">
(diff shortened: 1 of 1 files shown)


LucyGwilliamAdmin:
@jwestw for this one a multilingual feature branch is required, and then to check if html lang element changes: open the inspect panel and the lang element should be right near the beginning:
image

jwestw:
Feature branch created: http://sdgdev-813006012.eu-west-1.elb.amazonaws.com/dynamic-lang-attribute

jwestw:
With the site on English:

image


With the site switched to Russian

image

Looks good to me!

jwestw (approved):
Happy to approve according to last comments.

PR #1043, opened 2020-12-29
Multlingual search tests

### LINKED ISSUE #590: Tests for multilingual search
For better test coverage we should add tests for running a search in a non-English language. The thing to test for would be that the "stemmer" works as would be expected. For example in Spanish a search for "proporciónes" should find results that match "proporción".

brockfanning:
Fixes #590 

Since this PR is only automated tests, no manual testing is needed. As long as the automated tests pass (which they do) then it should be OK to merge.

### DIFF
--- a/tests/features/Search.feature
+++ b/tests/features/Search.feature
@@ -48,3 +48,17 @@ Feature: Search
     And I follow "popul"
     And I wait 5 seconds
     Then I should see "results found"
+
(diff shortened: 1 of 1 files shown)


LucyGwilliamAdmin (approved):
Approving as @brockfanning says no manual testing required and makes sense to me

PR #1045, opened 2021-01-02
Use data-package schema to sort fields/values

### LINKED ISSUE #716: Starting unit can change depending on language
Something is causing the starting unit to change depending on the language. For example see [links removed, they no longer illustrate this issue].

### LINKED ISSUE #814: Sorting of sub-categories and table headers
The subcategories are sorted alphabetically by their translations. This sometimes causes a strange order in the selection options and table headers especially when they contain numerical values like here:

grafik

If I am not mistaken the sorting used to be on the basis of the values coming from the csv-files not on the translations. In this case it was possible to sort them by adding an a), b), c) at the beginning of the value.
Is there anything to be said against sorting again according to the values and not according to the translations?

### LINKED ISSUE #972: Ordering of disaggregation categories
1. Go here: http://sdgdev-813006012.eu-west-1.elb.amazonaws.com/datapreview_Frensisbras-patch-13/17-18-1/
2. Expand the Goal dropdown

It's expected that the categories would go Goal 1, Goal 2, Goal 3, ..., Goal 17 but that is not the case.

Instead it goes Goal 1, Goal 10, Goal 11, Goal 12, ..., Goal 17, Goal 2, Goal 3, ..., Goal 9

Think this also an issue for other disaggregations e.g. Age

Need to either be able to set the order or add in something about numeric ordering

So that disaggregations that use numeric values appear in the expected order

Considered adding A., B., C., before the values to ensure that they're in order but this can be quite messy and goes against standardised values so not really a good fix.

**Additional context**

### LINKED ISSUE #1173: Sorting disaggregation dimensions
It seems that the disaggregation dimensions are ordered alphabetical.
Is it possible to use other sorting criteria?

E.g. I would like to have "Geflüchtete und geduldete Personen insgesamt" as first dimension....

[Aufenthaltstitel](https://wohnungnds.github.io/WK_Site/11-1-1/)

image

brockfanning:
Fixes #1173
Fixes #972
Fixes #814
Fixes #716

### Order of disaggregations

A new way to control the sorting within the disaggregation controls is with a `datapackage` and `sorting` setting in the **data** (_not site_) configuration. The possible options are `alphabetical` and `default`. For example:

```
datapackage:
    sorting: alphabetical
```

With this configuration, disaggregation dropdowns will be sorted alphabetically by their _untranslated_ value. This is an intentional return to the pre-1.0.0 behavior. Since 1.0.0, the dropdowns have been sorted alphabetically by their _translated_ value, which has caused inconsistency and has forced users to resort to adding number (eg, "1. Urban", "2. Rural") in their CSV files.

However with a data configuration of...

```
datapackage:
    sorting: default
```

...the behavior will be very different. Disaggregation dropdowns will appear in the order they are in the source data. For example, if the header line of a CSV file is `Year,Foo,Bar,Value`, then the dropdown for "Foo" would appear above "Bar" in the sidebar.

### Order of disaggregation values

Similarly, with the "alphabetical" sorting, disaggregation values (the checkbox options within each dropdown) will be sorted alphabetically by their untranslated value.

And similarly, with the "default" sorting, values will appear in the order they show up in the rows of data. For example, if a CSV file starts out like:

Year|Location|Value
---|---|---
2020|Urban|100
2020|Rural|200

The dropdown for "Location" would show a checkbox for "Urban" and then a checkbox for "Rural".

## Configuration

A low-tech way to control the order, then, is to tweak the order of the fields/values in the source data. But more direct control is also possible, with configuration. This configuration would be in the form of [Table Schema](https://specs.frictionlessdata.io/table-schema/) files, in a folder in the data repository. This is a bit technical but I think there is a big benefit to using a standard like Table Schema. The details still need to be added in documentation to this PR, but in a nutshell, a schema file might look something like this, for an indicator with a "Location" column:

```
fields:
  - name: Location
    constraints:
      enum:
        - Urban
        - Rural
        - My other value which should appear last
```

## Breaking changes

Note that if the `datapackage` and `sorting` options are omitted from the data config, it will default to the "alphabetical" option. This seems counterintuitive (you'd naturally expect it to "default" to the "default" option...) but it is done this way to avoid a backwards compatibility change: users may be accustomed to the alphabetical ordering, so it should be kept. However when we go to the next major version (2.0.0) we could take the opportunity to make "default" the actual default.

## Refactoring

In the course of testing this PR, a bad inefficiency was found, and so the `getCombinationData` function was refactored (for a second time since 1.3.0) here as well.


brockfanning:
@LucyGwilliamAdmin @jwestw This one could be a good discussion topic during a tech call. I believe it's ready for review, though I still need to add automated tests and documentation.

LucyGwilliamAdmin:
@WohnungNDS  Yes, you will need to point at `1.4.0-dev` here:
https://github.com/WohnungNDS/IM_Data/blob/develop/scripts/requirements.txt

and then also point at `brockfanning/open-sdg@data-schema` here:
https://github.com/WohnungNDS/IM_Site/blob/develop/_config.yml#L30

WohnungNDS:
@LucyGwilliamAdmin I have changed the reqiuirements.txt and point at brockfanning/open-sdg@data-schema
but there is no visible effect.... the disaggregation dimensions still are in alphabetical order

LucyGwilliamAdmin:
@WohnungNDS I've just had a look at the changed files, and I think that `fields` code needs to be in a indicator specific yml file .

See this example: https://github.com/open-sdg/open-sdg/blob/afb3d196dbf83f0982e434b9af4fa9e95ddfd0ed/tests/data/data-schema/3-1-1.yml

@brockfanning is there a way to apply the order site wide?

LucyGwilliamAdmin:
@WohnungNDS Also, this needs to be added to config_data.yml:
```
data_schema:
  class: DataSchemaInputTableSchemaYaml
  source: data-schema/*.yml
```

brockfanning:
@LucyGwilliamAdmin If I understand right, WohnungNDS' goal is to have disaggregations sorted by the order that they appear in the CSV file. To get this behavior, the config_data.yml file should have this:

```
datapackage:
    sorting: default
```

I don't think that the `data_schema` setting or any extra schema files will be needed.

@WohnungNDS Can you give this a try or correct me if I'm misunderstanding?

This PR was designed to avoid a change of behavior - and since the sorting is already alphabetical now, this PR maintains that behavior unless it is specifically changed to `default` as shown above. (It's counterintuitive I know, sorry about that.)

WohnungNDS:
@brockfanning for now, that's right. 
Perhaps it make sense to have such furtcher sorting options in the future. May be someone is updating data and does not follow the former structure....

brockfanning:
@WohnungNDS That makes sense. Were you able to confirm that this is working as intended, with that datapackage "sorting" set to "default" as shown above?

WohnungNDS:
@brockfanning  sorting default works as expected

MoBosse:
Hi @brockfanning @WohnungNDS , I was a bit nosey and tried it out as well but it has no effect on my site.
I tried here: [site repo](https://github.com/MoBosse/testSite), [data repo](https://github.com/MoBosse/testData), [site](https://mobosse.github.io/testSite/5-5-1/).
Neither the _alphabetical_, nor the _default_ option changed anything on the appearence of the disaggregations.

Example: The federal state "Niedersachsen" has the translation key "lower saxony". If understood this PR right it should appear above "Mecklenbug-Vorpommern" here when "alphabetical" is selected right? 
grafik

Did I do anything wrong?

WohnungNDS:
@MoBosse 

did you also change your requirements.txt in your data repo as @LucyGwilliamAdmin mentioned here: 



> @WohnungNDS Yes, you will need to point at `1.4.0-dev` here:
> https://github.com/WohnungNDS/IM_Data/blob/develop/scripts/requirements.txt
> 
> and then also point at `brockfanning/open-sdg@data-schema` here:
> https://github.com/WohnungNDS/IM_Site/blob/develop/_config.yml#L30

brockfanning:
@MoBosse Thanks for testing - I don't think we had tried it with translated fields. I'll think on this a bit more.

brockfanning:
@MoBosse Actually if possible could you try re-building the site? I just pushed another commit to this branch which might help.

MoBosse:
@brockfanning yes, it works (at least the "alphabetic" version):
grafik
Thanks!

MoBosse:
"default" works perfectly aswell!
grafik


As an aside, is it intentional that the disaggregations under the chart are in the reverse order compared to the order in the dropdowns?
grafik

Actually, I would expect the order of the disaggregations that are read from top to bottom in the dropdown to reappear from left to right in the chart (and table header)....

brockfanning:
@MoBosse Good catch. I tracked that issue down to our old friend the `getCombinationData` function. And in doing so I noticed that it now has a crippling inefficiency, which is causing javascript to grind to halt when a lot of disaggregations are selected. (Such as when pressing "Select all" on a disaggregation with 10+ checkboxes.)

Rather than complicate things with a separate issue I've added yet another refactor of `getCombinationData` here. If you are able to test, I'm very interested in these questions:
1. Is the issue with the chart legend order resolved?
2. Are the issues that you originally pointed out in #1158 still resolved? I just wanted to make sure that my re-refactor doesn't cause a regression.

@LucyGwilliamAdmin It may also be good to test the UK data with regards to #1209 to also make sure that is still working as expected. Again, just wanting to make sure a regression isn't caused.

MoBosse:
@brockfanning I can confirm that for both the default and the alphabetical option 

- the order of the legend items is correct (from top to bottom in the dropdowns corresponds to from left to right in the chart and table head)
- the order of the single disaggregations within a legend item corresponds to the order in the dropdowns (so yes #1158 is solved)

LucyGwilliamAdmin:
@brockfanning just pointing at this branch and not configuring anything, our disaggregation structure is now longer as expected. See here:
https://lucygwilliamadmin.github.io/sdg-indicators/5-2-1/

Ethnic group under Highest qualification instead of under Ethnicity

brockfanning:
Thanks @LucyGwilliamAdmin, so still a bit to work on here before we release this.

LucyGwilliamAdmin:
@brockfanning just checked and the fix for #1209 still works

brockfanning:
@WohnungNDS @LucyGwilliamAdmin @MoBosse I've just pushed a small change to the way the parent/children disaggregations are sorted in the sidebar. The intention is that children will always immediately follow their parent in the sidebar. Would it be possible to do another round of testing to make sure that this is working as intended?

brockfanning:
@LucyGwilliamAdmin @WohnungNDS @MoBosse Sorry, I spoke too soon - there are failing tests that I need to resolve. Please hold off.

brockfanning:
@LucyGwilliamAdmin @WohnungNDS @MoBosse Ok, after another commit I see the tests are passing, so any feedback would be greatly appreciated.

WohnungNDS:
@brockfanning @MoBosse 
alphabetical and default working fine....
To me the question is, if it makes sense do adjust the sorting not site wide then for each indicator.
Perhaps in future someone is making data updates and does not think about the data structure before....

But my config pages have dissapeared
[_config.yml
](https://github.com/WohnungNDS/IM_Site/blob/9f2de47fe498f0b1a173632b699246d8ccd776d4/_config.yml#L76)

MoBosse:
@brockfanning Seems to work fine!

LucyGwilliamAdmin:
@brockfanning causing no issues for 5.2.1 anymore! https://lucygwilliamadmin.github.io/sdg-indicators/5-2-1/

LucyGwilliamAdmin (approved):
Approving based on reviews from @MoBosse and @WohnungNDS and the fact that no issues being caused on problem UK indicator

LucyGwilliamAdmin:
@brockfanning by the way, does this functionality apply to series and units?

brockfanning:
@LucyGwilliamAdmin I believe so. It's been a while but I think that's what is happening [here](https://github.com/open-sdg/open-sdg/blob/1.5.0-dev/_includes/assets/js/indicatorModel.js#L53) and [here](https://github.com/open-sdg/open-sdg/blob/1.5.0-dev/_includes/assets/js/indicatorModel.js#L86).

LucyGwilliamAdmin:
@brockfanning just tested and it works, yes - thank you

LucyGwilliamAdmin:
@brockfanning I'm not sure if I'm doing something wrong but I'm trying the `sorting: default` option and the order doesn't seem to be matching the order of what's in the file but rather it's alphabetical.

See here: http://uk-sdg-feature-branches.s3-website.eu-west-2.amazonaws.com/feature-ordering-test-default/15-2-1/

Any idea?

brockfanning:
@LucyGwilliamAdmin You're right, the "Series" field in the datapackage is getting sorted alphabetically: http://uk-sdg-feature-branches.s3-website.eu-west-2.amazonaws.com/feature-ordering-test-default-data/en/data-packages/15-2-1/datapackage.json

That means the problem is happening on the sdg-build side. I'll look into it.

brockfanning:
@LucyGwilliamAdmin Actually maybe it's a typo - try `datapackage` instead of `data_package` in the config?

LucyGwilliamAdmin:
Oh yeah, it's working now - thank you :)

LucyGwilliamAdmin:
I am trying out the schema bit now and seem to be getting an error:
https://github.com/ONSdigital/sdg-data/runs/2965298605?check_suite_focus=true

LucyGwilliamAdmin:
@brockfanning just had a thought - could it be something to do with this: https://github.com/open-sdg/sdg-build/pull/245

PR #1072, opened 2021-01-23
Allow numeric disaggregation

### LINKED ISSUE #327: Numeric disaggregation columns cause issues
This is more of a hunch, and needs confirmation. It seems likely that numeric disaggregation columns may cause problems:
* Empty options in the disaggregation widget on the left
* Empty table and/or javascript errors

This is a follow up from #293

brockfanning:
This is intended to demonstrate that https://github.com/open-sdg/sdg-build/pull/198 is working. When pointing at that PR's branch, the tests here pass. But when pointed back at 1.3.0-dev, the tests fail.

After https://github.com/open-sdg/sdg-build/pull/198 is merged, the tests here can be re-run and they should pass, at which point this can be approved/merged.

### DIFF
--- a/tests/data/data/indicator_3-1-1.csv
+++ b/tests/data/data/indicator_3-1-1.csv
@@ -1,5 +1,5 @@
-Year,Group,Value
-2015,A,1
-2015,B,3
-2016,A,1
-2016,B,3
+Year,Numeric Disaggregation,Value
+2015,7,1
+2015,8,3
+2016,7,1
+2016,8,3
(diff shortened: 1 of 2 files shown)


jwestw:
@brockfanning I will retest this after open-sdg/sdg-build#198 has been merged into 1.3.0-dev. I'll leave the merging to you though

jwestw (approved):
The test passes now. We will leave this in as sample data so the Numeric Disaggregation always gets tested.

PR #1073, opened 2021-01-23
Precision, decimal separator, and data display alterations

### LINKED ISSUE #341: Zero as decimal place is not displayed
grafik

I have just seen that this problem does not occur on the UK site:

grafik

What do I need to change?

### LINKED ISSUE #516: Graph not showing same values as table when rounded
I've just added some rounding functionality for Kyrgyzstan but the rounded values aren't showing on the graph, only on the table.

For example, see [1.2.1](https://sdg-kyrgyzstan.github.io/open-sdg-site-starter/1-2-1/)
Table:
image

Graph:
image

### LINKED ISSUE #991: Decimal separator support for multiple languages
Some languages, like German or Spanish, use a comma as a decimal separator instead of a point.
Would it be possible to show numbers in this way in tables and graphics? E.g. `13,2` instead of `13.2`.

We have tried parsing numbers by means of a function in `custom.js` (inspired by the ONS site @LucyGwilliamAdmin):

```
opensdg.dataRounding = function(value) {
  if (value == null) {
    return value
  }
  else {
    return value.toFixed(1).replace('.', ',')
    // returns a character (e.g. "13.0")
    // for mose reason, "0.0" is still shown as "0" in tables
  }
};
```

Then values are shown as wanted in tables, but graphics stop to work.

brockfanning:
Fixes #341 
Fixes #991 
Fixes #516 

I took the opportunity to bundle together a few related features here:
* Per-indicator, per-series, and per-unit precision control
* Site-wide decimal separator control
* Site-wide data display alteration

## Precision control

There is a new special metadata field called `precision` which can be used in the following ways.

If you want to force 1 decimal place for an indicator (eg, `5` would become `5.0` and `5.99` would become `6.0`):

```
precision:
  - decimals: 1
```

If you want to force 1 decimal place only when the unit of measurement is "total":

```
precision:
  - unit: total
    decimals: 1
```

If you want to force 1 decimal place only when the unit of measurement is "total" and the series is "foo":

```
precision:
  - unit: total
    series: foo
    decimals: 1
```

If you want to force 1 decimal place when the unit of measurement is "total, and 2 decimals places when the series is "foo":

```
precision:
  - unit: total
    decimals: 1
  - series: foo
    decimals: 2
```

## Decimal separator

This is a new **site** config setting called `decimal_separator` which you can use to replace the default `.`. For example this would use a comma:

```
decimal_separator: ','
```

## Data alterations

If the above features are not enough and you still want to alter the data in some other ways before displaying, there is now a javascript hook: `opensdg.dataDisplayAlter`. This works similarly to `opensdg.tableConfigAlter` and `opensdg.chartConfigAlter`.

On a related note, I don't believe that any of these javascript hooks were documented anywhere, so I've gone ahead and documented them in this PR.

## Feature branch:
http://sdgdev-813006012.eu-west-1.elb.amazonaws.com/preserve-zeroes-as-decimals


brockfanning:
To be tested by the data team.

LucyGwilliamAdmin:
@brockfanning I've just tried out the precision part of this but it doesn't seem to be working:
http://sdgdev-813006012.eu-west-1.elb.amazonaws.com/preserve-zeroes-as-decimals/1-2-1/

file change:
https://github.com/ONSdigital/sdg-data/commit/1ac15bc9ed207cfcdc2d6d6faf7661394728524b

brockfanning:
@LucyGwilliamAdmin It might be because you're overriding _includes/scripts.html, which this PR changes. I would try deleting that override. Also FYI instead of overriding scripts.html you could use one of the 2 options here: https://open-sdg.readthedocs.io/en/latest/customisation/#adding-custom-javascript

LucyGwilliamAdmin:
@brockfanning that fixed it - thanks. For some reasons I thought I already implemented one of those.

LucyGwilliamAdmin:
@emmawoodons This is the corresponding pull request to the issue I tagged you in

This is the affect it has:
http://sdgdev-813006012.eu-west-1.elb.amazonaws.com/preserve-zeroes-as-decimals/1-2-1/
http://sdgdev-813006012.eu-west-1.elb.amazonaws.com/preserve-zeroes-as-decimals/1-5-1/

And the changes needed to the metadata file:
https://github.com/ONSdigital/sdg-data/commit/1ac15bc9ed207cfcdc2d6d6faf7661394728524b
https://github.com/ONSdigital/sdg-data/commit/f33d031c032d4e45c6dfe0f9dc6d6957247dbb5e

LucyGwilliamAdmin:
@emmawoodons replying here so not jumping back and fore

> @LucyGwilliamAdmin This looks really good, particularly with the ability to control number of dp based on units: The example you sent for 1.5.1 is fine for Units=Rate, but doesn't work for Units=Deaths, but it looks like the unit control will fix this. Would we be able to control it based on both series and units ?

Ah yes, I see what you mean - I thought because I only mentioned the rate units, that the change would only apply to that but I see deaths has also been given 5 decimal places.

@brockfanning is this behaviour expected?

brockfanning:
I think I know what that's happening; I'll investigate more.

brockfanning:
@LucyGwilliamAdmin I pushed up a fix for that - this is ready for a refresh.

LucyGwilliamAdmin:
@emmawoodons I believe Brock's latest push has fixed 1.5.1 , right?

LucyGwilliamAdmin:
`decimal_separator` is working - changed to comma:
http://sdgdev-813006012.eu-west-1.elb.amazonaws.com/preserve-zeroes-as-decimals/1-5-1/

EmmaWoodONS:
@LucyGwilliamAdmin Yep, looks good to me!

LucyGwilliamAdmin:
@brockfanning what's the difference between `opensdg.dataDisplayAlter` and  `opensdg.dataRounding`, which is used here:

https://github.com/ONSdigital/sdg-indicators/blob/preserve-zeroes-as-decimals/assets/js/custom.js

Should we use `dataDisplayAlter` instead?

brockfanning:
@LucyGwilliamAdmin The main difference is the time when each one happens. The `dataRounding` function happens before the numbers are passed along to the chart/table/map. This means, importantly, that the `dataRounding` function cannot return a string - it has to remain a number. By contrast, the `dataDisplayAlter` function happens right before the chart/table/map is displayed. This means that it is OK for `dataDisplayAlter` to return a string (which is how this PR is able to get zeroes to display after a decimal).

For example, dataRounding can return `5`, but cannot return `"5"`. And dataDisplayAlter can return `5`, `"5"`, `"foo"`, or whatever.

Another technical difference is that the parameter for dataRounding is the value itself (`dataRound(value)`), but the parameter for dataDisplayAlter is a function, which takes the value as its parameter (`dataDisplayAlter(function(value)`). So dataDisplayAlter is slightly more confusing if you are not used to "callback functions".

I think you're probably right that `dataDisplayAlter` should be preferred over `dataRounding`. It's more flexible in that it can return a string or a number. Maybe we could eventually deprecated `dataRounding`? I'd interested to hear your experiences.

LucyGwilliamAdmin:
@brockfanning that was one of the main reasons behind my asking actually - whether or not we should be using that instead of rounding. I think it depends:

is the value column still required to be a number? - if so, that data rounding function aligns with that and helps ensure that the wrong data type isn't passed as a value.

The other reason I was asking was for testing purposes? any suggestions on how to test that part?

brockfanning:
To be clear, both dataRounding and dataDisplayAlter expect the existing value to be a number. The difference is in what they can return. While dataRound has to return a number, dataDisplayAlter can return either a number or a string.

The "Value" column definitely still needs to be numeric in the source data, but I think that automated tests in the data repo will fail if that value is not numeric.

If you want to test the difference you could try removing your dataRounding function and putting the equivalent in dataDisplayAlter. For example, you could change your custom.js file from:

```
opensdg.dataRounding = function(value) {
  if (value == null) {
    return value
  }
  else {
    return Number(value.toPrecision(3))
  }
};
```

To this:

```
opensdg.dataDisplayAlter(function(value) {
  if (value == null) {
    return value
  }
  else {
    return Number(value.toPrecision(3))
  }
});
```

In theory it should work the same.

LucyGwilliamAdmin:
@brockfanning 
Tried setting precision on more complex example:
Used the following on [15.2.1](http://sdgdev-813006012.eu-west-1.elb.amazonaws.com/preserve-zeroes-as-decimals/15-2-1/):

```
precision:
  - unit: Growth rate (%)
    series: Woodland area net change
    decimals: 4
  - unit: Percentage (%)
    decimals: 1
  - series: Above-ground biomass stock
    decimals: 0
```

This gives the following results:
- the values for `Growth rate (%)` with `Woodland area net change`  are to 4 decimal places. However, 'NaN' is showing instead of '-' where values don't exist
- the values for `Percentage (%)` with `Percentage of woodland under long-term management plan` aren't to 1 d.p. when the decimal place is a zero (see year 2020). I'm not sure about the other series with `Percentage (%) but imagine possibly same issue.

I did also test swapping out the hook but that didn't seem to work. For example, for Growth rate (%) with Woodland area net change was showing all decimal places (even with the above settings set.

brockfanning:
Thanks @LucyGwilliamAdmin I believe I've got all these issues fixed and ready for a refresh/retest.

LucyGwilliamAdmin:
Ok, just refreshed now and 15.2.1 looks good - with and without the JS hook being changed

Is there anything else you think I should check? Otherwise I'm happy to approve

brockfanning:
@LucyGwilliamAdmin Sorry I didn't mention it before, but maybe a check of maps, if you haven't already? If I remember right, this functionality should work on maps too.

LucyGwilliamAdmin:
This also works on map as expected:
http://sdgdev-813006012.eu-west-1.elb.amazonaws.com/preserve-zeroes-as-decimals/3-a-1/

LucyGwilliamAdmin (approved):
All scenarios have been tested and are working well so approving - link in description

LucyGwilliamAdmin:
@brockfanning I've just approved but actually just realised there's a conflict - do you want me to refresh and recheck that everything is working as expected after the conflicts have been resolved?

brockfanning:
@LucyGwilliamAdmin It was a pretty straightforward conflict - I'd say if the automated tests still pass then it's fine for us to merge it.

PR #1075, opened 2021-02-02
Support different min/max values per map layer

brockfanning:
Work in progress - not sure if we'll need this.

Feature branch: http://sdgdev-813006012.eu-west-1.elb.amazonaws.com/min-max-values-in-maps/3-a-1


brockfanning:
@LucyGwilliamAdmin @jwestw This could be used as a feature branch to test out #356.

LucyGwilliamAdmin:
Link to feature branch in description and I've been in touch with geospatial - awaiting reply

brockfanning:
@LucyGwilliamAdmin Looking at the feature branch I realize that it only triggers when zooming *in*, but not when zooming out. But I guess it's enough to get across the idea, for feedback at least.

LucyGwilliamAdmin:
@brockfanning @phillipgardiner our Geospatial team provided the following advice re. min/max values:

> As the map incorporates a temporal slider, I think it is fine for the minimum and maximum values to represent the smallest and largest value from the entire dataset – taking into account the two different zoom levels this particular map offers (countries and English regions).

LucyGwilliamAdmin:
@brockfanning think we agreed to close this one right?

brockfanning:
Yes, closing now.

Issue #1077, opened 2021-02-02
Checklist for PRs


LucyGwilliamAdmin:
Example:
https://github.com/json-editor/json-editor/pull/882

LucyGwilliamAdmin:
@brockfanning doesn't seem like the PR and issue were properly linked as this wasn't closed by merging the PR

brockfanning:
@LucyGwilliamAdmin Yeah I think Github only does that when the line is just "Fixes #..." but that's OK, I still think it's useful to have the link easily visible in the table.

brockfanning:
Some feedback on what we have so far:
The checklist is great and very useful. There are only a couple of awkward bits:
1. The changelog item can not actually be "checked" until the PR is created. This is because changelog entries have to include the PR number, which we don't know until we create it. There's no way around this, but it does mean that, to be totally correct, the PR will need to be created with ❌ and then later updated with ✔️ when the changelog has been added.
2. There are times when I need to add a ❌ or a ✔️ but there is not another one on the page. So I need to either know the keyboard shortcut (eg `:x`) or go find another PR to copy/paste.

One possible solution to these is to use checkboxes for the yes/no questions, and then a table for everything else. For example:

Q | A
--- | ---
Feature branch/test site URL | [Link](add link here)
Fixed issues | Fixes #ISSUENUMBER
Related version | 1.3.0-dev
Bugfix, feature or docs? |

* [ ] Keeps backward-compatibility
* [ ] Includes tests
* [ ] Updated docs
* [ ] Updated config forms
* [ ] Added CHANGELOG entry

This avoids the need to copy/paste ❌ or ✔️, and makes it easier to update the checklist after adding the changelog (since you don't have to edit the PR's description, you can just check the box.

PR #1080, opened 2021-02-05
Create template for PRs

LucyGwilliamAdmin:
Currently looks like this (updated following feedback):
Q | A
--- | ---
Fixed issues | #`issue`
Related version | 1.3.0-dev
Bugfix, feature or docs |
Keeps backward-compatibility? |❌/✔️
Updated docs/config-forms? |❌/✔️
Added CHANGELOG entry? |❌/✔️

So filled in for this PR as an example (updated following feedback):
Q | A
--- | ---
Fixed issues | #1077
Related version | master
Bugfix, feature or docs | Feature
Keeps backward-compatibility? |NA
Updated docs/config-forms? |NA
Added CHANGELOG entry? |NA

Most of these fields came from this [example on json-editor repo](https://github.com/json-editor/json-editor/pull/882) but I've added some others.

Let me know if you think there's any we should remove or if there are others we should add?

### DIFF
--- a/pull_request_template.md
+++ b/pull_request_template.md
@@ -1,0 +1,12 @@
+|               Q              | 	 A    |
+|----------------------------- |--------|
+|Fixed issues                  |#`issue`|
+|Related version               | v. no. |
+|Is bugfix?	                   | ❌/✔️ |
+|New feature?                  | ❌/✔️ |
+|Breaks backward-compatibility?| ❌/✔️ |
+|Tests pass?                   | ❌/✔️ |
+|Feature branch/test site URL  |[Link]()|
+|Reviewed                      | ❌/✔️ |
+|Updated README/docs?          | ❌/✔️ |
+|Added CHANGELOG entry?        | ❌/✔️ |
--- a/pull_request_template.md
+++ b/pull_request_template.md
@@ -1,12 +1,9 @@
-|Q                                                 |A         |
-|-----------------------------------|--------|
-|Fixed issues                                |#`issue`|
-|Related version                           |v. no.   |
-|Is bugfix?	                            |❌/✔️|
-|New feature?                              |❌/✔️|
-|Breaks backward-compatibility?|❌/✔️|
-|Tests pass?                                 |❌/✔️|
-|Feature branch/test site URL     |[Link]()|
-|Reviewed                                   |❌/✔️|
-|Updated README/docs?           |❌/✔️|
-|Added CHANGELOG entry?       |❌/✔️|
+Q | A
+--- | ---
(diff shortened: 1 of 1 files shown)


brockfanning:
@LucyGwilliamAdmin Cool. This is just my preference, but when I see those symbols I think of ❌ as "bad" and ✔️ as "good". So I'm wondering if the questions can be worded in a way so that we eventually want all ✔️. Some notes on each one:

Q | A | Notes
--- | --- | ---
Fixed issues | #1077 | If this works with github's auto-linking feature, great. I'll test that out next time I open a PR.
Related version | NA | Most PRs will be for the upcoming version (1.3.0-dev right now) so maybe we can provide that default here.
Is bugfix? | ❌| Maybe this question could be "Bugfix, feature, or docs?" with plain text answers?
New feature? | ✔️ | See above
Breaks backward-compatibility? | ❌| I'd reverse this, like "Keeps backward-compatibility" so the ✔️is "good".
Tests pass? | ✔️ | I don't think we need this because it shows up automatically.
Feature branch/test site URL | NA | Sounds good
Reviewed | ❌| I don't think we need this because it shows up automatically.
Updated README/docs? | ❌| Maybe this could also include configuration form updates. Like: "Updated docs/config-forms?"
Added CHANGELOG entry? | ❌| Sounds good

LucyGwilliamAdmin:
@brockfanning I agree with those changes. I'll see if there's any other thoughts before I make the changes though

AnnCorp:
@LucyGwilliamAdmin this is a great idea. Agree with Brock's comments. 
Only one further thought: might be helpful to have the link to feature branch/test url nearer the top so quicker to access (I know those links will break eventually once feature branches will be removed, but while dev in progress think easier if the link at top).

LucyGwilliamAdmin:
@brockfanning @AnnCorp this has just been updated following your feedback - see top comment for how it looks now

brockfanning:
@LucyGwilliamAdmin My only suggestion is to add a "?" after "Bugfix, feature, or docs". Looks great!

LucyGwilliamAdmin:
@brockfanning oh yeah - missed that, just added

brockfanning (approved):

LucyGwilliamAdmin:
@brockfanning I should have merged into 1.3.0-dev since that is now the default branch - should I revert this PR and merge into 1.3.0 instead?

brockfanning:
@LucyGwilliamAdmin No I think we have some stuff in master that needs to go into 1.3.0-dev anyway. I'd suggest creating a new PR to merge master into 1.3.0-dev.

Issue #1014, opened 2020-12-09
Recommendations for CSV editing tools


brockfanning:
In the past we've discussed alternatives to the Prose.io CSV editing functionality (since Prose.io has a few quirks that can prevent it from being used in some cases). Some countries may be using Excel to edit CSV files, or automatically generating CSV files from some other maintenance tool.

I'm interested in hearing any recommendations for approaches to CSV file edits. Is anyone using anything else? Does anyone know of any web-service and/or open-source library for CSV file editing?

One feature I'm particularly interested in is integration with some sort of data schema (like an SDMX DSD for example) which could convert cells into dropdowns, to require particular values. For example, instead of a free-form cell, a dropdown with "Rural" and "Urban" as the only options.

brockfanning:
@jwestw Just wondering if you know of any popular CSV editors in the CSV-W world, which might allow for cells to become dropdowns as described above.

brockfanning:
Since Excel is universally available, maybe it would be a good choice? The one problem with Excel files is that they will not display differences in Github (it will just say "binary file changed" or similar). I suppose one way around this problem would be for the country to export to CSV before uploading, and for sdg-build to automatically generate Excel files per indicator (with the appropriate cells converted into dropdowns). So, from the platform's perspective nothing would change - it would still be CSV files. But the data manager would have the option of maintaining the data with .XLSX files with dropdowns.

LucyGwilliamAdmin:
I think Excel could work. However re. it being universally available, is there a web-based version that could be used? Otherwise another issue I see is that a Microsoft license would be required

brockfanning:
I was thinking that sdg-build could automatically produce the .xlsx files. Any countries that prefer not to use Excel could then upload them to Google Docs as an alternative.

But that is a bit of manual work, and it would still be nice to know of any Prose-like services out there, which might support dropdowns for cells.

phillipgardiner:
@brockfanning @LucyGwilliamAdmin I might be misinterpretating this, but we would also need to check around what formats are acceptable. For example I think our government promotes the use of Open Document Formats (ODF), therefore tieing into .xls may not be acceptable in regards to assessment etc.

jwestw:
RE: the above conversation I think ODF is the way to go as it's non-proprietary. 

These are not recommendations but I looked around on Github and saw a number of CSV editors that (I think) can be used online:

[...]

PR #1082, opened 2021-02-07
Config form improvements

brockfanning:
This PR includes a few improvements to the config forms:

This depends on https://github.com/open-sdg/jekyll-open-sdg-plugins/pull/82

## Data and Metadata config forms

This adds the ability to use a spreadsheet-like form for indicator data, and a config form for indicator metadata. There are some special configuration options for metadata forms:

* "language": This determines the language that the metadata field labels should appear in. In general this should be the preferred language of the people editing the metadata.
* "scopes": This "scopes" to include in the metadata form. In general this should be "global" and/or "national".
* "exclude_fields": Any fields that you would prefer to keep out of the form, such as any field that is automatically populated.
* "translated": This should be used if using the "subfolder approach" for translated metadata. Anyone using this approach will need to save separate metadata files for each language. When this option is enabled, the contents of the metadata forms is translated (based on the current language), to allow this usage.

Here is an example of using these metadata config form options:

```
indicator_metadata_form:
  language: ru
  scopes:
    - global
    - national
  exclude_fields:
    - national_metadata_update_url
    - national_metadata_update_url_text
    - national_data_update_url
    - national_data_update_url_text
  translated: true
```

## New way to enable config forms

Previous the way to enable config forms was:

```
create_config_forms:
  layout: config-builder
```

One problem is that this is a bit confusing. Also, it is unnecessarily complex because users will never want to change the layout. Also, it is possible that a user would want one type of config form but not another. So the proposed new way to enable config forms is like so:

```
site_config_form:
  enabled: true
indicator_config_form:
  enabled: true
indicator_data_form:
  enabled: true
indicator_metadata_form:
  enabled: true 
``` 

## "Go to repository" button

Because these downloaded files need to be uploaded to a specific folder in Github, I thought it would be useful to have a "Go to repository" button which takes the user to that specific folder. This way they can click the "Download configuration" button and then immediately click "Go to repository" to upload it. The way this is configured is like so:

```
site_config_form:
  repository_link: https://github.com/myorg/mysite/tree/develop/_data
indicator_config_form:
  repository_link: https://github.com/myorg/mydata/tree/develop/indicator-config
indicator_data_form:
  repository_link: https://github.com/myorg/mydata/tree/develop/data
indicator_metadata_form:
  repository_link: https://github.com/myorg/mydata/tree/develop/meta
``` 

## "Go to translation" links

Often configuration or metadata values consist of "translation keys" which refer to translations. In these cases, it is helpful to provide a link to the place where that translation could be performed - such as a folder in Github or a page on a translation management system. With this PR if a "translation_link" is provided, then a link will appear whenever the value of a field is a translation key. The link can be dynamic based on the "group" and/or the "key" of the translation key (eg, "my-group.my-key") by using the placeholders "[group]" and/or "[key]".

For example, to point to a Github folder, you might do something like:

```
indicator_metadata_form:
  translation_link: https://github.com/myorg/mydata/tree/develop/translations/en/[group].yml
```

Or to point to a translation management system like Weblate:

```
indicator_metadata_form:
  translation_link: https://hosted.weblate.org/search/my-project/[group]/?q=+context%3A%3D[key]
```


LucyGwilliamAdmin:
@brockfanning has this got the latest from 1.3.1?

brockfanning:
@LucyGwilliamAdmin Ah, probably not, I'll refresh the branch.

LucyGwilliamAdmin:
@brockfanning thank you :)

brockfanning:
@LucyGwilliamAdmin @jwestw Just a note that I believe this is ready for review now. Note the "translated" setting that I just added, which should help the metadata forms to be used even by countries using the subfolder approach.

brockfanning:
@LucyGwilliamAdmin @jwestw I just added spreadsheet-like data forms to this PR, if you want to give it a test. As mentioned [here](https://github.com/open-sdg/open-sdg/issues/1014#issuecomment-836840477), they are not fully accessible yet. I was able to hack in some keyboard navigation, but the screenreader support is bad. I've updated the top post here with instructions for testing.

brockfanning:
@LucyGwilliamAdmin @jwestw Correction - I found a library called ag-grid that has vastly better accessibility than handsontable, so I've switched. I think it's probably even good enough to show to DAC. :)

LucyGwilliamAdmin:
@brockfanning is the indicator config form supposed to change language when the language is changed on the site?

LucyGwilliamAdmin:
Can't test on this site as data input is SDMX but here's how the data form looks: https://open-sdg-nepal.github.io/site/1-3-1/data/

LucyGwilliamAdmin:
For the dropdown label options, is it possible for these to be translation keys?

LucyGwilliamAdmin:
Note: collating comments/feedback here and deleted separate ones

## Indicator configuration form
### Questions:
1.  Is the indicator config form supposed to change language when the language is changed on the site?
2. For the dropdown label options, is it possible for these to be translation keys?

### Comments:
1. "Go to translations..." link doesn't always go to new line:
  image

## Data form
### Questions:
1. When you add a new column is it possible for it to be added before the Value column, as that needs to be the final column right? No probs if not possible, could just add a note to say to move it, since it's possible to drag and drop columns
2. Should it be possible to delete and rename columns? Except Year and Value columns

### Comments:
I've noticed that you need to make sure that you click off your last edited cell before clicking 'Download data'

brockfanning:
@LucyGwilliamAdmin Great feedback, I'll look into those items.

Regarding the translation of the indicator configuration form itself - we have not really implemented anything like that. All of those titles/descriptions of the indicator configuration fields are hardcoded English right now. And no, the dropdown labels can't be translation keys. What I have done (in one case - Kazakhstan) is to configure the labels according to the preferred language of the editors (which in the case of Kazakhstan is Russian). [Example here](https://github.com/kazstat/sdg-site-kazstat/blob/develop/_data/site_config.yml#L88-L91)

brockfanning:
@LucyGwilliamAdmin I've updated this with a fix for the translation link line problem, the last-edited download problem, and the adding of columns second-to-last.

About renaming/removing columns, I agree this would be nice. I'm hesitant though for a couple reasons:
1. There are several things that are easier to do in Excel, like drag-filling, cutting/pasting several cells at once, etc. Since the library we're using (ag-grid community edition) is not capable of matching Excel's functionality, my thinking is that users should download the file and work in Excel for more involved changes, and only use this form for quick fixes or the adding of new rows. So in theory users could switch to Excel for renaming/removing columns. This data form actually does facilitate that a bit, thanks to the "Download data" and "Go to repository" buttons.
2. Since ag-grid doesn't have a feature for renaming/removing columns, we would need to code it ourselves. It's definitely possible, but would involve some UX decisions. Also we'd need to do it in a way that maintains full keyboard accessibility.

LucyGwilliamAdmin:
Graph title from series option doesn't seem to be showing on site config form, but I don't know if that's because I'm pointing at `meta-config-plus-series-titles`

Just tested the fixes you did, and they look good. Think no need to do anything about the renaming/removing of columns!

brockfanning:
@LucyGwilliamAdmin I think that would depend on the branch of jekyll-open-sdg-plugins that you're using - maybe you're using an older branch for that?

LucyGwilliamAdmin:
I'm pointing at `metadata-config-pages` for jekyll plugins. Although I can't see the field mentioned in the schema: 
https://github.com/open-sdg/jekyll-open-sdg-plugins/blob/eb03ae810dd9d234a3536ef6c502417fb4d434f2/lib/jekyll-open-sdg-plugins/schema-site-config.json

brockfanning:
@LucyGwilliamAdmin Ok sorry - I just refreshed that branch on jekyll-open-sdg-plugins, so if should appear now.

LucyGwilliamAdmin:
@brockfanning yes, it's there now - looks good - think we could approve this now?

brockfanning:
Yep I think so. :)

LucyGwilliamAdmin (approved):

Issue #504, opened 2020-02-19
Feedback on "Sources" functionality?


brockfanning:
This is a call for any feedback on the Sources functionality, as described [here](https://open-sdg.readthedocs.io/en/latest/metadata-format/#data-sources-metadata). Since we're getting close to release version 1, this would be the time to make any architectural changes related to sources. So:
1. Any annoyances or limitations with the current system?
2. Any source-related features you would like added?
3. Any source-related bugs you've noticed?
4. Other thoughts?

MoBosse:
Hi @brockfanning, here are a few things that come to my mind about sources:

- The placement of the third (fifth, seventh...) source depends on the length of the value in the last tablerow of source 1 and 2 (3 and 4,...). See screenshot:
![grafik](https://user-images.githubusercontent.com/50657629/74925842-3b860600-53d5-11ea-9b7d-bc0f9be0c24c.png)

- It looks strange when two sources are side by side, but have table rows with different heights:
![grafik](https://user-images.githubusercontent.com/50657629/74926503-44c3a280-53d6-11ea-9715-667f8722427b.png)

- Not sure if it's of any interest for anyone but us, but sometimes the source_organisation contains additional information beside the Organistaion name, so I added a **source_organisation_#_short** that is shown in the sources tab beside the original **source_organisation_#** that is shown in the footer. 


![grafik](https://user-images.githubusercontent.com/50657629/74927422-d849a300-53d7-11ea-87f8-7031183da860.png)

brockfanning:
Thanks for the feedback @MoBosse, very helpful.

ONSBen:
Hello @brockfanning 

I agree with the Source 3 / 5 / ... position and row alignment issue raised above.

Looking specifically at the 'Data last updated' and 'Metadata last updated' field, would it be possible for that to link to several of the most recent changes, rather than just one? Or to link to a GitHub page of all changes to that indicator? The intent is to make it easier to check older versions of the data / metadata.

While I think that the metadata tabs in general are a little too discrete, I don't know much about user-friendly design or what other user experiences are, so take that with a pinch of salt.

Many thanks

brockfanning:
@MoBosse @ONSBen In regards to the position/alignment issue, do you think it would be preferable to have them in a single column? To see what it would look like, I've got a pull-request up with a screenshot here: #591

MoBosse:
[...]

PR #1087, opened 2021-02-10
Fix layout issues with odd-numbered indicator sources

brockfanning:
This is a fix for one of the issues mentioned in #504 -- namely the problem where the placement of the odd-numbered sources is inconsistent depending on the height of the even-numbered sources. This change forces odd-numbered sources to always align to the left.

There is another issue mentioned in #504 -- namely the incongruity of the horizontal lines when different-height fields are next to each other -- which needs a design discussion, so is not addressed here.

Testing note - if using a feature branch, the UK's indicator 5-5-1 is a good one to test.

### DIFF
--- a/_sass/layouts/_indicator.scss
+++ b/_sass/layouts/_indicator.scss
@@ -595,3 +595,9 @@ body.contrast-high {
     }
   }
 }
+
+#sources {
+  .row > div:nth-child(odd) {
+    clear: left;
+  }
+}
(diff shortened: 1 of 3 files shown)


jwestw:
Here's indicator 5-5-1 on the [feature branch](http://sdgdev-813006012.eu-west-1.elb.amazonaws.com/sources-layout-fixes).

jwestw (approved):
This is forcing the odd numbered sources to left align which fixes the problem. 

@phillipgardiner I have already approved, but as it's about layout I am not sure if you want to look over this one before it's merged.

jwestw:
**The sources tab on [UK SDG website indicator 5-5-1](https://sdgdata.gov.uk/5-5-1/) as is:**

image


**And the sources tab on the feature branch indicator 5-5-1:**

image

phillipgardiner:
@jwestw Thank you. Looks great.

PR #1089, opened 2021-02-16
Do not use empty titles in head

brockfanning:
When using the config forms, empty values are saved as empty strings (`''`). But these are being placed in the page title, so it ends up like: `- Indicators for the Sustainable Development Goals`. This extra check avoids that, so it will properly be `Indicators for the Sustainable Development Goals`.

If you would like to test this one, you could try giving an empty title to the homepage. Eg:

```
create_pages:
  - folder: /
    layout: frontpage
    title: ''
```

EDIT: [Feature branch](http://sdgdev-813006012.eu-west-1.elb.amazonaws.com/skip-empty-titles)

### DIFF
--- a/_includes/title-tag.html
+++ b/_includes/title-tag.html
@@ -7,7 +7,7 @@
     </title>
 {% elsif page.goal %}
     <title>{{ page.t.general.goal | escape }} {{ page.goal.number }} - {{ page.goal.short | escape }} - {{ site.title | t | escape }}</title>
-{% elsif page.title %}
+{% elsif page.title and page.title != '' %}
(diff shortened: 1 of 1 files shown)


jwestw:
Here is the feature branch: http://sdgdev-813006012.eu-west-1.elb.amazonaws.com/skip-empty-titles

jwestw:
This is working as expected. I removed the title for both the front page and the search results, replacing them with ''. The title that came up was "U.K Indicators for the Sustainable Development Goals" in both cases with no "-" or space.

jwestw (approved):
Approving based on last test and comments.

PR #1090, opened 2021-02-16
Use https to avoid mixed content error

brockfanning:
This is a quick change to avoid "mixed content" errors in browsers. Some (all?) browsers do not like it when a page loads some assets using "https" and others using "http", and they will block the "http" assets.

I don't think this is worth a feature branch.


LucyGwilliamAdmin (approved):
Looks good and makes sense

PR #1091, opened 2021-02-16
Bug fix - variable name in config builder

brockfanning:
@LucyGwilliamAdmin @jwestw This is a straightforward bugfix -- "choices" is not a variable there, it was meant to be "dropdown".

I don't think this is worth a feature branch.


brockfanning:
@LucyGwilliamAdmin @jwestw Also by the way, to see an example of the functionality that this fixes - see here: https://kazstat.github.io/sdg-site-kazstat/1-1-1/config/#root[reporting_status]

Notice how the dropdown has custom labels - the maintainers' first language is Russian so I thought it would be helpful to have at least those options in Russian.

LucyGwilliamAdmin (approved):
Looks good to me!

PR #1092, opened 2021-02-16
Switch to Lodash instead of Underscrore

### LINKED ISSUE #943: Use Lodash instead of Underscore

This is not related to any particular problem, just a potential for a slight performance increase and dev experience improvement.

The library [Lodash](https://lodash.com/) is a superset of [Underscore](https://underscorejs.org/) (which we already use) and Lodash is reportedly faster. Being a superset it also has some additional utility functions. In theory we could replace Underscore with Lodash and nothing would change (except a slight speed increase and more utility functions).

No particular reason other than the need for faster rendering on the indicator pages.

We could always leave things as-is, so this is not a high priority.

brockfanning:
Fixes #943 

To test this, we should mainly focus on indicator pages, and making sure that everything still works.

EDIT: [Feature Branch](http://sdgdev-813006012.eu-west-1.elb.amazonaws.com/underscore-to-lodash/)

### DIFF
--- a/_includes/assets/js/chartjs/rescaler.js
+++ b/_includes/assets/js/chartjs/rescaler.js
@@ -23,7 +23,7 @@ Chart.plugins.register({
       return meta && meta.visible;
     });
 
-    var ranges = _.chain(datasets).pluck('allData').map(function (data) {
+    var ranges = _.chain(datasets).map('allData').map(function (data) {
       return {
         min: _.findIndex(data, function(val) { return val !== null }),
         max: _.findLastIndex(data, function(val) { return val !== null })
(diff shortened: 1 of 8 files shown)


jwestw:
I have looked at indicators on the [feature branch](http://sdgdev-813006012.eu-west-1.elb.amazonaws.com/underscore-to-lodash/). I can't see anything that is not working. 

@brockfanning Is there anything in particular that you think might break?

brockfanning:
@jwestw Not really, just indicator pages in general - the more complex the better. FYI, I just noticed there was a conflict that I had to resolve in the PR.

jwestw:
I rebuilt it after the conflict resolution. All seems good still.

jwestw (approved):
Approving. Looks good.

PR #1096, opened 2021-02-17
Add gutter to sources tab

brockfanning:
This is an attempt to address one issue mentioned in #504

### DIFF
--- a/_includes/components/indicator/sources.html
+++ b/_includes/components/indicator/sources.html
@@ -8,7 +8,7 @@
 {% endfor %}
 
 {% if n_sources > 0 %}
-  <div class="row no-gutters">
+  <div class="row">
   {% for i in (1..n_sources) %}
     {% assign src_active = "source_active_" | append: i %}
     {% assign src_scope = "source_" | append: i %}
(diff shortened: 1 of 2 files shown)


brockfanning:
Here's a screenshot:

screenshot-127 0 0 1_4000-2021 02 17-11_07_07

phillipgardiner:
@brockfanning thanks for the screenshot. I definitely think that helps :). I believe it is probably now a decision for @AnnCorp on weather any further time is spent addressing this (as per the slack conversation).

LucyGwilliamAdmin:
@brockfanning @phillipgardiner does this require anything from James or I?

brockfanning:
I'm not sure if this needs a feature branch, since it's straightforward enough that a screenshot is good. But I'll defer to @phillipgardiner.

AnnCorp:
@brockfanning @phillipgardiner I think this makes a big improvement - happy that this is good enough for now.

phillipgardiner:
@brockfanning @LucyGwilliamAdmin I'm happy don't think any need for a feature branch on this one.

LucyGwilliamAdmin (approved):
Approving based on comments

PR #1097, opened 2021-02-20
Make the meta/data edit links optional

brockfanning:
I don't remember the reasons for making these fields required, but with more and more implementations of Open SDG using data sources that are not file-based, it seems like these fields should be optional.

Corresponds to https://github.com/open-sdg/jekyll-open-sdg-plugins/pull/79


brockfanning:
On second thought, this can't be done without causing a breaking change, so we can leave it alone.

PR #1111, opened 2021-02-24
Add note item for tests and split docs/forms

brockfanning:
@LucyGwilliamAdmin  A tweak to the PR template - I realized that the test item was needed after all. And also thought maybe docs and forms could be on separate lines?

### DIFF
--- a/pull_request_template.md
+++ b/pull_request_template.md
@@ -5,5 +5,7 @@ Fixed issues | #`issue`
 Related version | 1.3.0-dev
 Bugfix, feature or docs? |
 Keeps backward-compatibility? |❌/✔️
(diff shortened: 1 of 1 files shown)


LucyGwilliamAdmin (approved):
@brockfanning sounds good to me

Also I had a thought about the issue-PR linking following on from what you said about needing `Fixes #issue-no` - wonder if having 'Fixes in the table cell would work?

Do you think worth a try when merging one of these and then could add it if works?

brockfanning:
Sounds good, I'll try that in a moment.

brockfanning:
It looks like that works. For example: https://github.com/open-sdg/open-sdg/pull/1117

Issue #1107, opened 2021-02-24
Change dropdown on focus text colour to black


phillipgardiner:
As a result of recent high contrast changes, it has been noticed that the colour contrast when a disggregation control has focus no longer meets 'AA' compliance.

Recommendation: Change text to black on focus for 
![Screenshot (150)_LI](https://user-images.githubusercontent.com/60919820/108985450-f8660f00-7688-11eb-9724-a50d11c6a3af.jpg)
high contrast mode.

PR #1116, opened 2021-02-24
Fix focus color on dropdowns in high-contrast

brockfanning:
Q | A
--- | ---
Feature branch/test site URL | [Link](http://sdgdev-813006012.eu-west-1.elb.amazonaws.com/high-contrast-dropdown-focus-color)
Fixed issues | #1107
Related version | 1.3.0-dev
Updated docs/config-forms? |NA
Added CHANGELOG entry? |❌

### DIFF
--- a/_sass/layouts/_indicator.scss
+++ b/_sass/layouts/_indicator.scss
@@ -521,6 +521,11 @@ body.contrast-high {
       background-color: $color-dark-highContrast;
       border-color: $color-light-highContrast;
       color: $color-light-highContrast;
+      &:focus {
+        span {
+          color: $color-dark-highContrast;
+        }
+      }
     }
   }
 


LucyGwilliamAdmin:
@brockfanning @phillipgardiner link to feature branch added. Looks good to me - text in now black in focus on high contrast mode except for in these cases but I assume the yellow here will change?

image

image

image

brockfanning:
@LucyGwilliamAdmin I think #1120 may fix those cases, but we should confirm when testing that one.

phillipgardiner:
@brockfanning @LucyGwilliamAdmin I don't think this one is correct yet? The text needs to be black when it first appears, not just when it is in focus.

LucyGwilliamAdmin:
@phillipgardiner I'm not too sure what you mean? if text is black won't it be same colour as background? Perhaps I'm looking in wrong place

phillipgardiner:
@LucyGwilliamAdmin Sorry, my fault, I was looking at the wrong ticket / feature branch combination!

LucyGwilliamAdmin:
@phillipgardiner ah no probs - #1120 does fix the highlight issue so think this one's ready?

LucyGwilliamAdmin (approved):
Looks good, text is black when dropdown is focussed on

PR #1121, opened 2021-02-24
More precise image switching code

### LINKED ISSUE #1118: Only keep same image in high contrast if image is SVG
#1039 had some unintended side effects. When looking at the site in high contrast mode, if images are PNG and not SVG then they're also not swapped for high contrast version.
 
Instead expected behaviour would be that if image is not an SVG, then use high contrast version of image instead

brockfanning:
Q | A
--- | ---
Fixed issues | Fixes #1118
Related version | 1.3.0-dev
Updated docs/config-forms? | NA
Added CHANGELOG entry? | NA (because this fixes a bug in the same release)

### DIFF
--- a/_includes/assets/js/accessibility.js
+++ b/_includes/assets/js/accessibility.js
@@ -107,19 +107,34 @@ var accessibilitySwitcher = function() {
 
 
   function imageFix(contrast) {
-    {% unless site.goal_image_extension == 'svg' %}
+    var doNotSwitchTheseSuffixes = ['.svg'];
+    var doNotSwitchThesePrefixes = ['https://platform-cdn.sharethis.com/'];
     if (contrast == 'high')  {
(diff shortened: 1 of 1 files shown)


jwestw:
On the [feature branch]()http://sdgdev-813006012.eu-west-1.elb.amazonaws.com/svg-icon-image-fix/ this appears to be working. 

This is a png image in high contrast mode:

image

This is a underlying link to the image:

image

As you can see it is "high-contrast" png. 

@Lucy does this fix the issue #1118 ?

@phillipgardiner can you also confirm this is okay?

phillipgardiner:
@jwestw Sorry I don't know anything about this one, one for @LucyGwilliamAdmin and @brockfanning

LucyGwilliamAdmin:
Yes png images are now showing in high contrast when in high contrast mode

jwestw (approved):
Approving according my tests and Lucy's comment.

PR #1128, opened 2021-02-28
Move goal header to include file

brockfanning:
Q | A
--- | ---
Fixed issues | NA
Related version | 1.3.0-dev
Includes tests? |NA (existing tests cover this)
Updated docs? |NA
Updated config forms? |NA
Added CHANGELOG entry? |❌

This removes some duplicate code and also makes it easier to customize the goal header.

### DIFF
--- a/_includes/components/goal/header.html
+++ b/_includes/components/goal/header.html
@@ -1,0 +1,14 @@
+<div class="heading goal-banner goal-{{ page.goal.number }}">
+  <div class="container">
+    <div class="row">
+      <div class="col-xs-4 col-md-3 col-lg-2 goal-tiles">
+        <img src="{{ page.goal.icon }}" alt="{{ page.goal.short | escape }} - {{ page.t.general.goal }} {{ page.goal.number }}" id="goal-{{ page.goal.number }}" class="goal-icon-image goal-icon-image-{{ site.goal_image_extension }}" />
+      </div>
+      <div class="col-xs-8 col-md-9 col-lg-10">
+        <h1>
+          <span class="hidden-sm hidden-md hidden-lg">{{ page.t.general.goal }} {{ page.goal.number }}: </span>{{ page.goal.name }}
+        </h1>
+      </div>
+    </div>
+  </div>
+</div>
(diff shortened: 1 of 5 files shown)


jwestw:
Hey @brockfanning I see your code. Looks like good application of [DRY](https://en.wikipedia.org/wiki/Don%27t_repeat_yourself). Is there anything here that you think I should particularly check? In particular, how would I know how this change "_makes it easier to customize the goal header_"?

brockfanning:
@jwestw What I meant by that is that if a country wants to override the header on goal pages, they only have to override the _includes/components/goal/header.html file, rather than overriding the entire _layouts/goal-by-target-vertical.html file (for example). As you've seen overriding files does carry a maintenance cost, and in general it is much easier to maintain a small override than a large one.

jwestw:
@brockfanning I am just wondering how to test https://github.com/open-sdg/open-sdg/pull/1128
I saw that _includes/components/goal/header.html did not exist, so I created it. Not sure if that's the right approach. I thought not, but I wasn't sure how to override it.

brockfanning:
@jwestw You can override any file in the _includes, _layouts, or assets folders of Open SDG just by creating a file at the same path in your site repository. So for example if Open SDG has a file called _layouts/foo.hml, and you want to override it, you can do so by creating a file at _layouts/foo.html in your site repository. Note I said "creating" because, yes, this file will not already exist. So in short, it sounds like you did the correct procedure for overriding the file.

In general, overriding files is something to avoid unless necessary, because it creates a maintenance burden: as soon as you override a file, you are no longer using Open SDG's version; so if Open SDG changed in some way that requires a change to the file, it will be up to you to change your version as needed. At the same time, the ability to override a file is a big part of how flexible Open SDG is.

So to reconcile these conflicting good/bad aspects of overrides, it's our job to make it possible for countries to have flexibility while overriding as little code as possible. That's exactly what I'm trying to do in this PR: allow flexibility in the goal headers by overriding a smaller chunk of code.

jwestw:
Thanks for your brilliant explanation @brockfanning

jwestw:
[Feature branch](http://sdgdev-813006012.eu-west-1.elb.amazonaws.com/goal-header-include)

jwestw (approved):
Approving according to last test

brockfanning:
I'll postpone merging until release 1.3.0 is out - since the changelog entry would cause a Git conflict.

PR #1133, opened 2021-03-04
replace series of IFs with CASE

EFT-Defra:
Q | A
--- | ---
Feature branch/test site URL | [EFT-Defra/open-sdg#EFT-Defra-patch-1](https://github.com/EFT-Defra/open-sdg/tree/1.4.0-dev)
Fixed issues | n/a
Related version | 1.4.0-dev
Bugfix, feature or docs? | n/a
Keeps backward-compatibility? | Not tested
Includes tests? |❌
Updated docs? |❌
Updated config forms? |❌
Added CHANGELOG entry? |❌

This is a very minor syntactic change motivated by my attempts to familiarise myself with Jekyll and OpenSDG. More than anything, it's intended to show willing, so feel free to reject!

### DIFF
--- a/_includes/components/reportingstatus/reporting-status-label.html
+++ b/_includes/components/reportingstatus/reporting-status-label.html
@@ -1,4 +1,4 @@
-{% assign status_css = indicator.reporting_status | slugify %}
+{% assign status_css = include.indicator.reporting_status | slugify %}
 {% case status_css %}
     {% when "notapplicable" %}
         {% assign status_desc = page.t.status.not_applicable %}
--- a/_includes/components/reportingstatus/reporting-status-label.html
+++ b/_includes/components/reportingstatus/reporting-status-label.html
@@ -1,0 +1,15 @@
+{% assign status_css = indicator.reporting_status | slugify %}
+{% case status_css %}
+    {% when "notapplicable" %}
+        {% assign status_desc = page.t.status.not_applicable %}
+    {% when "notstarted" %}
+        {% assign status_desc = page.t.status.exploring_data_sources %}
+    {% when "inprogress" %}
+        {% assign status_desc = page.t.status.statistics_in_progress %}
+    {% when "complete" %}
+        {% assign status_desc = page.t.status.reported_online %}
+{% endcase %}
+
+<span class="status {{ status_css }}">
+    {{ status_desc }}
+</span>
--- a/_layouts/goal-by-target-vertical.html
+++ b/_layouts/goal-by-target-vertical.html
@@ -2,80 +2,73 @@
 {% include header.html %}
 
 <div class="container">
-  {% include components/goal/breadcrumbs.html %}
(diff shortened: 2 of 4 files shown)


brockfanning:
@EFT-Defra FYI, there is something weird going on with the tests right now - as of yesterday. So those failures are not related to your PR.

I like that case syntax, thanks! Actually I'm wondering if you could take this a step further, and abstract it out into an include file, possibly `_includes/components/reportingstatus/status_label.html`? Then in the goal layouts, instead of `{{ status_desc }}` it could be `{% include components/reportingstatus/status_label.html indicator=indicator %}`. That way we're passing the `indicator` variable to that shared include file, reducing code duplication, and also making it easier for users to override that logic if needed.

EFT-Defra:
I was thinking the same thing, I'll have a look and update the PR.

_includes/components/reportingstatus/reporting-status-label.html
```diff
@@ -0,0 +1,15 @@
+{% assign status_css = indicator.reporting_status | slugify %}
```
brockfanning, line 1: @EFT-Defra Here, instead of "indicator" you should use "include.indicator". It probably works either way, but it's a good practice to avoid using shared global variables when possible.

brockfanning (changes requested):
@EFT-Defra This is going to be a welcome improvement. I only noticed one specific change to make (the "include.indicator" thing). Otherwise a general request though: would it be possible to update the PR to include only the minimum changes needed? This PR includes quite a bit of unrelated changes, like whitespace changes and rearranged html attributes. Thank you!

EFT-Defra:
@brockfanning apologies for all the unrelated changes, I'm using [this Liquid extension for VSCode](https://marketplace.visualstudio.com/items?itemName=sissel.shopify-liquid) for auto-formatting and I didn't think before committing...

brockfanning:
@EFT-Defra I just re-ran the tests and they passed this time. That search test is flaky and sometimes fails randomly.

brockfanning:
@EFT-Defra I wasn't sure if you were still working on removing the unneeded changes - just in case let me elaborate on the reason it's important: Many users override certain files to customize them, by putting a copy of the file in their site repository. When we release a new version of Open SDG, sometimes we change files that one or more users have made copies of. In these cases, the user needs to look at what changed in our version and then (ideally) make the same change in their copy. So, even though it's tempting to fix indentations (and things like that) it's better for the users if we make our changes as small as possible.

It might be easier to make a new PR and copy over only what needs to change (I do that a lot) and that's fine too.

EFT-Defra:
@brockfanning Okay, that makes sense. I've removed the unnecessary changes from this PR.

brockfanning:
Perfect, thank you for the contribution @EFT-Defra!

brockfanning (approved):

PR #1135, opened 2021-03-07
Pin to Ubuntu 18.04 because 20.04 doesn't work for tests

brockfanning:
This is needed because Github Actions is migrating to Ubuntu 20.04, but our test suite fails in those cases with random "Page crashed" errors from Puppeteer.


brockfanning:
@jwestw @LucyGwilliamAdmin No testing needed here, just a tweak to allow tests to pass.

jwestw (approved):
No need for testing.

PR #1136, opened 2021-03-07
Pin to Ubuntu 18.04 because 20.04 doesn't work for tests

brockfanning:
Another version of #1135 to go into 1.3.0-dev.


brockfanning:
@LucyGwilliamAdmin @jwestw No testing needed here, just a tweak to allow tests to pass.

jwestw (approved):
No need for testing.

PR #1139, opened 2021-03-08
1.3.0 backwards compatibility fixes

brockfanning:
This is a few scattered backwards compatibility additions to avoid errors that may happen when users upgrade to 1.3.0 while a changed file has been overridden.

Q | A
--- | ---
Related version | 1.3.0-dev
Includes tests? | NA
Updated docs? | NA
Updated config forms? |NA
Added CHANGELOG entry? |NA

### DIFF
--- a/_includes/assets/js/indicatorController.js
+++ b/_includes/assets/js/indicatorController.js
@@ -1,3 +1,8 @@
+// @deprecated start
+// Some backwards compatibiliy code after Lodash migration.
+_.findWhere = _.find;
+// @deprecated end
+
 var indicatorController = function (model, view) {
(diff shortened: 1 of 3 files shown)


brockfanning:
@jwestw I don't think that it's worth doing any testing with this. All this does is add some backwards-compatibility in case a user has overridden particular files. I've already done that testing locally (hence this PR) so in my opinion this can be merged.

jwestw:
Hey Brock. On your advice I'll pass this. Would there be anything to look out for when testing 1.3?

jwestw (approved):
Approving based on last comment.

PR #1145, opened 2021-03-09
Chart color tweaks and config option

brockfanning:
Q | A
--- | ---
Feature branch/test site URL | [Link](add link here)
Fixed issues | Fixes #1138 
Related version | 1.3.0-dev
Bugfix, feature or docs? | Bugfix
Keeps backward-compatibility? |✔️
Includes tests? |NA
Updated docs? |✔️
Updated config forms? |✔️
Added CHANGELOG entry? |NA

In addition to tweaking the colors, this also makes the headline colors configurable (they were hardcoded before).

PR #1147, opened 2021-03-10
Chart color tweaks 1 3 0 correct branch

brockfanning:
I accidentally merged #1145 into 1.4.0-dev, but we need this for 1.3.0-dev.


brockfanning:
@jwestw This is identical to #1145 so does not need testing.

PR #1157, opened 2021-03-17
Allow null values in data

### LINKED ISSUE #1154: Data gaps not shown in chart and table
Site repository URL: https://github.com/SdgTestEnvironment
Open SDG version: 1.3.0

**Describe the bug**
Gaps in time series are not shown (neither in the chart nor in the table), although there is a row with empty value column for these years in the csv file.

**To Reproduce**
1. Go to [https://sdgtestenvironment.github.io/sdg-indicators/3-d-1/](https://sdgtestenvironment.github.io/sdg-indicators/3-d-1/)
2. See missing years 2013, 2015 & 2016

**Expected behavior**
I would expect the chart and table to show all years listed in the csv file (including years with no values before the first or after the last value in the series - we are thinking of having all charts start at 2010, whether there are values or not).

Appearence of No-Value-Years:
For a line chart I would expect to show no Marker but a line spaned from previous to next value.
For a bar chart I would expect to show a gap in between the other bars
For the table I would expect to show a "-"

**Screenshots**

grafik

**Additional context**
After pointing to 1.3.0 it took me some time to notice this. I hope it is not caused by any overwrite I did untill then.

brockfanning:
Q | A
--- | ---
Feature branch/test site URL | [Link](http://sdgdev-813006012.eu-west-1.elb.amazonaws.com/hotfix-allow-empty-values/)
Fixed issues | Fixes #1154 
Related version | 1.3.0
Bugfix, feature or docs? | bug
Keeps backward-compatibility? |yes
Includes tests? |yes
Updated docs? |na
Updated config forms? |na
Added CHANGELOG entry? |na

This is meant to address a regression bug, so it's against the main branch and should be released as a hotfix.

This PR seems to fix the problem, but I'm not sure why this regression happened, because the code I've changed here didn't change in the latest version. So this needs thorough testing to make sure there are no side effects. I'll keep looking into what actually caused the regression.


brockfanning:
After looking more into it, I can see why the regression happened. It has to do with the place where the available years are calculated. I feel that the solution in this PR is best, but still think it needs testing to make sure there are no side effects.

brockfanning:
@LucyGwilliamAdmin @jwestw Another regression bug here, so this would get merged to the master branch and released in 1.3.1. The main change here is that we no longer filter out data rows without values. I don't know why we did that in the first place - it actually seems kind of unnecessary. Maybe there was a time when null values caused an issue. Or maybe they still do cause an issue in some edge-case that I'm forgetting about. Because of this uncertainty, I would recommend testing this against a wide variety of data situations among the UK data.

This also adds an automated test to make sure this regression doesn't happen again in the future.

jwestw:
[Feature branch](http://sdgdev-813006012.eu-west-1.elb.amazonaws.com/hotfix-allow-empty-values) has been created.

brockfanning:
I did some spot checks in goals 1-3 and didn't notice any issues.

brockfanning:
@MoBosse This pull-request should resolve the issue with empty values that was introduced in 1.3. (Just in case you have time to give it a test.)

jwestw:
@LucyGwilliamAdmin I am wondering if there are any indicators with Null values in. Or shall I create a data branch and create some fake data?

LucyGwilliamAdmin:
@jwestw I think there probably is but I don't know of any off the top of my head - it's probably quicker to create some fake data as it's difficult to know without going into a lot of data files

jwestw:
I made some [Null value data here](https://github.com/ONSdigital/sdg-data/blob/hotfix-allow-empty-values/data/indicator_1-1-1.csv). 2011 and 2013 should be Null but [1-1-1 ](http://sdgdev-813006012.eu-west-1.elb.amazonaws.com/hotfix-allow-empty-values/1-1-1/) is still showing values for these. The indicator branch is pointing at the data branch. What's going on?

LucyGwilliamAdmin:
@jwestw it looks like the data is still displaying in data build so might be worth forcing a rebuild of that data branch:
http://sdgdev-813006012.eu-west-1.elb.amazonaws.com/data/hotfix-allow-empty-values/en/data/1-1-1.json

jwestw:
The rebuild worked. 

image

The years with null data are still on the x-axis but data points are not plotted.

jwestw (approved):
Fixed

PR #1162, opened 2021-03-21
Tutorial for creating pages

brockfanning:
Q | A
--- | ---
Fixed issues | NA
Related version | 1.4.0-dev
Bugfix, feature or docs? | docs
Keeps backward-compatibility? |NA
Includes tests? |NA
Updated docs? |✔️
Updated config forms? |NA
Added CHANGELOG entry? |NA

### DIFF
--- a/docs/tutorials/create-pages.md
+++ b/docs/tutorials/create-pages.md
@@ -158,7 +158,7 @@ This step is not required, if you would like to add a link to the new page in yo
 Let's go over what these fields do:
 
 * **path**: This is the path for the page. This should correspond to the "permalink" we chose in the original page's "front matter".
-* **translation_key**: This is the label of the menu item. As its name implies, it can be a "translation key". In the example above, for simplicity we just used "My menu item". In a multilingual implementation, we would certainly want to use a translation key (eg, "custom.my-example-page") instead.
+* **translation_key**: This is the label of the menu item. As its name implies, it can be a "translation key" (for more details on translation keys, see the [glossary page in the "translation keys" section](../glossary.md#translation-keys)). In the example above, for simplicity we just used "My menu item". However, in a multilingual implementation, we would certainly want to use a translation key. For example, if you have a translation file called custom.yml with an item in it called "example-menu-label", then you would put "custom.example-menu-label" here.
 
 Now we can add this change to our branch.
 
--- a/docs/tutorials/create-pages.md
+++ b/docs/tutorials/create-pages.md
@@ -17,6 +17,16 @@ This tutorial is mostly about content, but does involve the editing of some conf
 
 In practice, you will need to come up with the title and content for the new page. This may involve some content work, but for now we will just use example text.
 
+## Start a branch
+
+Like many tutorials, we will be changing multiple files. So it is best to create a branch first.
+
+1. In your site repository on Github.com, open the dropdown in the top left which says "develop".
+2. Type "create-pages-tutorial"
+3. In the dropdown select "Create branch: create-pages-tutorial from 'develop'".
+
+This creates a branch called "create-pages-tutorial" and switches to that branch. We will be working on this branch for the rest of the tutorial.
+
 ## Start the new page
 
 1. In your site repository on Github.com, go into the `_pages` folder.
@@ -77,22 +87,8 @@ layout: page
 This is my page content.
 ```
 
-1. Towards the bottom, select "Create a new branch for this commit and start a pull request."
-1. **Change the name of the branch to `create-pages-tutorial`. You will use this same branch name for all changes in this tutorial.**
-1. Beneath this, click "Propose new file".
-1. Change the name of the pull request to "Create pages tutorial".
-1. Click on the green "Create pull request" button.
-
-> In contrast to some tutorials, **do not "merge" this pull request.** You will be adding more changes to the pull-request before merging it.
->
-> Before going further, switch to your newly-created branch:
->
-> 1. Go to the root of your site repository.
-> 1. Above the list of files, press the drop-down that says `develop`, and select `create-pages-tutorial`.
->
-> This *switches* you to the `create-pages-tutorial` branch, which you will use for the remainder of the tutorial.
->
-> Think of this process as bundling several related changes together.
+1. At the bottom select "Commit directly to the `create-pages-tutorial` branch".
+1. Click the "Commit changes" button.
 
 ## Multilingual: translated versions
 
@@ -104,7 +100,8 @@ In Open SDG, the expectation is that every page will be translated into every la
 
 This is nearly identical to the previous step, but we will put the file in a subfolder for Spanish pages: "es".
 
-1. In your site repository on Github.com, (still in the `create-pages-tutorial` branch) go into the `_pages` folder.
+1. In your site repository on Github.com, make sure you are still in the `create-pages-tutorial` branch. If not, select in the dropdown on the top left.
+1. Go into the `_pages` folder.
 1. Press "Add file" and then "Create new file".
 1. Name your file "es/example.md". (The "es/" is what creates the subfolder.)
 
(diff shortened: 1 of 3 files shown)


docs/tutorials/create-pages.md
```diff
@@ -0,0 +1,178 @@
...
+
+This creates a branch called "create-pages-tutorial" and switches to that branch. We will be working on this branch for the rest of the tutorial.
+
+## Start the new page
+
+1. In your site repository on Github.com, go into the `_pages` folder.
+2. Press "Add file" and then "Create new file".
+3. Name your file "example.md". Or substitute any other name, as long as it ends with ".md". This extension stands for [Markdown](https://daringfireball.net/projects/markdown/), which will be discussed in more detail below.
+
+## Add the page's "front matter"
```
jwestw, line 36: May want to either define “front matter” immediately or avoid using the term initially.

docs/tutorials/create-pages.md
```diff
@@ -0,0 +1,178 @@
...
+3. In the dropdown select "Create branch: create-pages-tutorial from 'develop'".
+
+This creates a branch called "create-pages-tutorial" and switches to that branch. We will be working on this branch for the rest of the tutorial.
+
+## Start the new page
+
+1. In your site repository on Github.com, go into the `_pages` folder.
+2. Press "Add file" and then "Create new file".
+3. Name your file "example.md". Or substitute any other name, as long as it ends with ".md". This extension stands for [Markdown](https://daringfireball.net/projects/markdown/), which will be discussed in more detail below.
+
```
jwestw, line 35: Potentially add an instruction here: "After writing the filename press enter or click out of the box. The editor is open below in which you can open edit the content of the file."

docs/tutorials/create-pages.md
```diff
@@ -0,0 +1,178 @@
...
+## Topics covered
+
+* Jekyll pages
+* Markdown
+* Multilingual pages
+* Menu configuration
+
+## Level of difficulty
+
+This tutorial is mostly about content, but does involve the editing of some configuration files. Anyone can follow the steps to complete the tutorial, however you will be using the YAML syntax involved in most Open SDG configurations. YAML requires particular indentation and spacing, so some attention to detail will be needed.
```
jwestw, line 14: Would it be a good idea to link to some examples of yaml syntax, [like this](https://blog.stackpath.com/yaml/)? Or would that be over complicating things?

jwestw (commented):
Hi @LucyGwilliamAdmin I think I have seen similar errors to this last time I was working on a language branch but I can't remember why this was happening. Please could you take a look at the test output? https://github.com/ONSdigital/sdg-indicators/runs/2166174879?check_suite_focus=true

LucyGwilliamAdmin:
@jwestw re. the error, compare what you've put in the front matter of the new files compared with what's in the front matter of the existing files

jwestw:
@brockfanning 

`custom.my-example.page` is showing up in menu. Do you know the exact term I should use [here](https://github.com/ONSdigital/sdg-indicators/blob/cccf0f24bdfe6d078168ce4ce10ead3bd15a4bf9/_config.yml#L140)? 

image

docs/tutorials/create-pages.md
```diff
@@ -0,0 +1,181 @@
...
+
+This step is not required, if you would like to add a link to the new page in your main menu, here are the steps:
+
+1. In the root of your site repository on Github.com, make sure you are still in the `create-pages-tutorial` branch. If not, select it from the dropdown on the top left.
+1. Click on `_config.yml`.
+1. Click the pencil icon on the right to begin editing the file.
+1. Look for the `menu` section.
+1. Add an item at the bottom of the list, using the same syntax and indentation as the others:
+
+        - path: my-example-page
```
jwestw, line 155: For the UK site, this had to be `  - path: /my-example-page/` (with slashes).

brockfanning:
@jwestw I added a bit more detail about the translation keys - does it help?

I didn't think the slashes in the path were necessary - did you see some problems in the browser?

LucyGwilliamAdmin:
@brockfanning we've been having issues with slashes for a while - if not like that, we seem to get errors in build:

https://github.com/ONSdigital/sdg-indicators/runs/2166504766?check_suite_focus=true#step:5:8

Although I'm sure I've seen code before that's supposed to deal with that

jwestw:
> @jwestw I added a bit more detail about the translation keys - does it help?
> 
> I didn't think the slashes in the path were necessary - did you see some problems in the browser?

I actually didn't try without, but I know **with** slashes worked. And previously I had a series of failed tests from not including slashes in the permalinks in front matter.

jwestw:
I used `menu.guidance` as the key so I didn't have to create another one. It works.

## Russian

image

##  French

image

jwestw:
Thank you for the extra guidance and the changes you have made so far, @brockfanning. I am happy with this tutorial so will pass it now.

jwestw (approved):
I think all the changes have been made. Thanks

brockfanning:
@LucyGwilliamAdmin @jwestw We may as well get the slash thing right. Questions I'm wondering about are:
1. Is the preceding slash necessary? Eg, "/my-path"
2. Is the trailing slash necessary? Eg, "my-path/"

My impression of that failing test is that it was due to having trailing slashes in FR and RU but not in EN. But I could be wrong. Anyway I think we should probably pick an approach and stick with it, for both permalinks and menu item paths, for consistency. From your experience it sounds like we should recommend always a preceding _and_ a trailing slash?

LucyGwilliamAdmin:
@jwestw can you investigate the points Brock has made please?

jwestw:
- Test with [proceeding slash only](https://github.com/ONSdigital/sdg-indicators/commit/3f396e406fa4e043fc51dfa22dcbcdc43ec49467): https://github.com/ONSdigital/sdg-indicators/actions/runs/680268927
- Test with [trailing slash only](https://github.com/ONSdigital/sdg-indicators/commit/d1afaa5d6c23476b395932523b4d9491fabd0379): https://github.com/ONSdigital/sdg-indicators/actions/runs/680283826
- Test with [no slashes](https://github.com/ONSdigital/sdg-indicators/commit/33f03d532577d9b0fb1e24929714acb6f74d6b92): https://github.com/ONSdigital/sdg-indicators/actions/runs/680298027
- Test with [mixture of slashes](https://github.com/ONSdigital/sdg-indicators/commit/d02be9975886315bdbd12b92a902171eff95953c) : https://github.com/ONSdigital/sdg-indicators/actions/runs/680303150

brockfanning:
@jwestw It might be helpful to extend the tests to play around with the "permalink" of pages also. My thinking is that there would be four possibilities to test. In each, the same change is made to both...

1. [The "path" for the About page in the menu](https://github.com/ONSdigital/sdg-indicators/blob/develop/_config.yml#L122)
2. And [the "permalink" for the About page itself](https://github.com/ONSdigital/sdg-indicators/blob/develop/_pages/about.md)

The four possibilities would be:

1. No slashes (eg, `path: about` and `permalink: about`)
2. Preceding slash (eg, `path: /about` and `permalink: /about`)
3. Trailing slash (eg, `path: about/` and `permalink: about/`)
4. Both slashes (eg, `path: /about/` and `permalink: /about/`)

It's possible that all four will work equally well, but it would be great to know if any of them *don't* work. Would it be possible to have a feature branch for each? I ask because, even if the tests pass, it still would be nice to try it in a browser.

jwestw:
Hey @brockfanning 

These are the results of the various permutations of slashes. Hopefully my commit messages are clear enough to know which is which.  

image

The links to the two that failed:

https://github.com/ONSdigital/sdg-indicators/commit/0ada36ac25fd2c8d6206a701a8faefda6850d461
https://github.com/ONSdigital/sdg-indicators/commit/80b7a13c87fd34dacc282df6eea012784f030331

brockfanning:
@jwestw Nice! So if I'm interpreting correctly, the trailing slash is always required but the preceding slash is not required?

brockfanning:
Brock to update this PR standardizing around the trailing slash, and look for other spots in the documentation that need changing.

PR #1164, opened 2021-03-23
Tutorial for using Weblate

brockfanning:


brockfanning:
@LucyGwilliamAdmin @jwestw For this PR I made it again "master" so that we could get this into the live documentation earlier. I was thinking this would be useful for both Lao and Vanuatu, who are both about the translate new languages.

LucyGwilliamAdmin:
@jwestw @brockfanning I will take a look at this one before I finish today

LucyGwilliamAdmin (approved):
@brockfanning I've just tested out following guidance with non-admin account. It works really well and I think it's clear.

I think once one of the countries need to start doing this we should ask them for feedback and maybe iterate but think fine to merge for now

PR #1168, opened 2021-03-29
Support for graph_title_from_series setting

### LINKED ISSUE #1167: Graph title by series

When importing data from SDMX sources, the graph titles always need to be specified per series, according to the series name. This is possible using the "graph_titles" indicator config setting, but it is a lot of manual work.

We need a site-wide config setting that will automatically set the graph title according to the selected series.

To reduce the manual work needed in getting accurate graph titles.

brockfanning:
Q | A
--- | ---
Fixed issues | Fixes #1167 
Related version | 1.4.0-dev
Includes tests? |✔️
Updated docs? |❌
Updated config forms? |✔️
Added CHANGELOG entry? |❌

### DIFF
--- a/_includes/assets/js/indicatorModel.js
+++ b/_includes/assets/js/indicatorModel.js
@@ -57,6 +57,9 @@ var indicatorModel = function (options) {
 
   this.refreshSeries = function() {
     if (this.hasSerieses) {
+      if (helpers.GRAPH_TITLE_FROM_SERIES) {
+        this.chartTitle = this.selectedSeries;
+      }
       this.data = helpers.getDataBySeries(this.allData, this.selectedSeries);
       this.years = helpers.getUniqueValuesByProperty(helpers.YEAR_COLUMN, this.data);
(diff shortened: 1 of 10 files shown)


LucyGwilliamAdmin:
A question has just come to mind re. this - does this only work if series_toggle: true

Also I can't remember if we discussed this before but should we actually make series_toggle true by default?

brockfanning:
Yes, it depends on that being true. And personally I would be in support of making it true by default, since it is not that hard for countries to set it to false if needed.

LucyGwilliamAdmin:
I also think we should make true by default when possible

I also just found this [other comment](https://github.com/open-sdg/sdg-build/pull/177#issuecomment-766406129) which reminded me that series in dropdowns don't get translated. I still need to speak to Kyrgyzstan about that - I'm just about to email now. I'm not sure if they're going to want to change to toggle though since they are who reuqested dropdown

brockfanning:
Just a tech note, when we decide to make it default to true, that change can be made here: https://github.com/open-sdg/open-sdg/blob/1.4.0-dev/_includes/assets/js/model/constants.js#L11

Issue #755, opened 2020-07-02
Rewrite the contrast toggles in plain HTML


SavvasStephanides:
Currently the high contrast toggles are being [built dynamically with Javascript](https://github.com/open-sdg/open-sdg/blob/1b7e787c736e8516cad1cd1ea8e48994b28b29af/_includes/assets/js/accessibility.js). This has worked for the most part but it can cause (and has caused) quite a few issues:

* if there's a Javascript bug anywhere in the site, it can cause the rest of the script to crash. This can result in the contrast switcher to not show up at all.
* Building the toggles in Javascript makes them unstable, such as the case of the [toggle disappearing completely in IE11](https://github.com/open-sdg/open-sdg/issues/749). Rewriting the toggles in basic HTML (and just leave the functionality to Javascript only) will make them a lot more stable and lot more compatible (the more basic the better). 
* Overall good practice to have as much markup as possible where it belongs: HTML

I guess my question at this point that I'm trying to figure out is: why was this built with Javascript to begin with? Is there some kind of feature that we can't achieve with basic HTML, or just couldn't at the time of developing?

Nonetheless, I think we will benefit greatly with a rewrite.

brockfanning:
My guess is that because it was done in javascript mainly because it uses cookies to remember the user's choice on the next page load. It also needs to change some image src attributes. I'd bet that we could get it more in HTML and less javascript, but I don't think we'll be able to get to 100% HTML.

SavvasStephanides:
@brockfanning My suggestion is to move as much of the layout logic as possible to HTML/CSS. I understand there's some dynamic Javascript functionality going on which is necessary, such as getting and setting the contrast cookie. What I am suggesting is the basic layout of the toggles to be made in HTML and leave the cookie/toggling logic to Javascript.

It is also common practice to to make dynamic page styling changes using CSS, using attributes:

```scss
#container{
    background-color: #FFFFFF;
}

body.contrast-high{
    #container{
        background-color: #000000;
    }
}
```

We can then use Javascript to toggle the body class name and CSS will do the rest.

```js
document.body.classList.add("contrast-high")
```

And obviously yes, we will need Javascript to read the user's cookie to set the `body` class name.

I'm guessing setting the `src` of images is a bit more involved, but that should be fine.

LucyGwilliamAdmin:
@brockfanning can this be closed due to your changes in #1171?

brockfanning:
Yes, I'll close this now.

PR #1171, opened 2021-04-07
Single option for contrast toggle

### LINKED ISSUE #1059: Convert enable high contrast 'A' to a single toggle icon
On desktop we currently have two 'A' icons. One to 'Enable high contrast' and one to 'Disable high contrast'. We do not require a  'disable high contrast' option on desktop view when you are not on the high contrast version (and visa versa).

Recommendation is to only have a single 'A' icon that works like a toggle between the two states, working the same way as when in mobile view (and copying the visual style too).

Additionally I have also noticed that when viewed on desktop the 'A' on the icons is perfectly centred in the circle. However when viewed on mobile the 'A' seems to increase in size and therefore no longer sits in the centre - could this be looked at please.

We have put the above to DAC who had the following feedback which I believe we will have addressed with this implementation.

Here’s DAC’s response:

_I understand what you are saying about the buttons where there are only two different style sheets. From an accessibility POV I don’t see there being any problems with only having the button available which switches to the state that is not already selected. However, it would very much depend on the implementation and there are a few accessibility points to consider – for example, if you are going to switch buttons as soon as a button is activated, will this be a different button? If so, where will focus go? If it is the same button but with a different purpose, how will it be labelled? Also, what would you do if you wanted to add additional style sheets to the page? 

The best option would be to have a toggle button, but toggles are not enormously different to what you already have really. I would say that a toggle implementation may be clearer for users reliant on audio feedback where the options are mutually exclusive, but that the current implementation works ok too._

brockfanning:
Q | A
--- | ---
Feature branch/test site URL | [Link](https://lucygwilliamadmin.github.io/sdg-indicators/)
Fixed issues | Fixes #1059 and #755
Related version | 1.4.0-dev
Keeps backward-compatibility? |(mostly)
Includes tests? |❌
Updated docs? |✔️
Updated config forms? |✔️
Added CHANGELOG entry? | ✔️ 

About the backwards compatibility, I think there may be a problem if an implementation has overridden the headers and is using the "long" contrast type. Though, I don't know of any implementations in that situation, so maybe it's not worth the trouble of avoiding that.

Beyond that issue, I believe this does a pretty good job of maintaining the HTML and CSS that was already being used. But definitely testing is needed.

The documentation update details the config setting necessary to test this.

To set this option with config forms, this PR is needed: https://github.com/open-sdg/jekyll-open-sdg-plugins/pull/95


LucyGwilliamAdmin:
@brockfanning I seem to be getting html errors;
https://github.com/ONSdigital/sdg-indicators/runs/2341922290?check_suite_focus=true

LucyGwilliamAdmin:
@brockfanning @phillipgardiner I've just added a link for this

phillipgardiner:
@LucyGwilliamAdmin @brockfanning 

Thanks for this, two things

1. I noticed that when viewed on desktop the 'A' on the icons is perfectly centred in the circle. However when viewed on mobile the 'A' seems to increase in size and therefore no longer sits in the centre - could this be looked at please?
2. When you select the contrast toggle it stays highlighted and thus looks like a square until you click elsewhere on the screen - I'm guessing this is the default behaviour?

LucyGwilliamAdmin:
@phillipgardiner I noticed 2. too but looked at live site and it seems to happen currently so wasn't sure if it was something to add at this point

brockfanning:
@LucyGwilliamAdmin I just pushed a change, so this is ready for a refresh.

To deal with the issues above I've tried making the focus treatment of this toggle stay circular and have a yellow outline. I also lowered the font size of the toggle on mobile, to try and help with the alignment issue. Hopefully it works - I remember this being a really tricky issue because of differences in browsers and configurations.

phillipgardiner:
@brockfanning Looks a lot better, thanks Brock

LucyGwilliamAdmin (approved):

PR #1199, opened 2021-05-09
Styles for the global/national metadata

brockfanning:
Q | A
--- | ---
Fixed issues | NA
Related version | 1.4.0-dev
Includes tests? |NA
Updated docs? |NA
Updated config forms? |NA
Added CHANGELOG entry? | ✔️ 

This PR adds some default styling specifically to the "Global" and "National" metadata tabs. The rules are:

1. Any H1 that is the very first element in a metadata field will be hidden
2. Any H2 that is immediately following an H1 (which itself is the very first element) will have no top margin.

We could conceivably mess with the font size of these but I thought I'd start with a small change.


brockfanning:
Here is what this is looking like now:

screenshot-127 0 0 1_4000-2021 05 17-11_17_02

That is produced by this markdown:

```
# This is an H1 and should be hidden

## This is an H2 and should have no top margin.

This is a paragraph.

## This is a normal H2.

This is another paragraph.

# This is a normal H1.

This is the last paragraph.
```

LucyGwilliamAdmin (approved):

PR #1202, opened 2021-05-10
Changes for SDMX metadata concepts and indicator config

### LINKED ISSUE #1201: Docs for new indicator config/metadata structure
Split [Metadata format](https://open-sdg.readthedocs.io/en/latest/metadata-format/) into metadata and indicator config

LucyGwilliamAdmin:
Q | A
--- | ---
Feature branch/test site URL | N/A
Fixed issues | Fixes #1201 (partly)
Related version | 1.4.0-dev
Bugfix, feature or docs? | Docs
Keeps backward-compatibility? |N/A
Includes tests? |N/A
Updated docs? |N/A
Updated config forms? |❌/✔️
Added CHANGELOG entry? |❌/✔️


LucyGwilliamAdmin:
@brockfanning I've started this ready for when we start using the new starter repos but let me know if you have any feedback so far.

A couple of notes:

1. For the definitions, so far I have included both global reporting and national reporting, but I'm thinking we remove the one for global reporting as I'm thinking this will be specific to national metadata fields
2. Where in the definition is says not applicable for country/national reporting the fields should be removed completely from this page and the 1.1.1 template file in data starter (and the national versions removed from prose)
3. Wondering if we should also include the fields for the main concepts e.g. under all of the detailed concepts for 0, could have something like:

  > If using the main concepts collate the information for these fields in a single field named `SDG_INDICATOR_INFO`

LucyGwilliamAdmin:
@brockfanning do you have any thoughts on my comments above?

brockfanning:
@LucyGwilliamAdmin
1. I don't fully understand this note, but in general I do think that the metadata maintained in Open SDG should only be national, and (eventually) we will be pulling global metadata from a single remote source (the UN metadata API).
2. That sounds right, given that we want to limit this metadata to national.
3. That seems like a good approach to me.

Another note - I know this will be a lot of work, but would it be possible to have the indicator configuration mimic the structure of the site configuration, by having headings for each setting? The benefit of having a heading is that we can link to directly to it, which will allow us to link to the individual settings from the indicator config form, like we do on the site config form.

LucyGwilliamAdmin:
@brockfanning what is the purpose of the `label` property in [metadata_schema.yml](https://github.com/open-sdg/open-sdg-data-starter/blob/1.4.0-dev/metadata_schema.yml) now that prose is not being used?

brockfanning:
@LucyGwilliamAdmin It's not serving a vital purpose now, but we do have some code that refers to it.

One example is [here](https://github.com/open-sdg/jekyll-open-sdg-plugins/blob/1.4.0-dev/lib/jekyll-open-sdg-plugins/metadata_schema_to_config.rb#L53-L54) which (like you say) uses it as a fallback for labels on the metadata config form.

The bigger issue is [this code](https://github.com/open-sdg/sdg-build/blob/1.4.0-dev/sdg/schemas/SchemaInputOpenSdg.py#L77-L88) which assumes it will be there. I suspect that if you remove it, it might actually cause an error during the build.

LucyGwilliamAdmin:
@brockfanning thanks for clarifying - re. restructure indicator configuration, should I make it exactly same as site i.e. in alphabetical order and then specify whether each one is optional or mandatory?

brockfanning:
That sounds good to me, if it's not too much trouble.

LucyGwilliamAdmin:
@brockfanning I think these are ready for review - I ended up taking a little bit of info about the main concepts due to the info you got from UNSD about them strongly advising against countries using them.

docs/indicator-configuration.md
```diff
@@ -0,0 +1,470 @@
...
+
+### Indicator configuration forms
+To help with configuring indicators, you can optionally enable user-friendly indicator configuration forms, which will allow you to fill in a form and then download a YML file which will be ready to upload to your platform.
+
+For more details on how to enable these forms, see the [indicator_config_form site configuration setting](configuration.md#indicator_config_form).
+
+## Note about translation keys
+Indicator configuration values can either be filled in with normal text ("My field value") or with [translation keys](glossary.md#translation-keys) (my_translations.my_translation). In the examples below, we will try to demonstrate both possibilities.
+
+As an optional shorthand, if the translation key is in the `data` group, then the group can be omitted. For example, the translation key `data.female` can be written as simply `female`.
```
brockfanning, line 15: Personally I think we should remove this part about the special handling of the `data` group. This is probably something we want to eventually remove, and I don't think anyone is relying on this shorthand, so it's best to start keeping it out of documentation.

docs/indicator-configuration.md
```diff
@@ -0,0 +1,470 @@
...
+
+**_Optional_** This setting is used when importing data from SDMX to specify a more useful label for the COMPOSITE_BREAKDOWN column (if present). Translation keys are supported, as always. Using the example below would change the COMPOSITE_BREAKDOWN label to "Hazard type" for this indicator:
+
+```nohighlight
+composite_breakdown_label: Hazard type
+```
+
+### computation_units
+
+**_Recommended_**: This setting is used to specify the unit used in the headline data for the indicator.
```
brockfanning, line 66: Do you think we should explicitly say where it will appear? In this case, the value will appear in the chart/table footers (just like Copyright below).

docs/indicator-configuration.md
```diff
@@ -0,0 +1,470 @@
...
+
+### data_not_statistical
+
+**_Required_**: This setting is used to specify whether the indicator is statistical (can be charted/graphed) or not. If you have uploaded data that you want to display on a chart/table for an indicator, set this to `false`.
+
+```nohighlight
+data_non_statistical: false
+```
+
+You may want to display some very important information which site viewers must keep in mind when using the data provided. To display a notice above the graph in a coloured box, you can use the following fields within the metadata file.
```
brockfanning, line 116: I think this needs to have a heading above it, like maybe: `### Data notice settings`. Or optionally it could be moved below `### data_notice_class` since that is the first one.

docs/indicator-configuration.md
```diff
@@ -0,0 +1,470 @@
...
+
+```nohighlight
+data_start_values:
+  - field: Sex
+    value: Females
+  - field: Age
+    value: 15 to 19
+```
+
+You may want to add an additional feature which isn't created from data, such as an iframe. You can create an extra tab to display this feature by adding the following fields to the metadata file. You can either specify a URL (use `embedded_feature_url`) or some HTML (use `embedded_feature_html`) for the feature you want to embed.
```
brockfanning, line 179: Similar to data notices, maybe this needs a heading above it like `### Embedded feature settings`?

brockfanning (commented):
@LucyGwilliamAdmin Awesome work! I added a few comments on the indicator config part.

brockfanning:
I think this may also need the latest code from 1.4.0-dev to pass tests.

LucyGwilliamAdmin:
@brockfanning do you think this should go under the 'Data providers' section?

brockfanning:
@LucyGwilliamAdmin I think that makes sense for the metadata, and the indicator config could go under the 'Developers' section?

LucyGwilliamAdmin:
@brockfanning yeah, that sounds like a good idea!

LucyGwilliamAdmin:
@brockfanning do you have anymore feedback for this one?

docs/metadata-format.md
```diff
@@ -1,379 +1,282 @@
-<h1>Metadata format</h1>
+<<h1>Metadata format</h1>
```
brockfanning, line 1: There's an extra `<` here.

docs/metadata-format.md
```diff
@@ -1,379 +1,282 @@
...
 
-In your [data repository](glossary.md#data-repository) the metadata is maintained on an indicator-by-indicator basis. This metadata can include any number of custom fields, as defined in a [schema file](https://github.com/open-sdg/open-sdg-data-starter/blob/develop/_prose.yml) (see the "Schema" section below) in your data repository. Some fields, however, are mandatory and/or have specific uses in Open SDG. This page details those fields.
+In your [data repository](glossary.md#data-repository) the metadata is maintained on an indicator-by-indicator basis. 
+
+By default the metadata fields are based on SDG SDMX detailed metadata concepts and the template metadata file in the data starter is based on these fields. It is recommended that you use the default fields as they follow the format that is requested by the UN and allow metadata to compared. However, the metadata can include any number of custom fields, as defined in a [schema file](https://github.com/open-sdg/open-sdg-data-starter/blob/develop/metadata_schema.yml) (see the "Schema" section below) in your data repository. However,
```
brockfanning, line 5: I think the `However,` at the end of this paragraph was a left-over.

docs/metadata-format.md
```diff
@@ -1,379 +1,282 @@
...
+In your [data repository](glossary.md#data-repository) the metadata is maintained on an indicator-by-indicator basis. 
+
+By default the metadata fields are based on SDG SDMX detailed metadata concepts and the template metadata file in the data starter is based on these fields. It is recommended that you use the default fields as they follow the format that is requested by the UN and allow metadata to compared. However, the metadata can include any number of custom fields, as defined in a [schema file](https://github.com/open-sdg/open-sdg-data-starter/blob/develop/metadata_schema.yml) (see the "Schema" section below) in your data repository. However, 
+
+Whilst it is possible to use the SDG SDMX main metadata concepts (which are a level above the detailed concepts) e.g. **0. Indicator information**, the UNSD strongly recommends that the detailed concepts are used.
```
brockfanning, line 7: Maybe give an example of a detail concept, like you did for the `**0. Indicator information**` main concept?

docs/metadata-format.md
```diff
@@ -1,379 +1,282 @@
...
+
+For more details on how to enable these forms, see the [indicator_metadata_form site configuration setting](configuration.md#indicator_metadata_form).
+
+## Note about empty metadata fields
+
+Due to the large number of metadata fields with some maybe not being applicable, empty metadata fields are hidden by default using `hide_empty_metadata: true` in the site configuration file.
+
+If you want metdata fields to show regardless, you should change this option to false. Alternatively, if you want some fields to hide by default but want to purposely not provide any information for other fields, you could populate the fields with some standard text e.g. "Not available"
+
+For more information about this configuration setting see the [hide_empty_metadata configuration option documentation](https://open-sdg.readthedocs.io/en/latest/configuration/#hide_empty_metadata).
```
brockfanning, line 27: Could this use `(configuration.md#hide_empty_metadata)` for the link?

docs/metadata-format.md
```diff
@@ -1,379 +1,282 @@
...
 
 ## Note about translation keys
 
 Metadata values can either be filled in with normal text ("My field value") or with [translation keys](glossary.md#translation-keys) (my_translations.my_translation). In the examples below, we will try to demonstrate both possibilities.
 
 As an optional shorthand, if the translation key is in the `data` group, then the group can be omitted. For example, the translation key `data.female` can be written as simply `female`.
```
brockfanning, line 33: Like with the indicator config, personally I would vote to remove this part about the data group.

brockfanning (commented):
Just a few more comments about the metadata section.

LucyGwilliamAdmin:
@brockfanning thanks for taking a look - any idea what's causing the failures?

brockfanning:
@LucyGwilliamAdmin Ah, I think there are links on other documentation page that are now broken, because of the new structure. It looks like there may be links to the metadata-format page on these other docs pages:
1. changelog.md
2. configuration.md
3. metadata-file-formats.md
4. open-sdg-features.md

LucyGwilliamAdmin:
@brockfanning ok, I've finally got them all - although PR is a little messy now

brockfanning (approved):
@LucyGwilliamAdmin Looks great. If you want to avoid merging all those commits you could do "Squash and merge".

PR #1220, opened 2021-06-04
Square points for headline on line charts

### LINKED ISSUE #1102: Add additional shapes to data points
In order to help distinguish the default data line to the disaggregation data lines displayed the following is proposed:

Default data lines will display a filled square as the data point marker (in the same colour as the lines) Screenshot (142)

Disaggregation data lines will display a filled circle as the data point marker (in the same colour as the lines) Screenshot (143)

brockfanning:
Q | A
--- | ---
Feature branch/test site URL | [Link](https://lucygwilliamadmin.github.io/sdg-indicators/)
Fixed issues | Fixes #1102 
Related version | 1.4.0-dev
Keeps backward-compatibility? |✔️ (technically a change, but an improvement)
Includes tests? | NA
Updated docs? | NA
Updated config forms? | NA
Added CHANGELOG entry? |❌

### DIFF
--- a/_includes/assets/js/indicatorView.js
+++ b/_includes/assets/js/indicatorView.js
@@ -444,6 +444,7 @@ var indicatorView = function (model, options) {
             _.each(chart.data.datasets, function(dataset) {
               text.push('<li>');
               text.push('<span class="swatch' + (dataset.borderDash ? ' dashed' : '') + '" style="background-color: ' + dataset.borderColor + '">');
+              text.push('<span class="swatch-inner' + (dataset.headline ? ' headline' : '') + '" style="background-color: ' + dataset.borderColor + '"></span>');
               text.push('</span>');
               text.push(translations.t(dataset.label));
               text.push('</li>');
--- a/_includes/assets/js/indicatorView.js
+++ b/_includes/assets/js/indicatorView.js
@@ -440,11 +440,11 @@ var indicatorView = function (model, options) {
         legendCallback: function(chart) {
             var text = [];
             text.push('<h5 class="sr-only">' + translations.indicator.plot_legend_description + '</h5>');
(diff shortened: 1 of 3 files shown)


LucyGwilliamAdmin:
@brockfanning the chart changes look good to me. I do have some comments about the legend though but be good to get @phillipgardiner and @AnnCorp thoughts too:

Currently legend looks like this:
image

I'm wondering if looks slightly cleaner if lines in legend were more representative of lines in chart - perhaps like this:
image

I have no idea if that's possible or how much work it is but what are other's thoughts??

Also on bar charts: do we want the legend to stay the same as before so like this:
image

Instead of like this:

image

phillipgardiner:
@LucyGwilliamAdmin @brockfanning @AnnCorp Good work getting to this stage. Yes I would agree with both of your comments @LucyGwilliamAdmin. 

1) Ideally the legend would match the data displayed (e.g. line) as closely as possible so as to support user recognition (no idea how much work it is though).
2) Good spot. Yes there is no need for the points on the bar graph legend as they do not exist on the bar chart.

brockfanning:
@LucyGwilliamAdmin This is ready for a refresh. It should only affect line charts now, and I've tweaked the dimensions of the legend items to look more like the lines on the chart.

LucyGwilliamAdmin:
@brockfanning @phillipgardiner link has been refreshed - do you have any further feedback Phill? It looks good to me

phillipgardiner:
@brockfanning @LucyGwilliamAdmin Looks great to me, brill. I could not find a bar graph one to check that but if you've checked it, its a thumbs up from me :)

LucyGwilliamAdmin (approved):
Line charts and bar charts look good so approving

PR #1222, opened 2021-06-04
Aria attributes to improve accessibility of map search suggestions

### LINKED ISSUE #1193: Improve accessibility of map search bar
Site repository URL: All sites
Open SDG version: All

**Describe the bug**
When users type a search, a list of suggested searches appears. Users can navigate this with the cursor keys; however, screen reader users are not made aware that this list has appeared. Voice activation users found that text dictated into the Search field, the autocorrect options were not triggered, so they were unable to use this control. 

**To Reproduce**
With screen reader on
1. Go to https://sdgdata.gov.uk/3-a-1/
2. Click on the Map tab
3. Click on the Map search icon to activate map
4. Enter 'north east'
5. See error

**Expected behavior**
Screen reader users are able to use Map search result controls

**Additional context**
From accessibility report:
WAI-ARIA roles can be utilised to ensure that all behaviour, information and functionality is exposed to assistive technologies. For more information, please refer to [WAI-ARIA Authoring Practices 1.1: Combobox with Listbox Popup Examples](https://www.w3.org/TR/wai-aria-practices/examples/combobox/aria1.1pattern/listbox-combo.html).

Please note that Dragon dictation does not trigger certain keyboard and mouse events. If the field relies on JavaScript event handlers to see if its populated, these may not be triggered by dictation.

brockfanning:
Q | A
--- | ---
Feature branch/test site URL | [Link](http://uk-sdg-feature-branches.s3-website.eu-west-2.amazonaws.com/feature-site-map-search-listbox-accessibility/3-a-1/)
Fixed issues | Fixes #1193 
Related version | 1.4.0-dev
Includes tests? |❌
Updated docs? |❌/✔️
Updated config forms? |NA
Added CHANGELOG entry? |❌

This does not address the issue mentioned about Dragon dictation. I'm not 100% this meets the requirements, but I think it is ready to be reviewed by DAC.


LucyGwilliamAdmin:
@brockfanning I've just added a link - space bar doesn't seem to be opening the search bar when using keyboard anymore - does on [UK live site](https://sdgdata.gov.uk/3-a-1/) though

brockfanning:
@LucyGwilliamAdmin For me, spacebar doesn't work, but enter does, and the same for the live site. I only tested in Firefox without a screenreader running, though.

LucyGwilliamAdmin:
@brockfanning just had another look, it works when screenreader isn't on but not when screenreader is on. On live site, it works in both cases

brockfanning:
@LucyGwilliamAdmin Ok, I think there is probably an issue with the aria attributes - maybe they are competing with each other. Do you think we could get help from DAC on this?

brockfanning:
@AnnCorp @LucyGwilliamAdmin I think this is the last ticket in 1.4.0, but there may be some issues with it, and we discussed getting help from DAC. Do you think we should start the beta testing of 1.4.0 without this ticket, and then add a "beta2" tag later if DAC approves. Or should wait until DAC approves before starting beta testing of 1.4.0?

AnnCorp:
@LucyGwilliamAdmin happy to take your decision on this one, as you'll be leading on the beta testing and when it's scheduled in.

LucyGwilliamAdmin:
@brockfanning @AnnCorp I think let's continue with beta release and then do another if needed

LucyGwilliamAdmin:
Reply from DAC on this:

I would say that it (the issue) is highly likely that this is due to the fact your ‘Show search’ button is nested within your combobox.

```
<div class="leaflet-control-search leaflet-control" aria-expanded="false" role="combobox" aria-haspopup="listbox" aria-owns="map-search-tooltip-list">
   <label class="search-input" for="searchtext10" style="display: none;"></label><input class="search-input" type="text" size="10" autocomplete="off" autocapitalize="off" placeholder="Search map" id="searchtext10" aria-label="Search map" aria-autocomplete="list" aria-controls="map-search-tooltip-list" style="display: none; max-width: 1025px;">
   <ul class="search-tooltip" id="map-search-tooltip-list" role="listbox" style="display: none;"></ul>
   <a class="search-cancel" href="#" title="Cancel" role="button" aria-label="Cancel" style="display: none;"><i class="fa fa-close" aria-hidden="true"></i></a>
   <a class="search-button" href="#" title="Show search" role="button" aria-label="Show search" aria-expanded="false" style="outline: none;"><i class="fa fa-search" aria-hidden="true"></i></a>
   <div class="search-alert" role="alert" style="display: none;"></div>
</div>

```

Aria roles can be hazardous unless they are implemented exactly as the design system specifies: https://www.w3.org/TR/wai-aria-practices/examples/combobox/aria1.1pattern/listbox-combo.html

Modifying the pattern or adding complexity can cause all sorts of problems to crop up. 

I did notice that there is also some intermittent strange behaviour with the full screen button too. If I press ‘Spacebar’ on the fullscreen button it takes focus to the ‘zoom in’, in fullscreen, if I press ‘Enter’ on the fullscreen button it takes focus to the ‘Zoom in’ button. However, the prior three buttons seem to work as expected. 

Sorry I couldn’t provide much more detail then this – I am not a developer, my expertise centres more around identifying accessibility issues and WCAG non-conformances, so the very technical details will be likely to be a bit beyond my expertise. But I would definitely try removing the button from the combobox and ensuring that the design pattern is followed to the letter.

```
<a class="search-button" href="#" title="Show search" role="button" aria-label="Show search" aria-expanded="false" style="outline: none;"><i class="fa fa-search" aria-hidden="true"></i></a>
<div class="leaflet-control-search leaflet-control" aria-expanded="false" role="combobox" aria-haspopup="listbox" aria-owns="map-search-tooltip-list">[…]</div>
```

brockfanning:
@LucyGwilliamAdmin I've given this another try, to it's ready for review again. FYI I can't recreate that reported issue with the full-screen button.

LucyGwilliamAdmin:
@brockfanning I've just refreshed the feature branch. Now I can expand search using enter, but still not space. However, I'm not being notified by screen reader that the list has appeared.

One thing I noticed is that the combobox doesn't seem to actually have a class - not sure if this is on purpose or may have something to do with it:
image

brockfanning:
@LucyGwilliamAdmin I modeled it after the "2. Listbox with automatic selection" here: https://www.w3.org/TR/wai-aria-practices/examples/combobox/aria1.1pattern/listbox-combo.html

That example doesn't announce any listbox items either, at least for the screenreader I use (XVDA).

About the space bar - it actually works for me in Firefox and Chrome, but only when the screenreader is active.

About the class - yes that was on purpose. Classes don't actually do anything accessibility-wise; they're just for styling or javascript. And since I didn't add any special styling or javascript there was no need to add a class to it.

brockfanning:
@LucyGwilliamAdmin Actually hold off a bit - I just noticed the part about "aria-activedescendant" on that WAI page.

brockfanning:
@LucyGwilliamAdmin This is ready for another refresh/review. Sorry for the trouble, I didn't notice that part about aria-activedescendant before. I think this is what triggers the announcements.

LucyGwilliamAdmin:
@brockfanning I've refreshed. The screenreader is announcing the list now but I;m getting some strange behaviour

1. Go to http://uk-sdg-feature-branches.s3-website.eu-west-2.amazonaws.com/feature-site-map-search-listbox-accessibility/3-a-1/
2. Go to map tab 
3. Turn on screen reader
4. Tab to map search icon
5. Press space - it expands
6. Press n (to start typing in North East)
7. Screen reader reads out "Chart details heading level 5" and this becomes focused:

brockfanning:
@LucyGwilliamAdmin Strange, I can't recreate that using NVDA and Firefox/Chrome. Here's a video of what it's like for me.

https://user-images.githubusercontent.com/1319083/123676292-c0770000-d811-11eb-9bba-f4c96a5019dc.mp4

LucyGwilliamAdmin:
@brockfanning that's strange - it might be my version of nvda.

I was just looking for another example as the w3 one seems a bit off but this is how I imagine it behaving: https://alphagov.github.io/accessible-autocomplete/examples/

brockfanning:
@LucyGwilliamAdmin I agree that example is better. I think the main difference is in the help text that is read at various points, right? We could definitely go in that direction, but we'll just need requirements on what should be read (and when). And there would need to be some translations added for any new help text. It could also be argued that is an enhancement and not required - but DAC would probably be the judge of that.

LucyGwilliamAdmin:
@brockfanning ah yeah, that's true - I'll send the latest to DAC and get their thoughts. I think at this point we just want to get it past AA and like you say that would be more of an enhancement so maybe to come back to.

LucyGwilliamAdmin:
@brockfanning got some good feedback from DAC:

> This looks like its working great to me on first check. 
> 
> The only thing I would say is that when ‘hide search’ is pressed, and the search function is collapsed, focus has not been managed so that it is taken to the ‘show search’ button. Instead, focus is taken to the document body. The ‘show search’ is then the next thing in tab focus, but if screen readers press their cursor keys to navigate in reading mode (rather than using the tab key) they are taken back to the top of the page. I would advise that focus be explicitly set to the ‘show search’ button after ‘hide search’ is actioned.

brockfanning:
@LucyGwilliamAdmin Cool, that seems like an easy fix - I've just pushed a commit for that. It works when I test locally.

LucyGwilliamAdmin (approved):
Just tested on feature branch and looks/sounds good!

PR #1227, opened 2021-06-09
Requirements updates (docs)

brockfanning:
This is a docs update to refresh the third-party libraries that are used. This also starts the process of removing/replacing a couple of references to Prose.io (since we now have built-in forms for data and metadata).

### DIFF
--- a/docs/architecture.md
+++ b/docs/architecture.md
@@ -35,7 +35,7 @@
       Github or<br>other source
       <i tabindex="0" class="fa fa-info-circle"></i>
       <span class="info">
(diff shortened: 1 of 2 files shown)


LucyGwilliamAdmin:
@brockfanning I'm wondering now that we are recommending the data, metadata and indicator config as 3 separate inputs (even though indicator config is technically a meta input) should the three be mentioned? or do you think fine just to mention data and metadata?

brockfanning:
@LucyGwilliamAdmin That makes sense - I've just updated the architecture diagram in this PR.

LucyGwilliamAdmin (approved):

PR #1228, opened 2021-06-09
Quick fixes for some zoom issues on maps

### LINKED ISSUE #1194: Overlap of content on zoomed in map pages
Site repository URL: All
Open SDG version: All

**Describe the bug**
When the page is zoomed up to 400%, content overlaps so that it is not possible to read. 

**To Reproduce**
1. Go to https://sdgdata.gov.uk/3-a-1/
2. Zoom into 400%
3. Click on Map tab
4. Zoom in
5. Scroll down to bottom of map 
5. See error

**Expected behavior**
No overlap of content when zoomed in

**Screenshots**
image

**Additional context**
From accessibility report (NOTE they also mentioned making sure there is no two direction scrolling but maps are exempt from this)
Solution: 
Ensure that when users increase the size of pages to 400% and the browser width 1280px to view the page content in one column, they can do this without having to scroll in two directions, and without the loss of content or functionality.For more information, please refer to [F102: Failure of Success Criterion 1.4.10 due to content disappearing and not being available when content has reflowed](https://www.w3.org/WAI/WCAG21/Techniques/failures/F102).

Please note that the reflow guidelines specify that maps are considered exceptional content as they may often require two-dimensional layout to preserve their usage and meaning. However, the associated user interface controls should not overlap.

brockfanning:
Q | A
--- | ---
Fixed issues | Fixes #1194
Related version | 1.4.0-dev
Includes tests? |NA
Updated docs? |NA
Updated config forms? |NA
Added CHANGELOG entry? |❌

This has a quick fix to a couple of zoom-related issues mentioned in #1194:

1. On small-screens/high-zooms the attribution element has a max-width of 100px and aligns to the right
2. Even when loaded on small-screens/high-zooms, the map will never be shorter than 400px high.

### DIFF
--- a/_includes/assets/js/plugins/jquery.sdgMap.js
+++ b/_includes/assets/js/plugins/jquery.sdgMap.js
@@ -521,7 +521,11 @@
           }
           // Make sure the map is not too high.
           var heightPadding = 75;
+          var minHeight = 400;
           var maxHeight = $(window).height() - heightPadding;
+          if (maxHeight < minHeight) {
+            maxHeight = minHeight;
+          }
           if ($('#map').height() > maxHeight) {
             $('#map').height(maxHeight);
           }
--- a/_sass/components/_maps.scss
+++ b/_sass/components/_maps.scss
@@ -78,6 +78,12 @@
         margin-right: 4px;
       }
     }
+    .leaflet-control-attribution {
+      @media only screen and (max-width: 480px) {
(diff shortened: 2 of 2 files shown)


LucyGwilliamAdmin (approved):
This looks good to me and seems to be solving the issue that DAC raised so approving. We will get feedback from DAC at beta stage though

PR #1229, opened 2021-06-09
Document repository_url_data and repository_url_site fields

brockfanning:
This depends on https://github.com/open-sdg/jekyll-open-sdg-plugins/pull/101

Q | A
--- | ---
Fixed issues | NA
Related version | 1.4.0-dev
Bugfix, feature or docs? | docs
Includes tests? |❌
Updated docs? |✔️
Updated config forms? |✔️
Added CHANGELOG entry? |❌

These 2 new config settings may be useful in the future for various purposes, but the immediate purpose is to allow for a smoother quickstart process. As we switch to using the config forms in the quickstart, we'll need to populate the "repository_link" settings, to enable the "Go to repository" links on the forms. This feature is critical to the user-friendliness of the forms, in my opinion.

With these new settings, a new user would only need to fill in these two URLs, and the four "repository_link" settings could be pre-filled so that they work out-of-the-box.

I'm working up a re-write of the quickstart that will show an example.

### DIFF
--- a/docs/configuration.md
+++ b/docs/configuration.md
@@ -725,6 +725,26 @@ reporting_status:
 
 As always, for multilingual support, the title/description settings can refer to translation keys, and description can include Markdown.
 
+### repository_url_data
+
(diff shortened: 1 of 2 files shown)


brockfanning:
Tests will fail until the dependency PR is merged.

LucyGwilliamAdmin (approved):

PR #1233, opened 2021-06-12
Max width so that status will wrap on goal-by-target-vertical

### LINKED ISSUE #1195: Overlap of reporting status tags with indicator names on goal pages
Site repository URL: All
Open SDG version: At least latest - not sure about other goal layouts

**Describe the bug**
On vertical by target goal pages, the reporting status tag overlaps with the indicator name/link when zoomed into certain levels.

**To Reproduce**
1. Go to https://sdgdata.gov.uk/1/
2. Zoom into 200%
3. Scroll down to 1.4.1
Note: if the error doesn't display, try with different zoom levels and check different indicators

**Expected behavior**
No overlap between reporting status tag and indicator name

**Screenshots**
image

**Additional context**
From accessibility report:
Ensure that when users increase the size of pages to 400% and the browser width 1280px to view the page content in one column, they can do this without having to scroll in two directions, and without the loss of content or functionality.For more information, please refer to [F102: Failure of Success Criterion 1.4.10 due to content disappearing and not being available when content has reflowed](https://www.w3.org/WAI/WCAG21/Techniques/failures/F102).

brockfanning:
Q | A
--- | ---
Feature branch/test site URL | [Link](https://lucygwilliamadmin.github.io/sdg-indicators/1/)
Fixed issues | Fixes #1195 
Related version | 1.4.0-dev
Includes tests? | NA
Updated docs? | NA
Updated config forms? | NA
Added CHANGELOG entry? |❌

Screenshot to refer to:
image

### DIFF
--- a/_sass/layouts/_goal-by-target-vertical.scss
+++ b/_sass/layouts/_goal-by-target-vertical.scss
@@ -30,6 +30,7 @@ $goal-by-target-indentation: 60px;
 
   .goal-indicator {
     @media screen and (min-width: 769px) {
+      min-height: 80px;
       margin-left: $goal-by-target-indentation;
       position: relative;
       width: calc(100% - #{$goal-by-target-indentation});
@@ -42,6 +43,12 @@ $goal-by-target-indentation: 60px;
         top: 0;
         margin-top: 0;
         width: min-content;
+      }
+    }
+    @media screen and (min-width: 992px) {
+      min-height: auto;
+      .status {
+        width: auto;
       }
     }
   }
--- a/_sass/layouts/_goal-by-target-vertical.scss
+++ b/_sass/layouts/_goal-by-target-vertical.scss
@@ -41,6 +41,8 @@ $goal-by-target-indentation: 60px;
         right: 0;
         top: 0;
         margin-top: 0;
+        max-width: 100px;
+        text-align: right;
       }
     }
   }
--- a/_sass/layouts/_goal-by-target-vertical.scss
+++ b/_sass/layouts/_goal-by-target-vertical.scss
@@ -41,8 +41,7 @@ $goal-by-target-indentation: 60px;
         right: 0;
         top: 0;
         margin-top: 0;
-        max-width: 100px;
-        text-align: right;
+        width: min-content;
       }
     }
   }


LucyGwilliamAdmin:
@brockfanning @phillipgardiner link and screenshor added to description - I think left margin needs to be smaller and also the text should be left aligned

phillipgardiner:
@brockfanning @LucyGwilliamAdmin Looking at the screen shot. Agree with Lucy, ideally the left margin would be the same as it would be typically and the text should be left aligned :)

brockfanning:
@LucyGwilliamAdmin @phillipgardiner I just pushed another tweak to address this. I had added the right-aligning because it looked weird in the other direction, because the max-width caused some extra white-space on the right side. But I agree it doesn't look good either way.

Here's another approach, which forces the block to use "min-content", which means that the width will be the minimum possible for the contents:

screenshot-127 0 0 1_4000-2021 06 15-08_25_12

LucyGwilliamAdmin:
@phillipgardiner @brockfanning Site has been rebuilt and this is what it's looking like at the moment - no space between Exploring data sources and whatever tag is below:
image

Also I'm not sure if it's just the screen shot or if it would look like it on a site, but there doesn't seem to be the white background to the text on the Not applicable badge

phillipgardiner:
@LucyGwilliamAdmin @brockfanning Good spots, these would need to be address. I was working off the screen shot  :)

brockfanning:
@LucyGwilliamAdmin This is ready for a refresh. I've added a minimum height which should hopefully resolve that spacing issue. Also I've limited this wrapping behavior to a particular breakpoint (between 769-992px) so desktop users should continue to see the single-line statuses.

By the way, I don't think we ever added the white background on not-applicable here. Maybe we only did that on the reporting status page. Definitely something else to do at some point.

LucyGwilliamAdmin:
@brockfanning @phillipgardiner link has been rebuilt - looks good to me. Any comments Phill?

phillipgardiner:
@LucyGwilliamAdmin @brockfanning Looks good to me too - thank you. Could someone please add a ticket somewhere around adding the white background on not-applicable - I think this was required due to accessibility so best to correct at some point.

LucyGwilliamAdmin (approved):

PR #1234, opened 2021-06-13
Remove unused assets: tether.js, d3

brockfanning:
These two assets appear to be left-overs from a previous incarnation of the platform. The "tether" library is included, possibly because some type of tooltips may have been used in the past. Removing these assets does not seem to have any effect (as would be expected, since they are unused).

This also removes a bit of unneeded comments and an unused d3 script.

Link (temp): https://lucygwilliamadmin.github.io/sdg-indicators/


LucyGwilliamAdmin:
@brockfanning this seems to change the font in some instances e.g. before:
image

after:
image

brockfanning:
@LucyGwilliamAdmin Whoops! Sorry about that - I searched the codebase for "Roboto" but not for "Open Sans" - we are indeed using that one. I just added it back in the PR.

LucyGwilliamAdmin (approved):
Looks good!

PR #1235, opened 2021-06-14
Cookie consent form

### LINKED ISSUE #1231: Cookie enable/disable functionality
In order to meet data protection regulations, we need to give end users the option to to enable and disable cookies, with them being disabled by default.

Requirements:

Need to give option for out of box cookies -  cookies that remember your settings e.g. contrast state and for cookies that measure website use e.g. google analytics.

However countries should be able to configure what cookies they want to provide the option for e.g. if they start using another analytical software or customise the platform so there's another cookies that remembers settings.

Banner text should be customisable

Cookie journey:
[Cookie_Consent.pdf](https://github.com/open-sdg/open-sdg/files/6638166/Cookie_Consent.pdf)

Include documentation

brockfanning:
Q | A
--- | ---
Feature branch/test site URL | [Link (just essential)](http://uk-sdg-feature-branches.s3-website.eu-west-2.amazonaws.com/feature-cookies/)<br> [Link (with GA)](http://uk-sdg-feature-branches.s3-website.eu-west-2.amazonaws.com/feature-site-cookies-analytics/)
Fixed issues | Fixes #1231 
Related version | 1.4.0-dev
Includes tests? |❌
Updated docs? |✔️
Updated config forms? |✔️
Added CHANGELOG entry? |❌

## Testing instructions

* This depends on translations - the "cookies" branch of https://github.com/brockfanning/sdg-translations
* To enable, use this site config:

        cookie_consent_form:
            enabled: true

* We should also test with it disabled to make sure everything still works (contrast modes, Google Analytics, and any other javascript functionality)


brockfanning:
Needs to make sure Google Analytics still works if Klaro is not enabled.

brockfanning:
@LucyGwilliamAdmin I believe this is ready for review, from the perspective of non-UK countries. See testing instructions in the top post.

LucyGwilliamAdmin:
@brockfanning I've created a feature branch for this and just starting to take a look - what is {purposes} supposed to be here https://github.com/brockfanning/sdg-translations/blob/cookies/translations/en/cookies.yml#L9?

As currently it looks like this on platform: 
image

brockfanning:
@LucyGwilliamAdmin It looks like that notice breaks down when 100% of the services are required (ie, if Google Analytics is disabled and the only one left is the contrast cookie). I've added a special notice to display in this case, and also removed the "Reject" button in this case.

If you use the latest code from this branch and the sdg-translations "cookies" branch, you should get the special notice. You can also test out the normal notice by enabling Google Analytics (you can just put anything, like "foo", in the analytics.ga_prod setting, that is good enough to trigger this).

LucyGwilliamAdmin:
@brockfanning re. essential cookies only (feature branch):
- Should we get rid of the button which lets users choose which settings they use? So it would look like this, taking into account translation suggestions:

image

instead of this:
image


re. additional cookies (feature branch):
- I've suggested some additional translations for this so that the purposes aren't capitalised and also so that the That's ok button says Accept additional cookies, so taking into account the translations it would look this:

image


rather than this:
image

LucyGwilliamAdmin:
Also another thought, on the UK journey after you click accept/reject you get a confirmation message (see example below) - is there something available with Klaro that could be added?

image

brockfanning:
Regarding the confirmation message, no that is actually not provided by Klaro, so it would need to be custom code on top of Klaro. For that reason I think that it should be left out of this solution.

LucyGwilliamAdmin:
@brockfanning is this ready for me to take another look?

brockfanning:
@LucyGwilliamAdmin Not quite yet, I still need to implement some from the comments above.

brockfanning:
@LucyGwilliamAdmin Ok, this is ready for another look. I didn't distinguish the upper/lowercase versions, because I'm not sure offhand how that would work. I'm wondering, if it's still a problem, whether we should control that in CSS.

LucyGwilliamAdmin:
@brockfanning looking at this now:

- The feature branch with GA enabled looks good to me. 
- Re. essential cookies only, it is possible that the button says 'That's okay' or something similar instead of 'Accept additional cookies' but then for it to stay as 'Accept additional cookies' when there's additional cookies?
- Question re. both scenarios though, it seems that Klaro also stores a cookie, I think we need to mention this as an essential service?

brockfanning:
@LucyGwilliamAdmin 
That sounds possible - so if there are only required services then instead of "Accept additional cookies" it would say something like "Accept cookies"? Or just "OK"? Or something else?

That's a good question. I suppose it wouldn't hurt to add that, right? We would just need to come up with a description and title for it, and add that to the sdg-translations branch.

LucyGwilliamAdmin:
@brockfanning I think just "OK" will be fine - "Accept cookies" makes it sound like there's an option.

Is it possible to change the name then? Is that what you mean by title. I just had a look at a couple of gov.uk sites and they named it "cookies_preferences_set" and description is "Lets us know that you’ve saved your cookie consent settings".

@phillipgardiner I think we will also need to add to UK cookies notice

phillipgardiner:
@LucyGwilliamAdmin @brockfanning Do we not already have this on the UK version? 

cookies_settings | Saves your cookie consent settings | 1 year

LucyGwilliamAdmin:
@phillipgardiner oops, missed that - can just go with that then for the Open SDG version aswell if there's already text

brockfanning:
For the UK we renamed the cookie from "klaro" to "cookie_settings" (yep that's possible). I can do that here too. I'll use the same description, and for the title (which is what will display when the user clicks "Cookie settings" in the footer) I'll use "Cookie settings", which we conveniently already have a translation set up for. :)

brockfanning:
@LucyGwilliamAdmin Ok, ready for review again.

LucyGwilliamAdmin:
@brockfanning looks good - I do think we should go ahead and use CSS to make the title of the cookies purpose (i.e. Analytics cookies in screenshot below) all lowercase though if possible:
image

brockfanning:
@LucyGwilliamAdmin Ok, that is done and ready for a refresh now.

LucyGwilliamAdmin (approved):

Issue #1058, opened 2021-01-14
change main navigation default colour.


phillipgardiner:
After team discussion its been decided to change the main navigation default colour to black. Countries will then be able to customise the colour should they wish via the Sass variables available.

The black colour to be used is: #0b0c0c

PR #1241, opened 2021-06-18
Remove Google Webfont and document how to add them

brockfanning:
Q | A
--- | ---
Feature branch/test site URL | [Link](http://uk-sdg-feature-branches.s3-website.eu-west-2.amazonaws.com/feature-site-remove-webfonts/goals/)
Related version | 1.4.0-dev
Bugfix, feature or docs? |
Keeps backward-compatibility? |❌ (breaking change will need documentation in changelog)
Includes tests? |NA
Updated docs? |✔️
Updated config forms? |NA
Added CHANGELOG entry? |❌

### DIFF
--- a/_includes/head.html
+++ b/_includes/head.html
@@ -41,7 +41,6 @@
         <link rel="stylesheet" href="//maxcdn.bootstrapcdn.com/font-awesome/4.5.0/css/font-awesome.min.css" crossorigin="">
         <link rel="stylesheet" href="https://cdn.datatables.net/fixedcolumns/3.2.3/css/fixedColumns.dataTables.min.css" crossorigin="">
         <link rel="stylesheet" href="{{ site.baseurl }}/assets/css/style.css?v={{ cache_bust }}">
-        <link href="https://fonts.googleapis.com/css?family=Open+Sans|Roboto" rel="stylesheet" crossorigin="">
         <link rel="stylesheet" href="https://unpkg.com/leaflet@1.3.4/dist/leaflet.css"
           integrity="sha512-puBpdR0798OZvTTbP4A8Ix/l+A4dHDD0DGqYW6RQ+9jxkRFclaxxQb/SJAWZfWAkuyeQUytO7+7N4QKrDh+drA=="
           crossorigin=""/>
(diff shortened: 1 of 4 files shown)


LucyGwilliamAdmin:
@brockfanning @phillipgardiner link added - main changes are that headings are bolder

phillipgardiner:
@LucyGwilliamAdmin @brockfanning Sorry can I just check, we are removing open-sans and defaulting to Helvetica then Arial? When I click inspect I still see references to open-sans - is that normal (please excuse my lack of tech knowhow!)

I think I may have spotted one things but needs double checking. I think we have lost the grey on the main navigation, the main navigation should be #595959.

@AnnCorp @LucyGwilliamAdmin @brockfanning on a personal level I'm not sure I like the change. I can't put my finger on it though. The writing seems more blurry, I think this could be due to the new fonts being naturally smaller (e.g. both 16px but one naturally smaller) and less spaced. Like I say though more a personal opinion.

LucyGwilliamAdmin:
What if we change font weight to 400 - it then looks very similar to what we currently have:
image

phillipgardiner:
@LucyGwilliamAdmin that does look better to me, is there anyway of updating the link to show that as I'm struggling to amke that change my end to view it in aciton.

LucyGwilliamAdmin:
@phillipgardiner here you go - with 400: http://uk-sdg-feature-branches.s3-website.eu-west-2.amazonaws.com/feature-site-remove-webfonts-400/

phillipgardiner:
@brockfanning @LucyGwilliamAdmin Ignore my comment above about the colour of the main navigation. I was having a moment! We changed it to black (https://github.com/open-sdg/open-sdg/issues/1058). 

I think looks better with that weight of 400 :)

brockfanning:
@phillipgardiner Regarding the "Open Sans" still being there - I did leave that on purpose, to make it easier for people to continue using it. If it's missing, browsers will fallback to the next one in line: Helvetica. By leaving it in, if a country wants to continue using it, they just need to change one thing (adding the call to Google Webfonts).

Also just wanted to reference the issue I made to discuss this change, in general (and give folks a chance to raise objections): https://github.com/open-sdg/open-sdg/issues/1243

Let me also call attention to the note in that issue about the possibility of self-hosting the Open Sans font, which may be an alternative to dropping it.

Finally a UK-specific note on the subject of fonts - in the cookies PR I'm trying out the GDS Transport font, in case you're curious what that looks like.

PR #1253, opened 2021-06-25
Add language-specific classes to the language-toggle options

brockfanning:
Q | A
--- | ---
Feature branch/test site URL | [Link](https://lucygwilliamadmin.github.io/sandpit-site/)
Fixed issues | NA
Related version | 1.5.0-dev
Includes tests? |❌
Updated docs? |❌
Updated config forms? |NA
Added CHANGELOG entry? |❌

This adds classes to the options in the language toggle according to the class, eg `language-option-es` for Spanish. This can be useful if a language needs to use a particular font.


LucyGwilliamAdmin:
@brockfanning what's the best way to test this?

brockfanning:
@LucyGwilliamAdmin I guess the only way to test it would be to try it out and then inspect the markup of the language toggle options. If it's working, each toggle option should get a class of "language-option-[language]" according to which language it is.

LucyGwilliamAdmin (approved):
Looks good!

PR #1285, opened 2021-07-28
Self-host Open Sans instead of using Google Webfonts

### LINKED ISSUE #1243: Self-host "Open Sans" Google Webfont
Right now we are using a Google Webfont called "Open Sans". As we're starting to think about a better cookie-consent system, it is clear that we would need to have special documentation of this as it relates to privacy. This adds the additional complication of requiring that documentation to be translated in SDG Translations, since it would be displayed on the site in a cookie-consent form.

Since that feature was added, we've also added a "head-custom.html" include file, which can be easily used to add Google Webfonts if needed.

Are there any objections to removing the font in the next minor release, and providing help on how to use head-custom.html to maintain that functionality?

Alternatively, we could look into the possibility of hosting the font files in Open SDG, so that Google Webfonts is not required. I'm not 100% that's possible but it seems reasonable. Any insights are welcome!

brockfanning:
Q | A
--- | ---
Feature branch/test site URL | [Link](http://uk-sdg-feature-branches.s3-website.eu-west-2.amazonaws.com/feature-site-self-host-open-sans/)
Fixed issues | Fixes #1243 
Related version | 1.5.0-dev
* [ ] Includes tests
* [ ] Updated docs
* [ ] Updated config forms
* [x] Added CHANGELOG entry


LucyGwilliamAdmin (approved):
Looks good!

PR #1296, opened 2021-08-20
Fix for config builder loop issue

brockfanning:
Q | A
--- | ---
Feature branch/test site URL | [Link](https://lucygwilliamadmin.github.io/site/1-1-1/metadata/)
Related version | 1.5.0-dev
Bugfix, feature or docs? |
* [ ] Includes tests
* [ ] Updated docs
* [ ] Updated config forms
* [ ] Added CHANGELOG entry

This seems to fix a problem causing a javascript error on the config pages in some situations. This simple fix seems to resolve it. The error that happens is:

`Uncaught SyntaxError: return not in function metadata:639:8`

My impression is that this happens when an indicator has absolutely no metadata set. Ie, all of the metadata fields on the metadata config form have no values. (Such as if an indicator has no metadata file - only an indicator config file).


LucyGwilliamAdmin:
@brockfanning does this need testing? if so what's the best way to test?

brockfanning:
@LucyGwilliamAdmin First, to see the problem, maybe try creating an indicator with only an indicator configuration file, but not a metadata file. If my hunch is right, the indicator metadata form would appear blank (because of a javascript error). For example, create a file called 1-1-z.yml with this content:

```
composite_breakdown_label: ''
computation_units: ''
copyright: ''
data_footnote: ''
data_non_statistical: false
data_notice_class: ''
data_notice_heading: ''
data_notice_text: ''
data_show_map: false
data_start_values: []
embedded_feature_footer: ''
embedded_feature_html: ''
embedded_feature_tab_title: ''
embedded_feature_title: ''
embedded_feature_url: ''
expected_disaggregations: []
graph_annotations: []
graph_limits: []
graph_stacked_disaggregation: ''
graph_title: 'Testing
graph_titles: []
graph_type: bar
indicator_available: ''
indicator_name: 'Testing'
indicator_number: 1.1.z
national_geographical_coverage: ''
page_content: ''
permalink: ''
precision: []
reporting_status: complete
sort: ''
standalone: false
tags: []
```

Also in order to see indicator metadata forms you would need to enable them in the site configuration, if not already:

```
indicator_metadata_form:
  enabled: true
```

I think after making these changes, the metadata form will appear blank. I'll also see if I can work a test into this PR to illustrate the issue.

brockfanning:
Note - I had a typo in the example indicator config above.

brockfanning:
@LucyGwilliamAdmin Actually I would recommend you don't spend time on this. I'll investigate more and post again if needed. I'll change it to "Draft".

brockfanning:
@LucyGwilliamAdmin Ok, I have a better handle on this now. The problem only happens when both of these are true:
1. An indicator has no metadata file (ie, does have a config file and/or a data file, but no metadata file)
2. The metadata-config.yml (or _prose.yml) has none of the auto-generated fields. (An example of an auto-generated field is "indicator_name", which sdg-build automatically populates based on the indicator number.)

So this is a fairly unusual bug, but I think it's something that needs fixing. I added a test for this PR and then applied the fix, so that you can see the failing test and then the passing test. I think if you're happy with that as proof then it doesn't need manual testing, but certainly wouldn't object to holding off for manual testing.

LucyGwilliamAdmin (approved):
Looks good - fields are shoing and no error in console (see link in description)

PR #1308, opened 2021-08-24
Revamp of analytics files for GTM and GTAG support

### LINKED ISSUE #1258: Add support for Google Analytics 4 (latest version)
The Google Analytics code we currently have is compatible with Google Analytics - Universal Analytics. However it is not compatible with the new version (GA4).

More information about GA4 is here:
https://support.google.com/analytics/answer/10089681?hl=en

The code Canada are having to use instead is here:
https://github.com/sdg-data-canada-odd-donnees/cif-cic_dev/blob/develop/_includes/footer.html#L42-L52

brockfanning:
Q | A
--- | ---
Feature branch/test site URL | [Link](https://open-sdg-test-prod.github.io/)
Fixed issues | Fixes #1258 
Related version | 1.5.0-dev
Bugfix, feature or docs? |
* [ ] Includes tests
* [x] Updated docs
* [x] Updated config forms
* [ ] Added CHANGELOG entry

This is supported by a config update in a separate PR: https://github.com/open-sdg/jekyll-open-sdg-plugins/pull/109


brockfanning:
@LucyGwilliamAdmin This PR should work, although there are some issues to mention. The documentation explains the new config structure. Here are those issues:

* As mentioned in the documentation, the `ua` option (ie, the current option) is the only one that supports our custom events, like when the contrast toggle is pressed. The other two options (`gtag` and `gtm`) will not track those events.
* For the `gtm` option I think we would need to treat that as a separate service, for the purposes of the cookie consent form. So we'll need to come up with a title and description for GTM to translate and add to the cookie notice.
* For the `gtag` option I think we would need to treat that as a separate service, for the purposes of the cookie consent form as well. But I think we just need to tweak the title slightly, to distinguish it from the existing service. Like maybe "Google Analytics 4" instead of just "Google Analytics"?

LucyGwilliamAdmin:
Ah ok - yes, I've had a look and that seems to be how it works. So you're saying the issue would be if someone is using GTM for something other than GA?

LucyGwilliamAdmin:
@brockfanning so I'm looking at an article and it says that GTM doesn't set its own cookies:
https://www.analyticsmania.com/other-posts/does-google-tag-manager-use-cookies/

brockfanning:
@LucyGwilliamAdmin Ok, I've removed any additional cookie consent item for the `gtm` option, and I've consolidated the `ua` and `gtag` options into a single cookie consent item (the existing one). Does that seem right to you? Anything else I've missed?

LucyGwilliamAdmin:
@brockfanning 
Apologies, I've just realised in my comment above (https://github.com/open-sdg/open-sdg/pull/1308#issuecomment-906305785) some text isn't showing as expected:

The second bullet _ga_<container-id> should be `_ga_<container-id>` - wasn't showing correctly as I didn't put it in code block.

Apparently this is used to persist session state - I'm having that cookie set on my test site.

Is there anything additional we need to add for that?

brockfanning:
@LucyGwilliamAdmin I don't know of any way to predict what those extra cookies would be, so I've added an extra config option "extra_cookies" inside the "analytics" config. The idea is that users can add the extra cookies there, so that the cookie consent form knows about them. If you think this is a good approach, I can add this to the jekyll-open-sdg-plugins PR as well.

LucyGwilliamAdmin:
@brockfanning I think that makes sense

LucyGwilliamAdmin (approved):

LucyGwilliamAdmin (changes requested):
@brockfanning just went to do final check and I'm not getting cookie banner anymore:

https://open-sdg-test-prod.github.io/

brockfanning:
@LucyGwilliamAdmin Ok, maybe I messed something up. Can you point me to the site repo for that?

LucyGwilliamAdmin:
@brockfanning here's the site repo link: https://github.com/open-sdg-test/site

brockfanning:
@LucyGwilliamAdmin I'm not seeing the cookie_consent_form site configuration setting, or maybe I'm missing it?

LucyGwilliamAdmin:
Ohh right - just added and rebuilding

LucyGwilliamAdmin:
@brockfanning ok, so the notice is showing now but it's only mentioning essential cookies even though I've configured google analytics

brockfanning:
@LucyGwilliamAdmin Ok thanks, I'm taking a look now.

brockfanning:
@LucyGwilliamAdmin Yes I think there was a logic problem. Can you give it another try?

LucyGwilliamAdmin:
@brockfanning great, working as it should

Just one more thing - I've just remembered that @phillipgardiner put this comment in Slack: 

> Just looking at the cookie links above, on the 'with analytics cookies' journey there is a sentence displayed 'There were some changes since your last visit.....' is this sentence necessary do we think? I have to admit because it was underlined it confused me and I was trying to click it on order to provide my consent (thinking it was a link and I would be taken through to another page to provide consent).

I personally think that the sentence is needed as some people might wonder why the message is showing again if they've already confirmed cookies. But I do agree that the underlining could be removed.

Thoughts?

brockfanning:
@LucyGwilliamAdmin Ok, I think we can remove the underlining with a CSS change. Let me try to recreate that.

brockfanning:
@LucyGwilliamAdmin I've made that change - I've got it now as bold instead of underlined.

brockfanning:
@LucyGwilliamAdmin @phillipgardiner 
screenshot-127 0 0 1_4000-2021 09 08-11_53_36

LucyGwilliamAdmin (approved):
@brockfanning discussed with Phill - looks good but think the sentence could make more sense, Just opened PR in sdg-translations: https://github.com/open-sdg/sdg-translations/pull/230

PR #1309, opened 2021-09-01
Updates to annotation defaults and simpler versions for target_lines and series_breaks

### LINKED ISSUE #1297: Make changes to annotation defaults
Annotation defaults currently take focus away from dataset. To avoid this the following defaults would be better:

```
drawTime: 'beforeDatasetsDraw',
borderColor: '#949494',
borderDash: [10, 5],
borderWidth: 1,
label: {
    backgroundColor: 'white',
    fontColor: 'black'
}
```

brockfanning:
Q | A
--- | ---
Feature branch/test site URL | [Link](http://uk-sdg-feature-branches.s3-website.eu-west-2.amazonaws.com/feature-annotations-test) (see 1.1.1)
Fixed issues | Fixes #1297 
Related version | 1.5.0-dev
Bugfix, feature or docs? |
* [ ] Includes tests
* [x] Updated docs
* [x] Updated config forms
* [ ] Added CHANGELOG entry

This PR is supported by a [PR on jekyll-open-sdg-plugins](https://github.com/open-sdg/jekyll-open-sdg-plugins/pull/107) and a [PR on sdg-translations](https://github.com/open-sdg/sdg-translations/pull/229).

### DIFF
--- a/docs/indicator-configuration.md
+++ b/docs/indicator-configuration.md
@@ -320,7 +320,7 @@ graph_stacked_disaggregation: Age
     * `label`: The text to display on the annotation. Defaults to "2030 target".
     * `value`: The value at which to draw the line. This number corresponds to your actual data.
 
-Note that this setting relies on the Chart.js annotation plugin. If you would like to tweak any other settings for more site-wide control, you can override the `_includes/components/annotation_presets.js` file.
+Note that this setting relies on the Chart.js annotation plugin. If you would like to tweak any other settings for more site-wide control, you can override the `_includes/components/charts/annotation_presets.js` file.
 
 ### graph_title
 
--- a/docs/indicator-configuration.md
+++ b/docs/indicator-configuration.md
@@ -234,7 +234,9 @@ The following fields affect the display of graphs. Currently only longitudinal g
 
 ### graph_annotations
 
-**_Optional_**: This setting can be used to add line annotations to the graph, such as target lines to show the progress towards the 2030 goal for an indicator. Like `graph_titles` it can include multiple annotations, and limited to particular units or series. Each item can have the following settings:
+**_Optional_**: Note that this setting is quite complex, and is not recommended unless you really need a particular type of graph annotation. For much simpler alternatives specifically designed for the most common use-cases, see the `graph_target_lines` and `graph_series_breaks` settings.
+
+This setting can be used to add line annotations to the graph, such as target lines to show the progress towards the 2030 goal for an indicator (though, again, there is a separate `graph_target_lines` setting specifically designed for target lines). Like `graph_titles` it can include multiple annotations, and limited to particular units or series. Each item can have the following settings:
 
     * `series`: If specified, the annotation will only display when the user is looking at this series.
     * `unit`: If specified, the annotation will only display when the user is looking at this unit of measurement.
@@ -290,6 +292,17 @@ graph_limits:
     maximum: 2000
 ```
 
+### graph_series_breaks
+
+**_Optional_**: This setting can be used to add "series break" annotations (ie, gaps in the years) to the graph. Like `graph_titles` it can include multiple items, and each can be limited to particular units or series. Each item can have the following settings:
+
+    * `series`: If specified, the series break will only display when the user is looking at this series.
+    * `unit`: If specified, the series break will only display when the user is looking at this unit of measurement.
+    * `label`: The text to display on the annotation. Defaults to "2030 target".
+    * `value`: The value at which to draw the line. This number should be between 0 (the left side of the chart) and the number of years minus 1 (the right side of the chart).
+
+Note that this setting relies on the Chart.js annotation plugin. If you would like to tweak any other settings for more site-wide control, you can override the `_includes/components/annotation_presets.js` file.
+
 ### graph_stacked_disaggregation
 
 **_Optional_**: This setting can be used with the "bar" graph type to place a certain disaggregation (such as "Age") into the same "stacked" bars.
@@ -297,6 +310,17 @@ graph_limits:
 ```nohighlight
 graph_stacked_disaggregation: Age
 ```
+
+### graph_target_lines
+
+**_Optional_**: This setting can be used to add "target line" annotations to the graph. Like `graph_titles` it can include multiple items, and each can be limited to particular units or series. Each item can have the following settings:
+
+    * `series`: If specified, the target line will only display when the user is looking at this series.
+    * `unit`: If specified, the target line will only display when the user is looking at this unit of measurement.
+    * `label`: The text to display on the annotation. Defaults to "2030 target".
+    * `value`: The value at which to draw the line. This number corresponds to your actual data.
+
+Note that this setting relies on the Chart.js annotation plugin. If you would like to tweak any other settings for more site-wide control, you can override the `_includes/components/annotation_presets.js` file.
 
 ### graph_title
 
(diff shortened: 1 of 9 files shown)


LucyGwilliamAdmin:
@brockfanning I seem to be having an issue with this one - the lines don't seem to be showing

I've updated to point at your branch as well as pointing at sdg-translations and jekyll branches, and here's the individual indicator configs:

feature branch link|type|config
--------------------|-----|-------
[link](http://uk-sdg-feature-branches.s3-website.eu-west-2.amazonaws.com/feature-site-annotation-default-updates/1-1-1/)|both|[link](https://github.com/ONSdigital/sdg-data/blob/feature-annotations-test/meta/1-1-1.md)
[link](http://uk-sdg-feature-branches.s3-website.eu-west-2.amazonaws.com/feature-site-annotation-default-updates/1-2-1/)|just series|[link](https://github.com/ONSdigital/sdg-data/blob/feature-annotations-test/meta/1-2-1.md)
[link](http://uk-sdg-feature-branches.s3-website.eu-west-2.amazonaws.com/feature-site-annotation-default-updates/1-2-2/)|just target|[link](https://github.com/ONSdigital/sdg-data/blob/feature-annotations-test/meta/1-2-2.md)

brockfanning:
@LucyGwilliamAdmin I think I got the issue - this is ready for a refresh.

LucyGwilliamAdmin:
@brockfanning great, they're showing now.

@phillipgardiner see links in above comment.

Both - I think if target line is at edge chart should be extended? But since that's already what it looks like, we could put that into next version perhaps? Unless it's a quick fix?

brockfanning:
@LucyGwilliamAdmin I agree, though it may be a bit tricky. I guess we would impose a "graph limit" that is some percentage higher/lower than the highest/lowest value, unless the user had already set a graph limit. I think it's probably possible but that it's complex enough that it's probably not a quick fix.

LucyGwilliamAdmin:
@brockfanning ok, no probs - def not something for this sprint then but I'll add an issue

LucyGwilliamAdmin:
@brockfanning so we've just been having a play around with the series and targets lines but having an issue.

So @phillipgardiner wanted to see how the following looked

target_line: borderwidth 1px; borderdash [15, 10]; colour: `#757575`
series_break: borderwidth 1px; borderdash [2, 2]; colour: `#949494`

but doesn't seem to be working, colour or borderdash- could you have a go instead please?

Here's the changes I made: https://github.com/ONSdigital/sdg-indicators/blob/feature-site-annotations-test-trials/_includes/components/charts/annotation_presets.js

docs/indicator-configuration.md
```diff
@@ -290,6 +292,17 @@ graph_limits:
...
+### graph_series_breaks
+
+**_Optional_**: This setting can be used to add "series break" annotations (ie, gaps in the years) to the graph. Like `graph_titles` it can include multiple items, and each can be limited to particular units or series. Each item can have the following settings:
+
+    * `series`: If specified, the series break will only display when the user is looking at this series.
+    * `unit`: If specified, the series break will only display when the user is looking at this unit of measurement.
+    * `label`: The text to display on the annotation. Defaults to "2030 target".
+    * `value`: The value at which to draw the line. This number should be between 0 (the left side of the chart) and the number of years minus 1 (the right side of the chart).
+
+Note that this setting relies on the Chart.js annotation plugin. If you would like to tweak any other settings for more site-wide control, you can override the `_includes/components/annotation_presets.js` file.
```
LucyGwilliamAdmin, line 304: Is this supposed to be `_includes/components/charts/annotation_presets.js`?

docs/indicator-configuration.md
```diff
@@ -298,6 +311,17 @@ graph_limits:
...
+### graph_target_lines
+
+**_Optional_**: This setting can be used to add "target line" annotations to the graph. Like `graph_titles` it can include multiple items, and each can be limited to particular units or series. Each item can have the following settings:
+
+    * `series`: If specified, the target line will only display when the user is looking at this series.
+    * `unit`: If specified, the target line will only display when the user is looking at this unit of measurement.
+    * `label`: The text to display on the annotation. Defaults to "2030 target".
+    * `value`: The value at which to draw the line. This number corresponds to your actual data.
+
+Note that this setting relies on the Chart.js annotation plugin. If you would like to tweak any other settings for more site-wide control, you can override the `_includes/components/annotation_presets.js` file.
```
LucyGwilliamAdmin, line 323: Same as above - should this be `_includes/components/charts/annotation_presets.js`?

docs/indicator-configuration.md
brockfanning: @LucyGwilliamAdmin Ah, that's right, I'll update those docs. Did you already try that or no?

LucyGwilliamAdmin: Do you mean where I'm trying to overwrite? Yeah, I've put it in the charts folder

brockfanning: Got it - OK I'll investigate.

phillipgardiner:
@brockfanning @LucyGwilliamAdmin Hey, as i'm not working tomorrow (and I think we wanted this complete by end of play tomorrow) I wont be able to provide comment, but in addition to the changes above to the lines, we also need to add information about the target and the series breaks to the tables (so the information is available there too). My current thoughts are for the target we would do something like this...
Screenshot (215)
and for the series break I think this would simply be written about in text within the footnote.

brockfanning:
@LucyGwilliamAdmin As far as the chart annotations go, I think this is ready for a refresh.

@phillipgardiner I agree about the table. I think that is enough work that it would delay 1.5.0 though, as we would have to make it dependent on the same config setting, and also dynamic based on series/unit, just like the chart annotations. Are we OK with making that a separate issue?

phillipgardiner:
@brockfanning @LucyGwilliamAdmin Looks good to me. My only debate now which is a knock on effect is should we change the mouse hover line to a solid line and maybe adjust its thickness as a result. I'm assuming that is configurable per country though?

@brockfanning Yep agree the table work is a separate ticket for 1.6.0. I'll work on the a actual design / layout to be used and add to a ticket.

LucyGwilliamAdmin:
@brockfanning these are how the lines are now looking:
http://uk-sdg-feature-branches.s3-website.eu-west-2.amazonaws.com/feature-site-annotations-test-trials/1-1-1/

target_line: borderwidth 2px; borderdash [15, 10]; colour: #757575
series_break: borderwidth 1px; borderdash [2, 2]; colour: #949494

Could you make these the default here please?

phillipgardiner:
@brockfanning in addition in the UK we believe the better experience as a resukt of the above changes to the target line and series line is to make the mouse hover line a solid line, borderwidth 2px; colour: #757575. @brockfanning do you agree we should therefore make this the default for OpenSDG, as i'm guessing countries could overwrite if they really wanted to revert back to a dashed line.

brockfanning:
@LucyGwilliamAdmin @phillipgardiner I'm OK with changing the hover style for the line charts. As you say it is possible for countries to override if necessary. It could be considered a "breaking change" but if it is clearly an improvement then I think it's fine. I've made those changes, so it's ready for a refresh.

LucyGwilliamAdmin (approved):
Looks good - thanks @brockfanning !

LucyGwilliamAdmin:
[Show time series break](https://trello.com/c/Bx1sOT94/224-show-time-series-break)

PR #1311, opened 2021-09-01
Remove invalid values from map ranges

### LINKED ISSUE #1248: NaN values are used for valueRange in maps
Site repository URL: [https://github.com/MoBosse/testSite](https://github.com/MoBosse/testSite)
Open SDG version: 1.4.0-dev

**Describe the bug**
If the csv file of an indicator with map contains a row without a value (I want those years to be displayed in the table) this NaN is included when determining the range of values according to which the colors are assigned on the map.

Example: Here is a map where all map relevant lines in the csv have a value
grafik

Now, when I delete only one value (for the state in the south-east), the value range seems to be calculated with a zero as minimal value as the colour of several other states have changed too. Furthermore, the legend in the upper right corner no longer shows a value.

grafik

**To Reproduce**
1. Go to [https://mobosse.github.io/testSite/en/1-4-2/](https://mobosse.github.io/testSite/en/1-4-2/)
2. Click on 'map'
3. See the described behaviour

**Expected behavior**
I would expect nan values to be ignored when the range of values is determined

brockfanning:
Q | A
--- | ---
Fixed issues | Fixes #1248 
Related version | 1.5.0-dev
Bugfix, feature or docs? |
* [ ] Includes tests
* [ ] Updated docs
* [ ] Updated config forms
* [x] Added CHANGELOG entry

### DIFF
--- a/_includes/assets/js/plugins/jquery.sdgMap.js
+++ b/_includes/assets/js/plugins/jquery.sdgMap.js
@@ -390,6 +390,11 @@
         }
 
         // Calculate the ranges of values, years and colors.
+        function isMapValueInvalid(val) {
+          return _.isNaN(val) || val === '';
+        }
+        minimumValues = _.reject(minimumValues, isMapValueInvalid);
+        maximumValues = _.reject(maximumValues, isMapValueInvalid);
         plugin.valueRange = [_.min(minimumValues), _.max(maximumValues)];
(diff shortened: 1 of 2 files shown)


LucyGwilliamAdmin (approved):
Approving based on Moritz's review on #1248

PR #1317, opened 2021-09-06
Draft of 2020 indicator update docs

LucyGwilliamAdmin:
Q | A
--- | ---
Related version | 1.5.0-dev
Bugfix, feature or docs? |
* [ ] Includes tests
* [ ] Updated docs
* [ ] Updated config forms
* [ ] Added CHANGELOG entry

### DIFF
--- a/docs/2020-indicator-updates.md
+++ b/docs/2020-indicator-updates.md
@@ -1,0 +1,20 @@
+As part of the UN IAEG-SDG's [2020 Comprehensive Review Process](https://unstats.un.org/sdgs/iaeg-sdgs/2020-comp-rev/), 36 changes to the SDG indicators were recommended. The breakdown of changes include:
+
+- 14 proposals for replacements of existing indicators
+- 8 proposals for revisions of existing indicators
+- 8 proposals for additional indicators
+- 6 proposals for deleting indicators, including the partial deletion of 4.2.1
+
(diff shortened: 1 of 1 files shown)


LucyGwilliamAdmin:
@AnnCorp @brockfanning I've put this together as a starting point for 2020 indicator update docs - any thoughts?

AnnCorp:
@LucyGwilliamAdmin jsut wondering is it worth briefly adding something about if want to keep a list of archive indicators, how to do it and poss link to UK one as an example (or any other platforms that have taken this approach too?)

LucyGwilliamAdmin:
@AnnCorp I think we possibly could but we have used customisation to create our archive page. However, maybe we could do one of the following:

1. Create a page in Open SDG so others can easily replicate (@brockfanning I think you may haven't mentioned this a while back)
2. Write a blog and link to it from these docs 

What are your thoughts?

brockfanning:
@LucyGwilliamAdmin @AnnCorp Given that the UK archive approach is custom, I feel like the blog approach is best. But yes if we decide that it should be a feature of Open SDG, then we could implement that and then add documentation about it.

AnnCorp:
@LucyGwilliamAdmin @brockfanning as UK approach was custom then agree blog approach better.

LucyGwilliamAdmin:
@brockfanning I think this is ready to go

PR #1295, opened 2021-08-17
Favicon cleanup to standardize around favicon.io

brockfanning:
Q | A
--- | ---
Feature branch/test site URL | [Link](http://uk-sdg-feature-branches.s3-website.eu-west-2.amazonaws.com/feature-site-favicon-cleanup/)
Fixed issues | Fixes #808 
Related version | 1.5.0-dev
Bugfix, feature or docs? | feature
* [x] Keeps backward-compatibility
* [ ] Includes tests
* [x] Updated docs
* [ ] Updated config forms
* [x] Added CHANGELOG entry

PR #1324, opened 2021-09-08
1.5.0-beta1 upgrade instructions

brockfanning:
This is not intended to be reviewed/merged yet, but is here to provide upgrade instructions as we release beta versions.

### DIFF
--- a/docs/changelog.md
+++ b/docs/changelog.md
@@ -1,6 +1,6 @@
 <h1>Change Log</h1>
 
-## Unreleased
+## 1.5.0
 
 * Fix for empty values in map ranges #1311
 * Allow for placeholder indicators #1310
--- a/docs/updates.md
+++ b/docs/updates.md
@@ -2,9 +2,26 @@
 
 For a more technical list of platform changes, see [the change log](changelog.md).
 
-## 1.4.0
+## 1.5.0
 
 Releasing soon
+
+* Cookie consent form
+* New ways to embed Google Analytics
+* Remove 2 third-party services: Google Webfonts and ShareThis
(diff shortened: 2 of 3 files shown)


LucyGwilliamAdmin:
@brockfanning I think there may be a breaking change from https://github.com/open-sdg/open-sdg/pull/1295. I'm testing now and needed to change the favicons so that our custom one shows

brockfanning:
@LucyGwilliamAdmin Thank you for flagging that - what was the change you needed to make?

LucyGwilliamAdmin:
@brockfanning I deleted everything that was in the favicons folder (see [commit](https://github.com/ONSdigital/sdg-indicators/commit/9a607de478cc69eecd71322dbcca70ead5936829#diff-98085b6004acaffa9a1a6f1f8777a00ec87e9d94835faa2e62016f82e24aa9ad)) and then followed the instructions to convert the image with favicon.io and upload the new files

brockfanning:
@LucyGwilliamAdmin Ah ok - I hate to cause a breaking change for something so trivial.... Unfortunately though, if we standardize around the favicon.io files, they are named differently from the old files.

One way around it is to have a `favicons` config setting, with the options of "classic" and "favicon.io". And depending on which is chosen, the favicon images will be named differently. And it would default to "classic". What do you think?

LucyGwilliamAdmin:
@brockfanning yeah I guess that could work, but I think def would be something to add to the list to deprecate in 2.0.0

brockfanning:
@LucyGwilliamAdmin Ok, I've got a PR up for that: #1330

PR #1350, opened 2021-09-26
Reporting status config setting

### LINKED ISSUE #1320: Reporting status tags - ability to choose which are displayed
Development is required in order to provide countries with the ability to choose which reporting statuses are displayed and which are not. 

For example in the UK, now the number of indicators being reported on have reached a high level, research has shown a btter experience would be to only show tags next to those indicators where data DOES NOT yet exist. 

An example of this approach can be found here: 
[Tag functionality.pdf](https://github.com/open-sdg/open-sdg/files/7116455/Tag.functionality.pdf)

brockfanning:
Q | A
--- | ---
Fixed issues | Fixes #1320 
Related version | 1.6.0-dev
Bugfix, feature or docs? |
* [x] Includes tests
* [x] Updated docs
* [x] Updated config forms
* [ ] Added CHANGELOG entry

This depends on two other PRs:
* https://github.com/open-sdg/jekyll-open-sdg-plugins/pull/112
* https://github.com/open-sdg/sdg-build/pull/277

Note: With this PR and the two related PRs, the `reporting_status` field can be removed from the `_prose.yml` schema (or whatever it is called) in the data repository.

### DIFF
--- a/_includes/components/reportingstatus/reporting-status-by-field.html
+++ b/_includes/components/reportingstatus/reporting-status-by-field.html
@@ -19,26 +19,26 @@
         </span>
         <div class="summary">
             <div class="statuses">
-                {%- for status in fieldreport.statuses -%}
-                    {% assign status_type = site.reporting_status.status_types | where: "value", status.status | first %}
-                    {% assign status_type_value = status_type.value | default: status.status %}
-                    {% assign status_type_label = status_type.label | default: status.status %}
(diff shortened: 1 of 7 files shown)


brockfanning:
Hold off on testing this - I may need to combine other PRs with it.

LucyGwilliamAdmin:
@brockfanning do we also need to configure disaggregation reporting status options?

Tagging @otis-bath for info

PR #1360, opened 2021-10-01
New approach for sorting hierarchical fields, based on top-level parents

### LINKED ISSUE #1334: Data ordering not working as expected
Data repository URL: https://github.com/ONSdigital/sdg-data
Data or metadata issue: Data
Input format: CSV

**Describe the bug**
We are using the below code to sort our dissaggregations based on the order they appear in the CSV. 

```
datapackage:
  sorting: default
```

However, we have at least one case when it is not working as expected. 9.3.1 data has been ordered so that the disaggregations show in the order: Industry sector, Industry higher sub-category, Industry middle sub-category, Industry lower sub-category.
However the platform is showing the order: Industry sector, Industry middle sub-category, Industry higher sub-category, Industry lower sub-category.

[9.3.1 data](http://uk-sdg-feature-branches.s3-website.eu-west-2.amazonaws.com/feature-data-ordering-data/en/data/9-3-1.csv)
[9.3.1 edges](http://uk-sdg-feature-branches.s3-website.eu-west-2.amazonaws.com/feature-data-ordering-data/en/edges/9-3-1.json)

**Additional context**
Looking at the edges file above, could it be something there causing this issue, as there's an edge between Industry sector and Industry higher sub-category and an edge between Industry sector and Industry middle sub-category

brockfanning:
Q | A
--- | ---
Fixed issues | Fixes #1334 
Related version | 1.6.0-dev
Bugfix, feature or docs? |
* [ ] Includes tests
* [ ] Updated docs
* [ ] Updated config forms
* [ ] Added CHANGELOG entry

Previously there was some (pretty complex) code dedicated to sorting the disaggregation fields for display, so that the "child" dropdowns would appear just beneath the "parent" dropdowns. This re-sorting, however, interfered with any order that might have been in the data schema (whether a custom order, or alphabetical, or based on the order of the columns/rows in the data).

This is a refactor of that sorting logic. The idea here is to give up on trying to handle multiple levels of hierarchy, and just deal with two layers: top-level parents (parents that are not children of any other parents) and everything else. Since Open SDG only visualizes those two layers (ie, there are only two different indentation-levels for the dropdowns) maybe we don't need to worry about it.

So the result of this change is that the order of the dropdowns should be identical to the order of the fields in the data schema, except that any children appear after their top-level parent, and before the next top-level parent. Hopefully this results in an intuitive and expected order.

But a lot of testing here would be good.

### DIFF
--- a/_includes/assets/js/model/fieldHelpers.js
+++ b/_includes/assets/js/model/fieldHelpers.js
@@ -186,25 +186,56 @@ function sortFieldsForView(fieldItemStates, edges) {
 
     // We need to sort the edges so that we process parents before children.
     var parents = edges.map(function(edge) { return edge.From; });
-    edges.sort(function(a, b) {
-      if (!parents.includes(a.To) && parents.includes(b.To)) {
-        return 1;
-      }
-      if (!parents.includes(b.To) && parents.includes(a.To)) {
-        return -1;
-      }
-      return 0;
+    var children = edges.map(function(edge) { return edge.To; });
+    var topLevelParents = parents.filter(function(parent) {
+      return !(children.includes(parent));
     });
 
-    edges.forEach(function(edge) {
-      // This makes sure children are right after their parents.
-      var parentIndex = fieldItemStates.findIndex(function(fieldItem) {
-        return fieldItem.field == edge.From;
+    var topLevelParentsByChild = {};
+    children.forEach(function(child) {
+      var currentParent = edges.find(function(edge) { return edge.To === child; }),
+          currentChild = child;
+      while (currentParent) {
+        currentParent = edges.find(function(edge) { return edge.To === currentChild; });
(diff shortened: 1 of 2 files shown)


LucyGwilliamAdmin:
@brockfanning I've just done a super quick test on the case that I raised the bug about and seems to work fine? Can you suggest any other cases I could test?

brockfanning:
@LucyGwilliamAdmin I think it would be good to test any indicators with complex hierarchies, but also it would be good to test some simple ones too. So, unfortunately just a general once-over. I would feel better if it got some close attention. We have improved this code many times over the past year, so it would be a big regression if we messed it up.

LucyGwilliamAdmin:
@brockfanning Ok, thanks - UK data team are going to do full site check over the next week

PR #1380, opened 2021-10-18
Fix javascript error on non-search pages

brockfanning:
This fixes a regression caused by hotfix 1.5.1. I believe that without this change, all non-search pages will have a javascript error in the console. The error doesn't seem to be causing problems with other functionality, as far as I can tell, but should be fixed. I'm putting this PR against 1.5.0-dev on the assumption that it will be another hotfix release.


brockfanning:
@LucyGwilliamAdmin @otis-bath Just checking to see if you had a look at this one.

PR #1413, opened 2021-11-03
Release notes for 1.6.0-beta

brockfanning:
Release notes for 1.6.0. To be updated as we work through the beta testing.

### DIFF
--- a/_includes/assets/js/model/helpers.js
+++ b/_includes/assets/js/model/helpers.js
@@ -40,6 +40,7 @@
 
   {% include assets/js/model/constants.js %}
   {% include assets/js/model/utils.js %}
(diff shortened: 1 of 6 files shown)


brockfanning:
@LucyGwilliamAdmin @otis-bath This is ready for review. I ran into two breaking changes while testing overrides, but was able to adjust some code to avoid them. It may be a good idea to do a quick feature-branch and confirm that indicator pages generally work as expected, with these backwards-compatibility fixes. Also let me know if you know of anything I missed in the changelogs.

LucyGwilliamAdmin (approved):
Just put on feature branch and looks fine

PR #1479, opened 2021-11-19
Handle case where starting value does not exist in selected series

### LINKED ISSUE #1477: data_start_values causing error when used on "Series"
Data repository URL: [https://github.com/MoBosse/testData](https://github.com/MoBosse/testData)
Data or metadata issue: data
Input format: csv

**Describe the bug**
data_start_values seem not to work in combination with Series_toggle and a combination of series with and without disaggregation.

I have the following situation: 

- Two Series (A and B). 
- A has no disaggregation but only a total series
- B has no total series but two disaggregations (x & y)
- A has the unit %
- B has the unit number
See data [here](https://github.com/MoBosse/testData/blob/develop/data/indicator_1-5-2.csv)

If I do not add any start values, A is selected by default. When I switch to B, the row "x" is selected. 
Now I want "y" to also be selected when I switch to B, so I added the following

```
 data_start_values: 
  - series: B
    field: disagg
    value: x
  - series: B
    field: disagg
    value: y
```

See meta [here](https://github.com/MoBosse/testData/blob/develop/meta/1-5-2.md)

With the start values the chart fails to load. See result [here](https://mobosse.github.io/testSite/1-5-2/)

I have tried many variations of the start values, but nothing seems to work.
I also tried with the same unit for all series but the result is more or less the same. See [here](https://mobosse.github.io/testSite/1-5-3/)

brockfanning:
Q | A
--- | ---
Fixed issues | Fixes #1477 
Related version | 1.7.0-dev
Bugfix, feature or docs? |
* [ ] Includes tests
* [ ] Updated docs
* [ ] Updated config forms
* [ ] Added CHANGELOG entry

### DIFF
--- a/_includes/assets/js/model/fieldHelpers.js
+++ b/_includes/assets/js/model/fieldHelpers.js
@@ -162,7 +162,9 @@ function fieldItemStatesForView(fieldItemStates, fieldsByUnit, selectedUnit, dat
           var fieldItemValue = fieldItem.values.find(function(valueItem) {
             return valueItem.value === selectedValue;
           });
-          fieldItemValue.checked = true;
+          if (fieldItemValue) {
+            fieldItemValue.checked = true;
+          }
         })
       }
     });


brockfanning:
I think I can add a test for this one.

brockfanning:
This has been manually tested in #1477

otis-bath (approved):
Already tested by Moritz -OB

PR #1484, opened 2021-11-19
Support for footer_fields indicator configuration

### LINKED ISSUE #1395: Dynamically change footer fields per series
User story:
I need the footnote area to reflect the source, units, and geographical area, of the data selected on the 'disaggregation controls', only
So that references are not made to data not displayed

There are currently four footer fields on Open SDG:

- Source
- Geographical Area
- Unit of measurement
- Footnote

None of this fields change dynamically when the series shown is changed but often there is a difference, which can lead to issues such as:
- countries will put info for all series in the fields which can lead to users not knowing which series they belong to.
- important information wont be shown

Unit of Measurement and Footnote map to SDMX series level attributes. Units is also changed dynamically else where: y-axis and chart tooltips. However footnote is also static.

brockfanning:
Q | A
--- | ---
Feature branch/test site URL | [[Link](http://uk-sdg-feature-branches.s3-website.eu-west-2.amazonaws.com/feature-footer-fields)]
Fixed issues | Fixes #1395 
Related version | 1.7.0-dev
Bugfix, feature or docs? |
* [x] Includes tests
* [x] Updated docs
* [x] Updated config forms
* [ ] Added CHANGELOG entry

### DIFF
--- a/_includes/assets/js/indicatorView.js
+++ b/_includes/assets/js/indicatorView.js
@@ -71,6 +71,7 @@ var indicatorView = function (model, options) {
     view_obj.createSelectionsTable(args);
 
     view_obj.updateChartTitle(args.chartTitle);
+    view_obj.updateSeriesAndUnitElements(args.selectedSeries, args.selectedUnit);
   });
 
   this._model.onFieldsComplete.attach(function(sender, args) {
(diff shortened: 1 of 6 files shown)


otis-bath (approved):
I have added these for 2.5.2 which you can see on the FB. Everything else works fine -OB

LucyGwilliamAdmin:
@brockfanning quick question - how do you activate the units footer field being dynamic?

brockfanning:
@LucyGwilliamAdmin I believe in the config for "footer_fields" you would set the unit that it should display for. Then if that unit is not selected, that footer field would not display.

If you're talking about the old "computation_units" field -- that wouldn't be dynamic, so I think you would want to omit that field. Is it possible to omit that one or do we still require it?

LucyGwilliamAdmin:
@brockfanning so it doesn't just pull it from the data?

I thought that's what #1395 was about

Apologies if I've misunderstood

brockfanning:
@LucyGwilliamAdmin Yeah I guess there was a misunderstanding. This PR allows for new footer fields (via the new "footer_fields" config setting) which can be associated with a series and/or unit. In other words, they only display when a particular series and/or unit is selected. This PR has no connection with data, and does not alter the behavior of any of the existing footer fields (like computation_units).

LucyGwilliamAdmin:
@brockfanning ok, so I think the thoughts were that the info could either be pulled from the data (if available) or from the config options. Should I create a new issue or reopen #1395

PR #1496, opened 2021-12-06
Change more font-sizes to use variables

brockfanning:


brockfanning:
This should not change anything, but the best way to test it would be to make a feature branch and them confirm that everything is the same.

otis-bath (approved):
Have put this on a feature branch with the goal format and compared it with a separate feature branch and they are exactly the same. All looks good -OB

PR #1498, opened 2021-12-06
Map disaggregation select (optional)

brockfanning:
Q | A
--- | ---
Related version | 1.7.0-dev
Bugfix, feature or docs? | feature
* [x] Includes tests
* [x] Updated docs
* [ ] Updated config forms
* [ ] Added CHANGELOG entry

This adds an optional select drop-down to maps that allows the user to switch between combinations of disaggregations. In order to turn this on, this site configuration is needed:

```
map_disaggregation_select: true
```


christoph-LSN:
@brockfanning I'd like to to test ist
but pointing to remote_theme: brockfanning/open-sdg@map-disaggregation-select-latest in my [_config.yml](https://github.com/christoph-LSN/MT_Site/blob/develop/_config.yml#L6)
and adding map_disaggregation_select: true to my [site_config](https://github.com/christoph-LSN/MT_Site/blob/develop/_data/site_config.yml) obviously has no effect....

Do I have to change something else?

LucyGwilliamAdmin:
@brockfanning I get the same as Christoph, even when I don't add `map_disaggregation_select: true`

christoph-LSN:
@brockfanning @LucyGwilliamAdmin 
at the first glance it works as you can see here: [https://christoph-lsn.github.io/MT_Site/3-1-3/](https://christoph-lsn.github.io/MT_Site/3-1-3/)

In this indicator both percent and numbers are present. The map changes correctly when chaning the dropdown, but the min max values do not change.

image

LucyGwilliamAdmin:
@brockfanning thanks - data is showing for me now.

The main thing I've found so far is that it doesn't ignore the special columns i.e. (obs_status, units multiplier, unit measure):
http://uk-sdg-feature-branches.s3-website.eu-west-2.amazonaws.com/feature-site-map-disaggregation-select-latest/3-a-1/

brockfanning:
Great feedback. Definitely seems like at least two improvements are needed:
1. Figure out how to ignore the "non-disaggregation columns"
2. Special treatment of the units of measurement - changing the unit of measurement does need to affect the color legend in the upper right

Regarding number 2, that does make me wonder if it is better to try and keep the left sidebar open, so that the series/unit toggles can still work. But maybe that is best left as a future improvement.

brockfanning:
I've updated this PR so that the min/max values in the "color legend" get updated whenever the drop-down is changed.

@christoph-LSN Can rebuild and give it another try?
@LucyGwilliamAdmin One way to avoid that issue might be to try adding the "exclude_columns" parameter in the map_layers config of your data repository. Admittedly it would be a little duplicative, because it would need to be repeated in each of the layers. It's also possible to automate this by removing "non-disaggregation columns" from the geojson, and I've put up a [pull-request](https://github.com/open-sdg/sdg-build/pull/289) for that, if you'd like to test it. I'm not sure whether it's better to keep it explicit by relying on the exclude_columns parameter, or whether it's better to automate it like in that PR, so definitely open to your thoughts.

christoph-LSN:
@brockfanning for [this ](https://christoph-lsn.github.io/MT_Site/3-1-3/) indicator (3.1.3) it works fine. The min max values are changing after changing the dropdown. But in [this indicator](https://christoph-lsn.github.io/MT_Site/2-1-3/) (2.1.3) it does not.

image

There are many disaggregations which might be usefull in the table or chart view but may be not in the map view. perhaps we can exclude "männlich" and "weiblich" and leave only "insgesamt"

As you said above:
> Regarding number 2, that does make me wonder if it is better to try and keep the left sidebar open, so that the series/unit toggles can still work. But maybe that is best left as a future improvement.

Does that mean, tha che map shows what is selected in the sidebar? if so, it is perhaps easier to use the sidebar than the additional dropdown in the map?

LucyGwilliamAdmin:
@brockfanning thank you, I've just tried editing in the data config for now so can get something ready to send to DAC. However, I still seem to be a having a bit of an issue. It's like it thinks England is the headline data for some reason. Other than the default map, everything only seems to show for England: http://uk-sdg-feature-branches.s3-website.eu-west-2.amazonaws.com/feature-map-disaggregation-select-latest/3-a-1/

I'll see if there's a less complex map which can be used for accessibility testing but just letting you know.

brockfanning:
@christoph-LSN Can you elaborate on how it is not working in 2.1.3? I do see it changing, but maybe I'm missing a case where it should change and is not?

@LucyGwilliamAdmin Thanks, I thought that adding to `exclude_columns` would work, but I guess not. I'll look into the reason that's not working.

brockfanning:
@LucyGwilliamAdmin I was confused - I think the exclude_columns parameter is actually working. At first I thought that "Country" was showing up, but now I see there is a separate column called "Country of birth", and that is what is appearing in the dropdown.

Actually that highlights a usability issue: The dropdown options only contain the disaggregation values, but not the disaggregations (subcategories) themselves. So for example we see "England" in the dropdown but we don't know that it is "Country of birth". I think that's definitely something that needs resolving, and DAC will probably notice it too. This seems to me like a deal-breaker for this dropdown approach, really. I guess we could add the disaggregation names in the dropdown options, but then the dropdown would be extremely wide, and I know that would introduce accessibility issues (like not being able to read it on mobile).

I might still be confused what you mean by "other than the default map, everything only seems to show for England". Maybe you could point to a specific case where the number is not what you expect?

christoph-LSN:
@brockfanning 
e.g. tick the first option in the dropdown: "insgesamt - besondere Aufenthaltsrechte"
in the legend min = 10 max = 95
But e.g. "Heidkreis" = 250

image

brockfanning:
@LucyGwilliamAdmin @christoph-LSN Given the remaining bugs here, along with the fact that there is a fundamental usability problem, I'm thinking we should abandon this PR and start over with a different approach. I'm thinking that this needs to integrate with the left sidebar, like I mentioned above. At least that would resolve the usability problem, and then we could dig into the bugs. It might be a waste of time to dig into the bugs here, when we can't use this approach anyway.

christoph-LSN:
> @LucyGwilliamAdmin @christoph-LSN Given the remaining bugs here, along with the fact that there is a fundamental usability problem, I'm thinking we should abandon this PR and start over with a different approach. I'm thinking that this needs to integrate with the left sidebar, like I mentioned above. At least that would resolve the usability problem, and then we could dig into the bugs. It might be a waste of time to dig into the bugs here, when we can't use this approach anyway.

@brockfanning this i will be a good approach. If there is anything I can do let me know! And perhaps, if it does not take to much time you can show what are the main obstackles to integrate the left sidebar. I am not a js devleoper but I will try to dig in it...

brockfanning:
Ok, thanks all for the testing! I will close this and loop you both in when there is something ready with the other approach.

PR #1508, opened 2021-12-15
Unit and series support for maps

brockfanning:
This is a draft just for review and testing. The changes in this PR are:

1. When the "Map" tab is selected, the left sidebar remains visible
2. When the "Map" tab is selected, the "Sub-categories" section is hidden
3. When the "Map" tab is selected, a new "Map sub-categories" section appears if there are any disaggregations being displayed on the map
4. The data on the map should always reflect any current selections in the "Unit" or "Series" toggles

This notably does _not_ provide any way to change the sub-categories being displayed on the map, but is a first attempt to try and integrate units and series, at least. However this does attempt to show the sub-categories that are being displayed on the map (which was not happening before).

Q | A
--- | ---
Feature branch/test site URL | [Link](add link here)
Fixed issues | Fixes #ISSUENUMBER
Related version | 1.6.0-dev
Bugfix, feature or docs? |
* [ ] Keeps backward-compatibility
* [ ] Includes tests
* [ ] Updated docs
* [ ] Updated config forms
* [ ] Added CHANGELOG entry

PR #1515, opened 2021-12-22
Map unit and series support with subcategories

brockfanning:
Q | A
--- | ---
Related version | 1.8.0-dev
* [ ] Includes tests
* [ ] Updated docs
* [ ] Updated config forms
* [ ] Added CHANGELOG entry

**Note**: This depends on an sdg-build PR: https://github.com/open-sdg/sdg-build/pull/290. So the data repository's requirements.txt file needs: `git+git://github.com/brockfanning/sdg-build-1@geojson-output-translate-disaggregation`

This is a draft variant of #1508 which also adds support for the subcategories, by changing them into radio buttons. Here are the differences from #1508:

1. After switching to the map tab, the subcategories remain visible
2. After switching to the map tab, the subcategories change to radio buttons instead of checkboxes. If multiple checkboxes were selected, only the most recent selection for each category is kept.
3. After switching back to the chart/table tab, they change back to checkboxes, and any previously-selected checkboxes become checked again.
4. If the unit/series/subcategory selections have no data, the usual no-data message appears over the map.


christoph-LSN:
@brockfanning @LucyGwilliamAdmin 
No, the map doesn't show at all.

[https://christoph-lsn.github.io/MT_Site/4-1-1/](https://christoph-lsn.github.io/MT_Site/4-1-1/)

I point to 1.8.0-dev or schould I point to 1.7.0-dev
[https://github.com/christoph-LSN/MT_Site/blob/3e7c58cad8cabaddb34cf6aefbb3a22b50c67de8/Gemfile#L7](https://github.com/christoph-LSN/MT_Site/blob/3e7c58cad8cabaddb34cf6aefbb3a22b50c67de8/Gemfile#L7)

brockfanning:
@christoph-LSN For the jekyll-open-sdg-plugins, I think either branch should be fine. But one important thing is pointing to the sdg-build branch. I put an example in the description above of what the data repo's requirements.txt should have, for sdg-build. Could you try that if you haven't already?

christoph-LSN:
@brockfanning 
I have tried it, but it's not working
Did I do something wromng with the[ requirements.txt](https://github.com/christoph-LSN/IM_Data/blob/develop/scripts/requirements.txt)

clicking on the map tab shows the "Data not avaliable" message

hitting "Auswahl aufheben"   the wheel in the map begins to turn

image

christoph-LSN:
@brockfanning 
Only an idea: Maybe it works if excluding 'Gebietseinheit' from the disaggregation

christoph-LSN:
@brockfanning 
may be the problem is, that I have all "Gebietseinheiten" have to be displayed on the map.  And if only one "Gebietseinheit" is selected in the disagg menue may be this caueses the problem.

e.g. the number of peoble with migation status, sex, and time they live in lower saxony: [indicator 1.2.3](https://christoph-lsn.github.io/MT_Site/1-2-3/) has to show all "Gebietsenheiten" on the map but only sex or time they live in lower saxony (Aufenthaltsdauer)

brockfanning:
@christoph-LSN I do think that Gebietseinheiten should be excluded from the geojson - which you can do by uncommenting this line: https://github.com/christoph-LSN/IM_Data/blob/develop/config_data.yml#L110 - because I believe it "mirrors" the GeoCode column.

There were also some javascript errors happening which I hopefully fixed by my last commit. So you could try again.

This does bring up an important point: There are going to be data columns that "mirror" the regions of the map - for example Gebietseinheiten in your case. These columns are needed in order to show the regional breakdown on the the table and chart. But on the map they are unneeded and actually cause a problem. So really, when the user switches to the map tab, Open SDG needs to identify these columns, uncheck any selections in the sidebar, and then disable or hide them in the sidebar.

For now when you are testing, please uncheck any selections in Gebietseinheiten in order to get the map to work.

christoph-LSN:
@brockfanning @LucyGwilliamAdmin 
It does not work more or less...
In [indicator 1.1.1](https://christoph-lsn.github.io/MT_Site/1-1-1/) I have to click "Auswahl aufheben" then a small piece of map appears. switching fullsreen and back shows the map correctly.

In [indicator 3.1.3](https://christoph-lsn.github.io/MT_Site/3-1-3/) no map is shown at all. may be at least becuase I only can tick"Kinderbetreuung" when also one city of "Gebietseinheit" is ticked also...

image

If I inspect the map it shows this warning: `DevTools failed to load source map: Could not load content for https://cdnjs.cloudflare.com/ajax/libs/javascript-canvas-to-blob/3.15.0/js/canvas-to-blob.min.js.map: HTTP error: status code 404, net::ERR_HTTP_RESPONSE_CODE_FAILURE`

brockfanning:
@christoph-LSN Thanks for the feedback, that's yet another great point to consider: not only is it a problem when there are data columns that mirror the map regions, but even worse, sometimes those columns are "parents" of other columns. So the solution is not as easy as hiding them.

The impression that I'm getting from this is that the left sidebar disaggregation controls are great for the chart and table, but really not so great for maps -- even if we are able to switch them to radio-buttons instead of checkboxes. So that makes me lean more towards going back to one of the other approaches we've tried.

christoph-LSN:
> So that makes me lean more towards going back to one of the other approaches we've tried.

yes, that have been my thoughts too....

brockfanning:
As discussed, we'll close this and purse the other approaches.

PR #1548, opened 2022-01-24
Chart.js 3 and graph_type per unit/series

brockfanning:
The changes in this PR include: (EDIT - cleaned up this PR and these notes on 2022-02-11)
1. Upgrade to latest version of Chart.js (3.7.0)
2. A `graph_types` indicator configuration setting for controlling the chart type by unit/series
3. Support for using 0 (instead of -1) in binary charts to represent "No"
4. Shapes (circle/square) in chart tooltips should match the shape from the legend below the chart
5. Refactor the indicatorView.js file to split it up into many separate files

Note that all of these things rely on a site config setting:

```
chartjs_3: true
```

As for testing, I would recommend this:

1. A feature branch with `chartjs_3: true` as well as:
    * Some testing of the `graph_types` setting on one or more indicators
    * Some testing of the binary chart type
2. A feature branch with `chartjs_3: false`, which should hopefully work the same as before -- ie, none of the features mentioned above, but nothing should break

EDITS: I've removed the tab stuff from this PR, to keep it more focused. I've also changed the name of the configuration setting from "version_2_preview" to "chartjs_3".


brockfanning:
@LucyGwilliamAdmin @otis-bath I believe this is ready for review - let me know if you run into any problems.

otis-bath:
Thanks @brockfanning, I will take a look at it tomorrow and let you know what comes up

otis-bath:
Hi @brockfanning, I have taken a look at each of these on various FBs. 

For the graph types code - it has worked mostly although as you can see here http://uk-sdg-feature-branches.s3-website.eu-west-2.amazonaws.com/feature-version-2-preview/3-3-3/ where I changed the graph type for number of cases and rate per 100,000 it has dropped the y-axis label and despite setting the graph type to line when the units are Percentage (%), it is now defaulting to the 'Born on the Indian subcontinent' disaggregation. Making this change has also dropped the y-axis label on every indicator.

For the binary graph testing, I have it here on 15-6-1 which has worked as intended for No=0 http://uk-sdg-feature-branches.s3-website.eu-west-2.amazonaws.com/feature-version-2-preview/15-6-1/ I have used the series setting to ensure it only applies to the latter two series

When the version 2 preview is set to 'false', everything is as normal without breaking changes http://uk-sdg-feature-branches.s3-website.eu-west-2.amazonaws.com/feature-version-2-off/ and the y-axis is restored to each indicator.

brockfanning:
@otis-bath Thanks for the testing - that was very helpful. The issue with the y axis label should be resolved now, and that alerted me to some other settings that may not have been working. For example I noticed that the line graphs were no longer starting at 0. Do you mind refreshing the feature branches? By the way I believe that the "born on Indian subcontinent" behavior is already happening, so probably is not related to these code changes.

LucyGwilliamAdmin:
@phillipgardiner re. binary, US are the main users of the binary charts right now - they may be the only. I think re. how it looks, we agreed on Tues we would leave as is due to the low number of users. The main thing to be changed was that 0 could be used as well as -1 so from my POV, that looks fine.

However, re. bars on bar charts (and binary charts), there seems to be some weird outline - I think we may have had this before

phillipgardiner:
@LucyGwilliamAdmin Yes I did wonder why the outlines were now appearing?

@LucyGwilliamAdmin Thanks for the response. Yes I couldn't find any examples of this being used anywhere. My initial thoughts were there is something not quite right about it. It's kind of like the dates should run through the middle and then the No's come down from that and the Yes's go up or something? I was also wondering if the Yes and No should be different colours. Like I say, I felt I needed to understand the use case before I could really input.

But as you've said due to the low number of users happy to leave as it, so @otis-bath happy to sign off sort of speak.

@LucyGwilliamAdmin @otis-bath My only query would be in the example of the feature branch, is the axis label 'numbers' what we would expect? or does this not matter as only relevant for the feature branch.

otis-bath (approved):
Approving at the code fixes what it should, probably need a discussion of binary graphs down the line if we want to implement -OB

brockfanning:
@otis-bath I noticed a few other issues, particularly related to high-contrast and font colors. I've tried to fix them. I've also tried to remove that border around the bars. Do you mind refreshing the feature branches? But also note that I've changed the configuration setting from "version_2_preview" to "chartjs_3", so you'll need to update that in the feature branches.

I've also removed the tab stuff from this PR, to try and keep it more focused.

@phillipgardiner I wanted to make sure I understand your note about the axis label 'numbers'? Are you talking about the Y axis label when "Number of cases" is selected?

phillipgardiner:
@brockfanning for clarity I was referring to the Y axis label when viewing this feature branch:  http://uk-sdg-feature-branches.s3-website.eu-west-2.amazonaws.com/feature-version-2-preview/15-6-1/. The axis label is 'Number' and yet the outputs (not sure thats the right word) are Yes and No (which I deemed not to be numbers). Hope that makes sense.

brockfanning:
@phillipgardiner Thanks, that makes sense. The country can control that - I believe - with their "Units" column. However I can't think of any reason that they wouldn't want it to be "Yes/No". So maybe we should force it to be "Yes/No" whenever it's using that "binary" chart type?

otis-bath:
@brockfanning - have updated the site config and refreshed the branches. Series/units control on graph still working as planned (for when Chartjs_3 is true only) however grey bordering still visible on binary graph and defaulting to bar charts despite metadata stipulating it should be line. What was the issue related to the high-contrast and font colors as can check those out? 

http://uk-sdg-feature-branches.s3-website.eu-west-2.amazonaws.com/feature-version-2-preview/
http://uk-sdg-feature-branches.s3-website.eu-west-2.amazonaws.com/feature-version-2-off/

Indicators are 3-3-3 for series/unit control and 15-6-1 for binary graph

brockfanning:
@otis-bath I've pushed up attempts to fix those issues.

otis-bath:
> @otis-bath I've pushed up attempts to fix those issues.

Hi @brockfanning - thank you for this, now all working as intended, the grey bordering has disappeared and the graph controls are as they should be.

PR #1550, opened 2022-01-25
Remove rescaler chartjs plugin

brockfanning:
This removes a bit of Chart.js custom code which we currently use, that we called "rescaler". The code is complex enough that I don't understand exactly what it is supposed to do. My theory is that it has something to do with the ticks on the X and/or Y axis of the charts. Like - maybe it affects the min/max Y values, for example. It's a little problematic in the Chart.js version 3 upgrade, and I would prefer to remove this code, unless we still need it.

@LucyGwilliamAdmin @otis-bath Would it be possible to make a feature branch of this? Then we could compare with the live site and see what changes.


otis-bath (approved):
Hi @brockfanning, I have this on a FB http://uk-sdg-feature-branches.s3-website.eu-west-2.amazonaws.com/feature-site-remove-rescaler-plugin and have checked through a number of graphs and indicator pages and it looks the same

brockfanning:
@otis-bath Thanks for that. I also can't see any difference. Do you or @LucyGwilliamAdmin have any objections to removing this code?

otis-bath:
@brockfanning - fine by me

LucyGwilliamAdmin:
@otis-bath @brockfanning removing doesn't seem to affect the setting of limits. For example, this [metadata](http://uk-sdg-feature-branches.s3-website.eu-west-2.amazonaws.com/feature-limits-no-rescalar-data/en/meta/14-2-1.json) states that graph limits for the percentage units should be 0-100 and on this feature branch with the rescaler plugin removed, the limits are working:
http://uk-sdg-feature-branches.s3-website.eu-west-2.amazonaws.com/feature-limits-no-rescalar/14-2-1/

So fine with me also

LucyGwilliamAdmin:
@otis-bath @brockfanning removing doesn't seem to affect the setting of limits. For example, this [metadata](http://uk-sdg-feature-branches.s3-website.eu-west-2.amazonaws.com/feature-limits-no-rescalar-data/en/meta/14-2-1.json) states that graph limits for the percentage units should be 0-100 and on this feature branch with the rescaler plugin removed, the limits are working:
http://uk-sdg-feature-branches.s3-website.eu-west-2.amazonaws.com/feature-limits-no-rescalar/14-2-1/

So also fine with me

PR #1560, opened 2022-03-07
Dynamic unit of measurement in data footer

### LINKED ISSUE #1549: Dynamically populate Units footer fields with data column values
#1484 allows for footer fields to be configured on an indicator-by-indicator basis (including limiting to a specific series or units).

However, where possible these fields should be populated using the data. For example, if a CSV has a Units column, the values in this column will be used to populate the Unit of measure footer field and it would change dynamically when a toggle option is changed.

brockfanning:
Q | A
--- | ---
Feature branch/test site URL | [Link](http://uk-sdg-feature-branches.s3-website.eu-west-2.amazonaws.com/feature-dynamic-unit-in-data-footer/15-2-1/)
Fixed issues | Fixes #1549 
Related version | 1.8.0-dev
Bugfix, feature or docs? |
* [ ] Includes tests
* [ ] Updated docs
* [ ] Updated config forms
* [ ] Added CHANGELOG entry

### DIFF
--- a/_includes/assets/js/indicatorView-legacy.js
+++ b/_includes/assets/js/indicatorView-legacy.js
@@ -399,12 +399,14 @@ var indicatorView = function (model, options) {
 
   this.updateUnitElements = function(selectedUnit) {
     var hasUnit = typeof selectedUnit !== 'undefined';
-    if (hasUnit) {
-      $('.data-controlled-footer-field.unit-from-data').show();
-      $('dd.data-controlled-footer-field.unit-from-data').text(translations.t(selectedUnit));
+    var fallback = this._model.measurementUnit;
+    if (hasUnit || fallback) {
+        var unitToDisplay = selectedUnit || fallback;
+        $('.data-controlled-footer-field.unit-from-data').show();
+        $('dd.data-controlled-footer-field.unit-from-data').text(translations.t(unitToDisplay));
     }
     else {
-      $('.data-controlled-footer-field.unit-from-data').hide();
+        $('.data-controlled-footer-field.unit-from-data').hide();
     }
   }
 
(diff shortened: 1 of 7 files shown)


LucyGwilliamAdmin:
@brockfanning @otis-bath @phillipgardiner I've added feature branch link to description.

Seems good - there's a Units footer field which has been added that reacts to the Series/Units toggle option being changed. However the Unit of measurement from the metadata still remains in the footer

brockfanning:
@LucyGwilliamAdmin I wasn't sure how to handle that. Should we stop showing the metadata unit field? Maybe hide it if the data has a "Units" column? That may be possible, but I'm usually hesitant to add "automagic" stuff like that. As it is now, my impression was that the user has full control: if they don't want the metadata unit field to display, they can remove "computation_units" from their metadata, and if they don't want the data unit field to display, they can remove the "Units" column from their data. That said, we should test that out, and also see how this interacts with the hide_single_units setting.

Here some scenarios that come to mind, along with what I would expect to happen:

1. Data has no "Units" column, metadata has a "computation_units" field. The value for "computation_units" displays in the footer, the Y axis, and in tooltips.
2. Data has "Units" column with only a single unit, and nothing in "computation_units" metadata field. That single unit displays in footer, the Y axis, and in tooltips. The Unit toggle appears only if hide_single_units is false.
3. Data has "Units" column with multiple units, and nothing in "computation_units" metadata field. The unit in footer, Y axis, and tooltips, is dynamic based on the currently-selected unit.
4. Data has "Units" column and a metadata has "computation_units" field: Not sure - Is this something that users would want to do? If so what would they expect to happen?

I can go back and try to test these out locally, but feedback is welcome.

LucyGwilliamAdmin:
@brockfanning I think they all seem right re. footer, Y axis and tooltips but my only other concern is the field in the metadata. What would this be in all cases as kind of needs to be representative of the whole indicator - should that be what is in the computation_units field?

brockfanning:
@LucyGwilliamAdmin I've updated this PR to try and accomplish this - the feature branch can be refreshed when you have time.

LucyGwilliamAdmin (approved):
Checked all scenarios and looks good!

Agree that Unit of measurement in metadata sections should use Units from metadata file

PR #1565, opened 2022-03-09
Chartjs 3 accessibility fix

### LINKED ISSUE #1559: Chart keyboard navigation not working with Chart.js 3
If `chartjs_3` is set to true, the tooltip keyboard navigation doesn't work. It's supposed to allow users to cycle through the tooltips using the arrow keys.

brockfanning:
Fixes #1559 

Testing instructions: Currently with the `chartjs_3` setting enabled, it is not possible to use the keyboard's arrow keys to cycle through the tooltips on a chart. With this PR, that should work again.

One note - only applicable when multiple datasets are being shows (ie, when multiple disaggregations have been selected):

As before, the line charts are quite different from other chart types: With line charts, all data points along a vertical line (eg, year) are active at one time. For example, all points for "2021" would be active at once.

With other chart types (eg, bar) only a single data points (eg, a single bar) should be active at one time.

Feature branch: http://uk-sdg-feature-branches.s3-website.eu-west-2.amazonaws.com/feature-site-chartjs-3-accessibility-fix/


brockfanning:
@LucyGwilliamAdmin Got it, I'll work on that problem.

brockfanning:
@otis-bath I've made a code change that is intended to prevent that issue that @LucyGwilliamAdmin noticed.

otis-bath:
> @otis-bath I've made a code change that is intended to prevent that issue that @LucyGwilliamAdmin noticed.

Great thanks @brockfanning - have just refreshed and tested this and now working properly with the graph points staying the same

otis-bath (approved):
Happy to approve this as the fix worked, everything else looking good from what I can see - OB

PR #1567, opened 2022-03-11
More consistent filename changes for chartjs3

brockfanning:
This change keeps the `indicatorView.js` file at the same name as before, with the chartjs3 version named differently.

In version 2.0, we will just have one version, but for now let's be consistent and have the new stuff in separately-named files.

Ideally nothing should change with this PR - everything should work the same, whether `chartjs_3` is set to `true` or `false`.


brockfanning:
@otis-bath I believe this is straightforward and is not worth the time to create a feature branch, so I will go ahead merge this. If there are any problems with it, they will come up during the testing of other PRs.

otis-bath:
Okay great thanks @brockfanning

PR #1569, opened 2022-03-15
Time series attributes

brockfanning:
Q | A
--- | ---
Related version | 1.8.0-dev
Bugfix, feature or docs? |
* [ ] Includes tests
* [ ] Updated docs
* [ ] Updated config forms
* [ ] Added CHANGELOG entry

This depends on a PR in jekyll-open-sdg-plugins: https://github.com/open-sdg/jekyll-open-sdg-plugins/pull/126

That PR provides default support for one time-series attribute (COMMENT_TS) but a new `time_series_attributes` site config setting can be used to have full control over this.

More notes to come.

### DIFF
--- a/_includes/assets/js/indicatorModel.js
+++ b/_includes/assets/js/indicatorModel.js
@@ -287,6 +287,14 @@ var indicatorModel = function (options) {
       filteredData = helpers.getDataByUnit(filteredData, this.selectedUnit);
     }
 
+    var timeSeriesAttributes = [];
+    if (filteredData.length > 0) {
+      timeSeriesAttributes = helpers.getTimeSeriesAttributes(filteredData);
+    }
+    else if (headline.length > 0) {
+      timeSeriesAttributes = helpers.getTimeSeriesAttributes(headline);
+    }
+
     filteredData = helpers.sortData(filteredData, this.selectedUnit);
     if (headline.length > 0) {
       headline = helpers.sortData(headline, this.selectedUnit);
@@ -328,6 +336,7 @@ var indicatorModel = function (options) {
       chartType: this.graphType,
       indicatorDownloads: this.indicatorDownloads,
       precision: helpers.getPrecision(this.precision, this.selectedUnit, this.selectedSeries),
+      timeSeriesAttributes: timeSeriesAttributes,
     });
   };
 };
--- a/_includes/assets/js/indicatorView-chartjs3.js
+++ b/_includes/assets/js/indicatorView-chartjs3.js
@@ -73,6 +73,7 @@ var indicatorView = function (model, options) {
         helpers.updateChartTitle(args.chartTitle);
         helpers.updateSeriesAndUnitElements(args.selectedSeries, args.selectedUnit);
         helpers.updateUnitElements(args.selectedUnit);
+        helpers.updateTimeSeriesAttributes(args.timeSeriesAttributes);
     });
 
     MODEL.onFieldsComplete.attach(function (sender, args) {
--- a/_includes/assets/js/model/dataHelpers.js
+++ b/_includes/assets/js/model/dataHelpers.js
@@ -118,3 +118,27 @@ function inputEdges(edges) {
   {% endif %}
   return edgesData;
 }
+
+/**
+ * @param {Array} rows
+ * @return {Array} Objects containing 'field' and 'value', to be placed in the footer.
+ */
+function getTimeSeriesAttributes(rows) {
+  if (rows.length === 0) {
+    return [];
+  }
+  var timeSeriesAttributes = [],
+      possibleAttributes = {{ site.time_series_attributes | jsonify }},
+      firstRow = rows[0],
+      firstRowKeys = Object.keys(firstRow);
+  possibleAttributes.forEach(function(possibleAttribute) {
+    var field = possibleAttribute.field;
+    if (firstRowKeys.includes(field) && firstRow[field]) {
+      timeSeriesAttributes.push({
+        field: field,
+        value: firstRow[field],
+      });
+    }
+  });
+  return timeSeriesAttributes;
+}
--- a/_includes/assets/js/model/helpers.js
+++ b/_includes/assets/js/model/helpers.js
@@ -105,6 +105,7 @@
     getGraphAnnotations: getGraphAnnotations,
     getColumnsFromData: getColumnsFromData,
     inputEdges: inputEdges,
+    getTimeSeriesAttributes: getTimeSeriesAttributes,
     inputData: inputData,
     // Backwards compatibility.
     footerFields: deprecated('helpers.footerFields'),
--- a/_includes/assets/js/model/utils.js
+++ b/_includes/assets/js/model/utils.js
@@ -65,6 +65,10 @@ function nonFieldColumns() {
     'Unit multiplier',
     'Unit measure',
   ];
+  var timeSeriesAttributes = {{ site.time_series_attributes | jsonify }};
+  timeSeriesAttributes.forEach(function(tsAttribute) {
+    columns.push(tsAttribute.field);
+  });
   if (SERIES_TOGGLE) {
     columns.push(SERIES_COLUMN);
   }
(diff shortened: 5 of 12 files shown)


otis-bath (approved):
Have this working and visible on 9-c-1 and 10-4-1 on this FB http://uk-sdg-feature-branches.s3-website.eu-west-2.amazonaws.com/feature-time-series-attributes/ The rest of the site looking and working as intended -OB

brockfanning:
@otis-bath @LucyGwilliamAdmin The next steps for this ticket are:

1. Decide if any other columns should be treated as "time-series attributes". To be clear, the list of time-series attributes is configurable, but we need to decide on the defaults. Right now there is only one default: "COMMENT_TS". If we want to add any more, we should decide that now. I think we had discussed date of last data update, and if we are going to follow the SDMX model then I guess that would be "DATA_LAST_UPDATE", correct?
2. After we decide on the defaults, we need to decide on the label to use for each one, in the footer. If we go with what is in the SDMX DSD, this is what they would be: "Time series-level footnotes" and "Date of the most recent change of the series.".
3. After we decide on the labels, we need to put them in SDG Translations so they can be translated.

brockfanning:
@otis-bath @LucyGwilliamAdmin The next steps for this ticket are:

1. Decide if any other columns should be treated as "time-series attributes". To be clear, the list of time-series attributes is configurable, but we need to decide on the defaults. Right now there is only one default: "COMMENT_TS". If we want to add any more, we should decide that now. I think we had discussed date of last data update, and if we are going to follow the SDMX model then I guess that would be "DATA_LAST_UPDATE", correct?
2. After we decide on the defaults, we need to decide on the label to use for each one, in the footer. If we go with what is in the SDMX DSD, this is what they would be: "Time series-level footnotes" and "Date of the most recent change of the series.".
3. After we decide on the labels, we need to put them in SDG Translations so they can be translated.

LucyGwilliamAdmin:
@otis-bath @brockfanning  I thought for the naming we were thinking just leaving as 'Footnotes' and then having'Date data last updated'

I'm sure @phillipgardiner also has some thoughts on wording so tagging!

Also is it worth testing with data and metadata controls (if not already done)?

otis-bath:
Okay great sounds good, will let Phill input tomorrow but think this makes more sense so thanks for flagging

brockfanning:
@LucyGwilliamAdmin The only thing new in this PR is these `time_series_attributes`. We could definitely test this in combination with the indicator configurations [`footer_fields`](https://open-sdg.readthedocs.io/en/latest/indicator-configuration/#footer_fields), [`data_footnote`](https://open-sdg.readthedocs.io/en/latest/indicator-configuration/#data_footnote), [`copyright`](https://open-sdg.readthedocs.io/en/latest/indicator-configuration/#copyright), and [`computation_units`](https://open-sdg.readthedocs.io/en/latest/indicator-configuration/#computation_units) -- but I wouldn't expect there to be any technical problems, because this PR only adds functionality without effecting existing functionality.

That said, if we do use "Footnote" as the label for COMMENT_TS then testing in combination with `data_footnote` will result in two footer items with the same label. So a feature branch with `data_footnote` populated might be a good idea to get that conversation started.

In the meantime I've updated the jekyll-open-sdg-plugins PR to add DATA_LAST_UPDATE to the defaults, and for labels I'm using a couple that we already have translations for:
* ["Footnote"](https://github.com/open-sdg/sdg-translations/blob/1.8.0-dev/translations/en/indicator.yml#L31)
* ["Data last updated"](https://github.com/open-sdg/sdg-translations/blob/1.8.0-dev/translations/en/metadata_fields.yml#L18)

I also forgot I need to try and make this work without requiring chartjs_3.

brockfanning:
@otis-bath If you change [this label](https://github.com/ONSdigital/sdg-indicators/blob/feature-time-series-attributes/_config.yml#L269) to "Footnote", we'll see that case where "Footnote" appears twice in the footer.

My questions are:
1. Is this possibility of two footnotes a problem?
2. If so, is it an acceptable fix to have the default label be slightly different? (eg, "Footnote (time-series)" or something like that)

Also a side note - this should now work regardless of whether chartjs_3 is true or false.

phillipgardiner:
@LucyGwilliamAdmin @otis-bath @brockfanning happy to provide some thoughts around wording..

Personally, I don't think having 'Footnotes' appear twice is ideal. The primary reason being, it makes users unsure if it should be there or if its a mistake. I think the solution to this is related to the context and what / how much information is likely to appear in each of the footnotes, here's my thinking...

1) Footnote (time-series) - I believe anything like this is a bit too technical. This may make sense to expert users, but our lesser experts will not necessarily know what time series is, and could cause more confusion and questions than help. 

2) We could potentially (but don't know if its possible) Change the first footnote title to be 'Footnote(s), then have no title for the second one (e.g.) 

Footnote(s)  First footnotes comments here
                    Second footnotes comments here

Like I say we may not even be able to do this but a weakness of this approach is it may remove visibility of the second footnote changing as users select different time series.

3) We could possibly call the first' Footnote' and the second 'Footnote1' (e.g.)

Footnote    First footnotes comments here
Footnote1  Second footnotes comments here

This may work as it will make it look like a list and would support when only the second footnote is present on its own (as it would view as if 'Footnote1' is the first and only. Where this may fail is depending on how much / what information is likely to appear in 'footnote'. If this ends up being a list of multiple things then it would be strange to have a list (e.g. more than one item) then have something underneath saying Footnote 1, when in fact it is the 4th or 5th thing mentioned.   

So that's a couple of things for further discussion, please feel free to think out loud on here....

brockfanning:
@phillipgardiner I agree that showing two "Footnotes" is not ideal. I also think it's not unreasonable to argue that the _possibility_ of showing two "Footnotes" is OK.

Assuming we give the new field a default name of "Footnote", there would be two ways to get "Footnote" into the footer:

1. Using the `data_footnote` indicator configuration setting (already implemented)
2. Using the `COMMENT_TS` data column (implemented in this PR)

(the above is exactly what @otis-bath is demonstrating in the FB)

The argument (that the _possibility_ of showing two "Footnotes" is OK) could be something like:

1. Countries have full control of whether they use `data_footnote`, or `COMMENT_TS`, or both/neither
2. Countries would rarely want to use both on the same indicator

For example in feature branch's [9.c.1](http://uk-sdg-feature-branches.s3-website.eu-west-2.amazonaws.com/feature-time-series-attributes/9-c-1/) the `data_footnote` indicator configuration is being used to display this catch-all statement: "The original source of these data varies across the time series. Please see the source data for full footnotes.". The `COMMENT_TS` column, presumably, would be what "Please see the source data for full footnotes" refers to. So, if the `COMMENT_TS` were actually being displayed in the footer (which is what this PR accomplishes), then that catch-all statement might be unnecessary. (At least that could be argued.)

I'm trying to think of a use-case for displaying both, and I think it might be something like this:

1. The `data_footnote` has general text that applies to every row of data
2. The `COMMENT_TS` has extra stuff that only applies to certain rows of data

I'd suspect that the labels in this case would probably need to convey that somehow. Like maybe "Footnote" and "Footnote (for displayed data)". Or maybe "Footnote" and "Special footnote".

Side note: The country can easily control the label of the second one (the `COMMENT_TS` label`). The label for `data_footnote` can also be overridden, just not as easily.

If we decide that the possibility of two "Footnotes" is a problem, and we want to keep "Footnote" as the label for both `COMMENT_TS` and `data_footnote`, and we also decide that countries would never need both at the same time, then we could take a different approach with this PR. We could have the `COMMENT_TS` actually _replace_ the `data_footnote` when it is present. But that seems a little risky in that it removes some flexibility and maybe there is a use-case for needing both that we hadn't considered.

phillipgardiner:
@LucyGwilliamAdmin @otis-bath @brockfanning , hey @brockfanning thanks for the further explanation and context. I guess my comments were made more from a pure UX perspective. However a couple of things jumped out for me from the above statement:

- Countries would rarely want to use both on the same indicator
- Side note: The country can easily control the label of the second one (the COMMENT_TS label)

Putting a product owner type hat on alongside my UX one. If we believe that countries rarely will be using both and even if they do they have the option of controlling the label, then I think having footnote x 2 is acceptable. Moving away from this here and now would be very low priority.

PR #1576, opened 2022-03-23
Specificity fix for chart focus in high-contrast

brockfanning:
Quick fix of chart background color when focused (clicked on) in high-contrast mode. The chart should only have a yellow outline; not a full yellow background.


otis-bath (approved):
Great thanks Brock, this is working as well -OB

PR #1577, opened 2022-03-23
Styling for data notices in bootstrap 5

brockfanning:
Quick fix to add styling for data notices if using Bootstrap 5.

### DIFF
--- a/_sass/bootstrap5/layouts/_indicator.scss
+++ b/_sass/bootstrap5/layouts/_indicator.scss
@@ -5,6 +5,7 @@
         padding-top: 10px;
     }
 
+    &.contrast-high canvas,
     canvas {
         @include noselect;
         &:focus {


otis-bath:
Great thanks @brockfanning - this has worked and the exclamation mark is now the correct size, know this is a small detail but can we have it to the left of the text as on the live site as currently quite a lot of white space? See pictures: 
image
image

brockfanning:
@otis-bath Can you try removing the override of _includes/components/indicator/data-notice.html? I'm curious how that would look, and wondering if maybe the override is no longer needed.

otis-bath:
So have this on this test FB without the override and it looks perfect so don't think we need this any more http://uk-sdg-feature-branches.s3-website.eu-west-2.amazonaws.com/feature-site-bootstrap-5-notice-test/archived-indicators/1-a-3-archived/

otis-bath (approved):
The combination of the updated archived indicators code and removal of the data-notice.html file has produced the desired result - OB

PR #1578, opened 2022-03-24
Recover rescaler

brockfanning:
This recovers the "rescaler" plugin which had been removed. This plugin is responsible for "cropping" the chart datasets so that empty years at the beginning/end of each dataset are not shown. This ensures that the chart will take up all the available space.

This PR also adds a bit of documentation and simplifies the code a bit, to avoid confusion in the future.

This show work whether `chartjs_3` is true or false.


otis-bath (approved):
Great thanks Brock, have tested this with chartjs_3 on and off and has worked in both instances http://uk-sdg-feature-branches.s3-website.eu-west-2.amazonaws.com/feature-site-recover-rescaler/ -OB

PR #1581, opened 2022-03-25
Bootstrap5 accessible tabs fix

brockfanning:
Quick fix for accessible tabs if bootstrap_5 is enabled. Without this fix users cannot use the tab key to navigate to the Chart/Table/etc tabs.


brockfanning:
I don't believe this warrants dedicated testing so will merge.

PR #1582, opened 2022-03-26
Keep sr-only class to avoid unneeded file changes

brockfanning:
The recent Bootstrap 5 upgrade changed a lot of files, but a large proportion of them were only changing the "sr-only" class to "visually-hidden". We can just keep support for "sr-only" with a few lines of CSS, and not have to change a lot of files. This should make the upgrade process easier for users.


brockfanning:
I don't believe this warrants dedicated testing, so will merge.

PR #1649, opened 2022-06-02
Map disaggregation in leaflet

### LINKED ISSUE #1505: Map tab to react to series/units being changed
### Discussed in https://github.com/open-sdg/open-sdg/discussions/1456

<div type='discussions-op-text'>

<sup>Originally posted by **LucyGwilliamAdmin** July 27, 2021</sup>
If an indicator has a map and more than one series, the map tab doesn't react to the series being changed. However, the title changes.

This can be misleading as users could think that the map is showing data for the series but it's not (unless it's the first series)

When the series is changed, the map should change to reflect the changes in data

So that users aren't misled in thinking that the map is showing something it isn't

Depending on how long this may take and difficulty, maybe something quicker/easier could be implemented in short term e.g. that makes it clear that the map is showing the first series.</div>

brockfanning:
Q | A
--- | ---
Fixed issues | Fixes #1505 
Related version | 2.0.0-dev
Bugfix, feature or docs? |
* [ ] Includes tests
* [x] Updated docs
* [x] Updated config forms
* [ ] Added CHANGELOG entry

Yet another approach to map disaggregation. This has been discussed in #1456 but needs design/UX feedback and more testing.


christoph-LSN:
@brockfanning sorting works!

anything else what I can look for?

image

brockfanning:
@christoph-LSN Thanks for testing! I can't think of anything else at the moment but I will be sure to let you know if some more testing could help get this finished.

christoph-LSN:
@brockfanning ok...
Another point is the timeslider. I have many long time seires, almost starting in 2005. Can I set another starting year, like the data start values for the disaggergations?

brockfanning:
@christoph-LSN Would it solve the problem if it started on the most recent year?

christoph-LSN:
@brockfanning Yes, that would be great!

brockfanning:
@christoph-LSN That certainly seems reasonable to me. Do you mind starting a new discussion about that on https://github.com/open-sdg/open-sdg/discussions?

christoph-LSN:
@brockfanning 

The change breakdows button is visible if the units colum exists

https://christoph-lsn.github.io/MT_Site/1-1-3/

image

brockfanning:
@christoph-LSN Thanks - I've changed it so that it should only display the button and form if there are disaggregations or at least 2 series or units.

brockfanning:
@christoph-LSN Do you happen to have any indicators with extremely long lists of disaggregation values, or a large number of types of disaggregations? I'm curious to see what happens when there is too much content in the popup. Hopefully it gives you a scrollbar but I'm not sure.

christoph-LSN:
@brockfanning 

have a look at this indicator: https://christoph-lsn.github.io/MT_Site/4-1-1/
I Thinks this is one with the most disaggregation values

brockfanning:
@christoph-LSN The latest change should hopefully help with mobile and long lists of disaggregations. But note that the "Change breakdowns" button is now opt-in, so it won't appear unless you have a certain site configuration. The site configuration is the `disaggregation_controls` option within the `map_options` setting. For example:

```
map_options:
  disaggregation_controls: true
  ...
```

Without this setting, the current disaggregations will still display, but the "Change breakdowns" button won't display.

brockfanning:
@otis-bath Given that this is opt-in now, I think it would be appropriate to do a feature-branch. No config change is needed - the expected change is that a box should appear on maps showing the series/unit/disaggregations that the data reflect.

otis-bath:
Hi @brockfanning - The box appears on the FB: http://uk-sdg-feature-branches.s3-website.eu-west-2.amazonaws.com/feature-site-map-disaggregation-in-leaflet/3-3-3/
Have noticed this has introduced variable line thickness on the goals page between the indicators? Otherwise the box is working.
@phillipgardiner Do you have any thoughts on the box design?

brockfanning:
I can't think of why this would have any effect on goal pages. The only style changes here are limited to maps.

I do see a couple of things to dig into:
1. How should it change with the zoom changes to a different set of boundaries? (ie changing from country boundaries to region boundaries)
2. Why does "Country: England" appear, when the map is displaying regions outside of England?

otis-bath:
@brockfanning - goal pages issue is now resolved, appears to have just been a weird visual bug that hasn't repeated since. Agree we need workarounds for these other issues.

brockfanning:
@otis-bath @christoph-LSN If you are able to test again, feedack is appreciated. The latest change is intended to add this:
1. Updates the info when the visible layer changes by zooming in or out
2. "Prune" the subcategories by removing any that are identical across all of the variations for one part of the map. This one is tough to explain, but I can describe the intended effect. This should remove things like "Region" and "Country" that are merely reflecting a part of the map. It will also have the side-effect of removing duplicative subcategories like (in the case of the UK) "Observation status" and "Unit multiplier".

otis-bath:
Hi @brockfanning - have refreshed our FB in the UK. Getting some slightly unusual behaviour, clicking on the regions is bringing them up in the top RHS however the values are overlapping the text for the region. 
http://uk-sdg-feature-branches.s3-website.eu-west-2.amazonaws.com/feature-site-map-disaggregation-in-leaflet/3-3-3/
image

brockfanning:
@otis-bath I think that problem is unrelated to this change - can you compare with the live site? Definitely an issue that needs fixing.

otis-bath:
@brockfanning - ah it is a problem on the live site. This is working properly then as behaves the way you intended. I'll open a separate issue. Was there anything else for this PR or can it be approved?

christoph-LSN:
@brockfanning @otis-bath 
I have this overlapping too.....
The Unit is dissapeared and where can I translate "Apply", "Cancel" and "Change breakdowns" ?
image

brockfanning:
@christoph-LSN I just tweaked it to hopefully get the units back. Can you give it another try?

I've also made the buttons translated, however in order to see the translations you would need to be pointing at the latest translations in SDG Translations (the branch 2.0.0-dev). Here are those translations: https://github.com/open-sdg/sdg-translations/blob/2.0.0-dev/translations/de/indicator.yml#L81-L83

Alternatively if you have a local "indicators.yml" translation file then you could copy those items into your file.

Note that if the translations are not present then these buttons will appear empty.

P.S. Happy to adjust the German there, as I'm sure those automated translations are not exactly right.

christoph-LSN:
@brockfanning I have added the translation im my local translations folder indicator.yml https://github.com/christoph-LSN/IM_Data/blob/f5106eea84fed07ce38b7bb7e4b9354f908f659a/translations/de/indicator.yml#L48

The units are visible again!

but it shows undefined

image

christoph-LSN:
@brockfanning sorry i have been to fast.... translations changed as well ;-)

image

brockfanning:
@christoph-LSN Thanks for testing!

@otis-bath @LucyGwilliamAdmin @phillipgardiner We've resolved all the known issues in the PR, and so this could conceivably be added to 2.0.0, barring any objections. To sum up the changes:
1. A box will appear at the bottom left of maps which displays the disaggregations/units/series in the mapped data
2. Optionally a "Change breakdowns" button can be enabled to allow the end-users to change the disaggregations/units/series and see the change on the map. By default this button does not appear.

As we've discussed, the first change above is a straightforward enhancement of the maps, because currently the breakdowns in the mapped data are hidden from the end-user. The second change above is more experimental, but a rationale for including it in 2.0.0 might be that - since it is optional - it is OK to release.

LucyGwilliamAdmin:
@brockfanning @otis-bath  I do have one comment on this - so if you look on the default page [here](http://uk-sdg-feature-branches.s3-website.eu-west-2.amazonaws.com/feature-site-map-disaggregation-in-leaflet/3-3-3/), the default series is Rate per 100,000 pop - Chart and chart footer all look fine:
image

However, go to the map tab - you see that the series that shows on the map is 'Number of cases' (maybe because the order is being controlled for chart data and not for map data) but this does mean the the map footer is incorrect - it's pulling through the footer for the rate per 100,000 pop series rather than the Number of cases series:

image

brockfanning:
@LucyGwilliamAdmin Great catch. Everything inside the map is correct, and the only difference there is that now the series/unit is being displayed in the box.

But the footer _below_ the map is dynamically controlled by the disaggregation controls in the chart/table tabs. This dynamic behavior was added relatively recently, but unfortunately causes a problem here, since it is incorrect when showing the map tab, as you point out.

So basically this PR is highlighting an existing problem - that the unit of measurement shown in the footer below the map does not necessarily match what is actually on the map.

As for solutions, my first thought is to try to hide the unit of measurement in the footer when the map tab is active. I'll see how hard that is to do.

otis-bath:
@brockfanning @LucyGwilliamAdmin - have refreshed this and should have sorted the footer:
http://uk-sdg-feature-branches.s3-website.eu-west-2.amazonaws.com/feature-site-map-disaggregation-in-leaflet/3-3-3/

Have noticed the download GeoJSON buttons have grouped together, do we need some padding between them here?

brockfanning:
@otis-bath You're right, that probably happened with Bootstrap 5. Though I think that should be done in a separate ticket since it's unrelated to these changes.

otis-bath (approved):

christoph-LSN:
@brockfanning @MoBosse @otis-bath 

One last thing: 
[example data](https://github.com/christoph-LSN/IM_Data/blob/develop/data/indicator_1-3-2.csv)
[example indicator](https://christoph-lsn.github.io/MT_Site/1-3-2/)

In indicator 3-1-2 I have different disaggregations depending on the units column.
Selecting "in 1000" you can select "Migrationserfahrung" and "Nationalität" 
Selecting "%" you can select  "Migrationserfahrung" only

on the map dissagregation only "in 1000" is displayed and "Migrationserfahrung shows "All"

image

brockfanning:
@christoph-LSN Since this PR has been merged, any new bugs should be posted in new issues.

Issue #1593, opened 2022-04-07
Identify self-tested accessibility fixes


LucyGwilliamAdmin:
Indicator pages:
1. [ ] The "Sub-categories" label is an H4, and I don't believe its parents have H3, H2, H1. Lighthouse points to this page: https://web.dev/heading-order/?utm_source=lighthouse&utm_medium=devtools
2. [x] The chart footer items (a `<dl>` element) has an empty `<dl>` inside of it. Lighthouse points to this page: https://web.dev/definition-list/?utm_source=lighthouse&utm_medium=devtools
3. [ ] The chart canvas, the active data tab, and the active metadata tab are giving this: https://developer.mozilla.org/en-US/docs/Web/Accessibility/Understanding_WCAG/Keyboard#focusable_elements_should_have_interactive_semantics
4. [x] The “Clear selections” does not meet color contrast
5. [ ] The goal icon on indicator pages doesn't have a tab focus outline
6. [ ] When I switched to the iframe tab (as opposed to "Chart" and "Table") the screenreader didn't say anything.
7. [ ] There is no screenreader feedback when a table column is sorted
8. [x] There is no screenreader feedback when data is added using the disaggregation controls (I believe we had that working before, so might be a regression bug)

Site-wide:

1. [ ] Anything using a Font Awesome icon generates and error related to this:
https://developer.mozilla.org/en-US/docs/Web/Accessibility/Understanding_WCAG/Keyboard#clickable_elements_must_be_focusable_and_should_have_interactive_semantics. Examples: search bar icon, language toggle icon, disagg clear button icon
Also occurring on the <div id=”indicatorData”> element. These may be a false positive - need to investigate.
2. [x] The “Cookie settings” link is giving this: https://developer.mozilla.org/en-US/docs/Web/Accessibility/Understanding_WCAG/Keyboard#interactive_elements_must_be_able_to_be_activated_using_a_keyboard (I also noticed that the "Cookie settings" item cannot be tabbed to, so definitely needs fixing)
3. [ ] There is no screenreader feedback when the contrast toggle is used
4. [x] Screenreader misreads Goal 8 as ‘Decent Jobs and Economic Growth’ instead of 'Decent Work and Economic Growth' - PR opened: https://github.com/open-sdg/sdg-translations/pull/277
[...]

PR #1662, opened 2022-06-09
Various chart-related fixes

### LINKED ISSUE #1656: No screenreader feedback when using disaggregation controls on chart/table
Site repository URL: N/A
Open SDG version: 2.0.0-dev

**Describe the bug**

Previously we had functionality where screenreaders would announce any changes to the chart/table after the disaggregation controls had been used, but this seems to have stopped working.

**To Reproduce**
1. Go to any indicator with disaggregation
2. Make sure a screenreader is running
3. Change something with the disaggregation controls
4. Note there is no announcement

**Expected behavior**

Each change should have an informative announcement.

**Additional context**

This was previously implemented in #875

brockfanning:
This should fix a few random chart-related things that I've noticed:

## Avoids an infinite loop that was happening when the no-data message was clicked on

To replicate:

1. Open the developer console
2. Go to an indicator in which you can get the no-data message (ie, "This data is not available. Please choose alternative data to display.")
3. Click on the no-data message
4. Wait a few seconds and notice the infinite loop in the developer console

## Proper color contrast for the no-data message

To replicate:

1. Switch to high contrast
2. Go to an indicator in which you can get the no-data message (ie, "This data is not available. Please choose alternative data to display.")
3. Notice that the no-data message is not visible

## Fix for chart updates to allow the screenreader announcement (and other things) to happen

To replicate:

1. Turn on a screenreader
2. Go to an indicator with disaggregation
3. Tick any disaggregation checkbox
4. Notice there is no screenreader announcement

In addition to testing the three items above, I would recommend a general spot-checking of a few charts to make sure that things generally still work. The reason I say that is that, due to a bug, a whole chunk of code was being skipped. That code is now no longer being skipped, and so I want to make sure it doesn't introduce a problem.

Q | A
--- | ---
Fixed issues | Fixes #1656 
Related version | 2.0.0-dev
* [ ] Includes tests
* [ ] Updated docs
* [ ] Updated config forms
* [ ] Added CHANGELOG entry

This is also mentioned in #1593

### DIFF
--- a/_includes/assets/js/chartjs/accessibleCharts.js
+++ b/_includes/assets/js/chartjs/accessibleCharts.js
@@ -95,6 +95,10 @@ Chart.register({
         return isEmpty;
     },
     activateNext: function() {
+        // Abort early if no data.
+        if (this.chart.data.datasets.length === 0) {
+            return;
+        }
         this.selectedIndex += 1;
         if (this.selectedIndex >= this.meta.data.length) {
             this.selectedIndex = 0;
@@ -110,6 +114,10 @@ Chart.register({
         this.activate();
     },
     activatePrev: function() {
+        // Abort early if no data.
+        if (this.chart.data.datasets.length === 0) {
+            return;
+        }
         this.selectedIndex -= 1;
         if (this.selectedIndex < 0) {
             if (this.chart.config.type !== 'line') {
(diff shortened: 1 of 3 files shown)


otis-bath (approved):
@brockfanning Have this on a FB here: http://uk-sdg-feature-branches.s3-website.eu-west-2.amazonaws.com/feature-site-screenreader-announcement-of-disaggregation-contro/
The no data message no longer produces the infinite loop and appears fully in the high contrast mode, also the screenreader is now reading properly and I have checked a number of different graphs and functionality appears to be preserved as normal. Thanks

PR #1670, opened 2022-06-13
Hotfix for 1.8.0 - Pin to 2.6.1 of json-editor

brockfanning:
This is a duplicate of #1660 but for 1.8.0, so that we can release a hotfix for that.

### DIFF
--- a/_layouts/config-builder.html
+++ b/_layouts/config-builder.html
@@ -35,7 +35,7 @@
   <div class="row" id="results-status" role="status"></div>
   <div class="row" id="editor_holder"></div>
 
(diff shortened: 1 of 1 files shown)


brockfanning:
@otis-bath This should be tested with a 1.8.0 site. In other words, if doing a feature branch, you can use the UK platform as-is, only pointing to this `brockfanning/open-sdg@jsoneditor-fix-for-1.8.0` instead of `open-sdg/open-sdg@1.8.0`. This fixes #1673.

I believe you will also need to enable the site config form to test this:

```
site_config_form:
  enabled: true
```

otis-bath:
Great thanks for this Brock - I am going to demonstrate the review process to Sarah and since she found this error will be a good one to show

otis-bath (approved):
This works thanks Brock

PR #1676, opened 2022-06-14
Remove some chart-related code that is no longer used as of 2.0.0

brockfanning:
Q | A
--- | ---
Fixed issues | N/A
Related version | 2.0.0-dev
Bugfix, feature or docs? |
* [ ] Includes tests
* [ ] Updated docs
* [ ] Updated config forms
* [ ] Added CHANGELOG entry

This PR removes some code related to charts which was unintentionally left in during the switch to 2.0.0 (and Chart.js 3).


brockfanning:
@otis-bath This is yet another ticket removing stuff that I forgot to remove before. The main thing to test here is that charts still work as expected. It should not affect anything because the code being removed here is not being used.

otis-bath (approved):
Looks good charts behaving normally - OB

PR #1680, opened 2022-06-15
Selection legend revamp moving name to the left

### LINKED ISSUE #1677: Low values in selection legend overlap poorly with text names
When on a map, if you click on the regions the value is displayed in the top RHS next to the key for the upper and lower limits of the chart. However the text and values do not overlap cleanly leading to the region, value and the 'x' mark often looking messy. To reproduce load this map: https://sdgdata.gov.uk/3-3-3/ and click on some of the regions on the map. 

image

brockfanning:
Q | A
--- | ---
Feature branch/test site URL | [Link](http://uk-sdg-feature-branches.s3-website.eu-west-2.amazonaws.com/feature-site-selection-legend-text-to-the-left/3-3-4/)
Fixed issues | Fixes #1677 
Related version | 2.0.0-dev
* [ ] Includes tests
* [ ] Updated docs
* [ ] Updated config forms
* [ ] Added CHANGELOG entry

This changes the markup and styling of the "selection legend" on maps, so that the names of regions appear to the left of the horizontal bar. It also changes from `<ul>` tags to `<dl>` tags, which is a bit more semantically correct.

### DIFF
--- a/_includes/assets/js/plugins/leaflet.selectionLegend.js
+++ b/_includes/assets/js/plugins/leaflet.selectionLegend.js
@@ -66,30 +66,45 @@
 
     update: function() {
       var selectionList = L.DomUtil.get('selection-list');
-      var selectionTpl = '' +
+      var selectionTplHighValue = '' +
         '<dt class="selection-name"><span class="selection-name-background">{name}</span></dt>' +
         '<dd class="selection-value-item {valueStatus}">' +
           '<span class="selection-bar" style="width: {percentage}%;">' +
-            '<span class="selection-value">{value}</span>' +
+            '<span class="selection-value selection-value-high">' +
+              '<span class="selection-value-high-background">{value}</span>' +
+            '</span>' +
           '</span>' +
           '<i class="selection-close fa fa-remove"></i>' +
         '</dd>';
+      var selectionTplLowValue = '' +
+      '<dt class="selection-name"><span class="selection-name-background">{name}</span></dt>' +
+      '<dd class="selection-value-item {valueStatus}">' +
+        '<span class="selection-bar" style="width: {percentage}%;"></span>' +
+        '<span class="selection-value selection-value-low" style="left: {percentage}%;">' +
+          '<span class="selection-value-low-background">{value}</span>' +
+        '</span>' +
+        '<i class="selection-close fa fa-remove"></i>' +
+      '</dd>';
       var plugin = this.plugin;
       var valueRange = this.plugin.valueRange;
       selectionList.innerHTML = this.selections.map(function(selection) {
         var value = plugin.getData(selection.feature.properties);
         var percentage, valueStatus;
+        var templateToUse = selectionTplHighValue;
         if (value) {
           valueStatus = 'has-value';
           var fraction = (value - valueRange[0]) / (valueRange[1] - valueRange[0]);
           percentage = Math.round(fraction * 100);
+          if (percentage <= 50) {
+            templateToUse = selectionTplLowValue;
+          }
         }
         else {
           value = '';
           valueStatus = 'no-value';
           percentage = 0;
         }
-        return L.Util.template(selectionTpl, {
+        return L.Util.template(templateToUse, {
           name: selection.feature.properties.name,
           valueStatus: valueStatus,
           percentage: percentage,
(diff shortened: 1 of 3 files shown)


christoph-LSN:
@brockfanning 
Looks good to me. https://christoph-lsn.github.io/MT_Site/3-1-2/
may be you can change the backgroung color of the value field into grey as well. Now it looks like it is another field in the bar

image

phillipgardiner:
@brockfanning think it looks a bit odd having a white bar? then the grey background to the number. Is this supposed to be like a bar graph (e.g. higher the number, longer the bar is?). If so I would consider making them the same colour.

I also notice if you select something like 'Lhst. Hannover' then the number falls off the right hand side of the page and / or is covered by the 'x'.

brockfanning:
Great points all - I've made some tweaks:
* The "x" buttons appear to the right of the whole thing, so that they don't cover up any data
* The values no longer a different color from the bar
* The names have a slight overlay so that it's easier to read

@christoph-LSN Do you mind refreshing your example?
@otis-bath Would it be possible to do a feature branch to see this with the UK data?

brockfanning:
@otis-bath @christoph-LSN Sorry - hold off - there is still a problem when the value is the maximum.

christoph-LSN:
@brockfanning @otis-bath @phillipgardiner 
looks good to me: https://christoph-lsn.github.io/MT_Site/3-1-2/

image

LucyGwilliamAdmin:
@brockfanning @christoph-LSN the text is sometimes falling over the white bar i.e. in the cases of Wolfenbüttel and Helmstedt - not sure what the fix is here as could make the overall bar longer so the white part will be longer but never know how long a number might be?

otis-bath:
@brockfanning @LucyGwilliamAdmin @phillipgardiner - Have updated this for the UK data, have the same problem of small data overlapping part of the bar which is difficult to read 
image

christoph-LSN:
@LucyGwilliamAdmin @brockfanning do you mean the legend entry (Wolfenbüttel) or the value?
image
may be it is possible to put the value on top of the bar 
like in my excel example
image

LucyGwilliamAdmin:
@christoph-LSN so do you mean having it just after the white part on the grey?

christoph4711:
Yes…

Von meinem iPhone gesendet

brockfanning:
The trick is to be able to handle both very high values and very low values. What if the value appears after the bar when the value is less than 50%, and the value appears inside the bar when the value is more than 50%? Example:
Screenshot from 2022-06-16 09-23-31

brockfanning:
@christoph-LSN Thanks - I can try making the available space wider to help avoid that.

christoph-LSN:
@brockfanning Great!

image

otis-bath:
@brockfanning - looks good on my side and functionality all working

@phillipgardiner - are you happy with this change? Any feedback?
http://uk-sdg-feature-branches.s3-website.eu-west-2.amazonaws.com/feature-site-selection-legend-text-to-the-left/3-3-3/

brockfanning:
@LucyGwilliamAdmin @phillipgardiner Floating one last idea in case it's desirable - what if the color of the bars matched the color on the map, instead of being white? That's an easy change and looks like this:
Screenshot from 2022-06-16 13-21-59

christoph-LSN:
> 



> just a quick question, for you what do those additional bars add to the map functionality over being able to view the figures on the map?
> 
> Just trying to understand if that part of the map is really needed?

@LucyGwilliamAdmin it is not really necessary, but it is nice to have. you can comapre regions that are far away from each other, so you might zoom in looking at one region but can't see the oher anymore...

like Grafshaft Bentheim here:

image

phillipgardiner:
@brockfanning @LucyGwilliamAdmin I guess adding colours wouldn't hurt as it may help convey the relativeness between the different figures and tie it back to the scale, in addition makes it stand out more. Note: I'd imagine contrast of the colours etc is an issue for accessibility but I think that is a potential issue across maps full stop.

LucyGwilliamAdmin:
@christoph-LSN thanks for clarifying - that makes sense!

@phillipgardiner @brockfanning ok - let's add the colours and then think we're good to go on this one!

brockfanning:
Sounds good. @otis-bath I've pushed up one last tweak to this PR.

christoph-LSN:
@brockfanning looks good to me !

image

otis-bath (approved):
Looks good thanks Brock -OB

Issue #1684, opened 2022-06-16
Hidden 'target line' and 'series break' labels


phillipgardiner:
Both the 'target line' and the 'series break' graph labels have the potential to be covered by the graph data making them unreadable. 

Q: Is the positioning of the labels able to be changed?

If not consider bring labels to font, using alternative colours etc

![image](https://user-images.githubusercontent.com/60919820/174040932-3d10ff8a-f136-4884-8c09-2a14a69d279c.png)
![image](https://user-images.githubusercontent.com/60919820/174040974-3824ce07-4d61-41a7-a807-581889797572.png)

PR #1696, opened 2022-06-20
Draw annotations after datasets

### LINKED ISSUE #1683: Target line appearing behind graph elements
**Describe the bug**
On a graph when a target line is present the target line appears hidden behind the data points. This makes it hard to see the target line in question.

**Expected behavior**
The target line should appear in front of the data points so as not to be hidden

**Screenshots**
image

**Additional context**
Note: This may be relevant for the 'series break' line also

brockfanning:
Q | A
--- | ---
Feature branch/test site URL | [Link](http://uk-sdg-feature-branches.s3-website.eu-west-2.amazonaws.com/feature-draw-annotations-after-datasets/)
Fixed issues | Fixes #1683 
Related version | 2.0.0-dev
Bugfix, feature or docs? |
* [ ] Includes tests
* [ ] Updated docs
* [ ] Updated config forms
* [ ] Added CHANGELOG entry

@phillipgardiner @otis-bath @LucyGwilliamAdmin This could help with #1683 and possibly #1684 too. If you would like, this could be added for 2.0.0 as a bug.

I would recommend testing this with at least four scenarios:
1. Line chart with a target line
2. Bar chart with a target line
3. Line chart with a series break
4. Bar chart with a series break

It's possible this may prompt other changes. Here is what it looks like locally with the 2nd scenario above:

image

### DIFF
--- a/_includes/assets/js/view/chartTypeBase.js
+++ b/_includes/assets/js/view/chartTypeBase.js
@@ -183,7 +183,7 @@ opensdg.chartTypes.base = function(info) {
                 options: {
                     plugins: {
                         annotation: {
-                            drawTime: 'beforeDatasetsDraw',
+                            drawTime: 'afterDatasetsDraw',
                             annotations: annotations
                         }
                     }


otis-bath:
Hi @brockfanning @phillipgardiner @LucyGwilliamAdmin - apologies this has taken longer than planned as GitHub Actions has been down, would be interested to hear any thoughts/ideas on this and can always discuss at tech call. Have this on bar charts ([6-5-1](http://uk-sdg-feature-branches.s3-website.eu-west-2.amazonaws.com/feature-draw-annotations-after-datasets/6-5-1/) [3-d-1](http://uk-sdg-feature-branches.s3-website.eu-west-2.amazonaws.com/feature-draw-annotations-after-datasets/3-d-1/) and on line charts [1-1-1](http://uk-sdg-feature-branches.s3-website.eu-west-2.amazonaws.com/feature-draw-annotations-after-datasets/1-1-1/) [1-2-1](http://uk-sdg-feature-branches.s3-website.eu-west-2.amazonaws.com/feature-draw-annotations-after-datasets/1-2-1/)

Think the bar charts look fine and are generally thick enough for the white box to work however with the line chart can hide data points which then only become readable by hovering over them - not sure if there's a way around this that keeps the label on the graph but would be keen to hear any ideas.

PR #1697, opened 2022-06-20
Default to accessibile chart colorset

### LINKED ISSUE #1681: Graph colour set documentation
On the page: https://open-sdg.readthedocs.io/en/latest/configuration/. The documentation suggests that the default graph colours used are a deprecated 6 color set. I believe this is wrong and the default is now the accessible set. Documentation requires updating to reflect this. 

graph_color_set: 'accessible' a 6-color set that is specifically chosen for optimal accessibility (recommended)
graph_color_set: 'default' a deprecated 6-color set that is still the default (for reasons of backwards compatibility)

brockfanning:
Q | A
--- | ---
Feature branch/test site URL | [Link](http://uk-sdg-feature-branches.s3-website.eu-west-2.amazonaws.com/feature-site-default-to-accessible-chart-colors/)
Fixed issues | Fixes #1681 
Related version | 2.0.0-dev
Bugfix, feature or docs? |
* [ ] Includes tests
* [x] Updated docs
* [ ] Updated config forms
* [ ] Added CHANGELOG entry

@phillipgardiner @otis-bath @LucyGwilliamAdmin This actually is not just a documentation issue - we actually forgot to make "accessible" the new default. So I think this should go into 2.0.0, if possible.

The PR changes "accessible" to be the default, and also changes the old "default" colorset to be called "classic". If a site was previously configured to be "default", with this PR that site will now display the accessible colors. Sites who want to continue using the old default can configure their site to "classic".


otis-bath:
@brockfanning - have taken a look at this and the default and classic options both working as intended

PR #1709, opened 2022-06-22
Translate the chart tooltip labels

### LINKED ISSUE #1708: Tooltip label not being translated
It seems just to be us having this issue (can't see it on other Open SDG sites), but this is an annoying one :) 

image
See example: https://sdg.hagstova.fo/sdg-site/1-2-1/

As you can see, the country label in the legend is translated into Faroese ("Føroyar") but the series hover label is not ("data.faroeislands"). 

In our site_config.yml we have added:
image

We have added the translation in this file: https://github.com/hagstovan/sdg-data/blob/develop/translations/fo/data.yml 

image

Do you have any tip why it is not working?

Our site repository can be found here: https://github.com/hagstovan/sdg-site
Our data repository can be found here: https://github.com/hagstovan/sdg-data

_Originally posted by @karihj in https://github.com/open-sdg/open-sdg/discussions/1707_

brockfanning:
Q | A
--- | ---
Fixed issues | Fixes #1708 
Related version | 2.0.0-dev
Bugfix, feature or docs? |
* [ ] Includes tests
* [ ] Updated docs
* [ ] Updated config forms
* [ ] Added CHANGELOG entry

### DIFF
--- a/_includes/assets/js/view/chartTypeBase.js
+++ b/_includes/assets/js/view/chartTypeBase.js
@@ -76,7 +76,7 @@ opensdg.chartTypes.base = function(info) {
                     backgroundColor: 'rgba(0,0,0,0.7)',
                     callbacks: {
                         label: function (tooltipItem) {
-                            return tooltipItem.dataset.label + ': ' + alterDataDisplay(tooltipItem.formattedValue, tooltipItem.dataset, 'chart tooltip');
+                            return translations.t(tooltipItem.dataset.label) + ': ' + alterDataDisplay(tooltipItem.formattedValue, tooltipItem.dataset, 'chart tooltip');
                         },
                         afterBody: function () {
                             var unit = MODEL.selectedUnit ? translations.t(MODEL.selectedUnit) : MODEL.measurementUnit;


brockfanning:
@karihj This pull-request should fix that issue with the chart tooltips not getting translated. To try out this pull-request, you can change the `remote_theme` in your _config.yml file like so:

```
remote_theme: brockfanning/open-sdg@translate-tooltip-labels
```

However note that this PR is based on the latest code, which is the 2.0.0-dev branch. So in order to test this out, you would also need to make the changes needed to upgrade to 2.0.0. A draft of those instructions is available here: https://github.com/brockfanning/open-sdg/blob/2.0.0-release-docs/docs/upgrades/upgrading-2-0-0.md

As you can see from those instructions, there are a few changes needed. If you would like to try a faster way to test this PR, you can instead just override the appropriate file in your site repo. To do that, you would need to copy [this file](https://github.com/brockfanning/open-sdg/blob/translate-tooltip-labels/_includes/assets/js/view/chartTypeBase.js) into your site repo, in the same folder location (`_includes/assets/js/view/chartTypeBase.js`).

Either way, if you are able to try this out, we'd greatly appreciate if you could post your feedback here.

PR #1724, opened 2022-07-04
Map series/unit behaviors

### LINKED ISSUE #1723: Series dependent content not represented in maps
### Discussed in https://github.com/open-sdg/open-sdg/discussions/1717

<div type='discussions-op-text'>

<sup>Originally posted by **MoBosse** June 27, 2022</sup>
I just noticed that some series and unit dependent items are not present in the map:

- Titles 
  - Title of map sticks to the title, set for the first series when using `graph_titles`
  - The `indicator_name` is shown above the map when having no `graph_titles` (although `graph_title_from_series` in the site config is set to true)

-  Footnotes
   - The content of the `footer_fields` disappears compleatly when changing to the map view
   - `data_footnote` is shown under the map but is obviously not series dependent 

- Units (when there is only one unit available)
  - The content of `computation_units` disappears completely from the footer when switching to map view
  - The only way to display the unit seems to be to have a Units column in the data. The downside here is that it looks like you have the option to change it, since it is displayed with a radio button when you click the "Change Brakedowns" button. 

- Precision
   - has no impact either. Actually I would rather not have decimal places in the maps legend, but a) I wonder how the legend would look for e.g. values within a range from 0 to 1 and b) I think the tooltip values cound get decimal places if they are defined for the selected series

There are probably other elements that are relevant to maps, but these are the ones I've noticed so far.</div>

brockfanning:
Q | A
--- | ---
Feature branch/test site URL | [Link](http://uk-sdg-feature-branches.s3-website.eu-west-2.amazonaws.com/feature-map-series-unit-behaviors/3-3-3/)
Fixed issues | Fixes #1723 
Related version | 2.1.0-dev
Bugfix, feature or docs? |
* [ ] Includes tests
* [ ] Updated docs
* [ ] Updated config forms
* [ ] Added CHANGELOG entry

Testing instructions:
1. Pick an indicator that has series, units, and geocodes
3. Enable maps for that indicator's config

        data_show_map: true
4. Enable the map disaggregation option in site config

        map_options:
          disaggregation_controls: true
5. Configure graph titles, precision, and footer_fields to be dependent on the series
6. Confirm that, on the map, changing the series has the expected effects
7. Confirm that, on the map, changing the unit updates the footer field

### DIFF
--- a/_includes/assets/js/indicatorModel.js
+++ b/_includes/assets/js/indicatorModel.js
@@ -1,6 +1,7 @@
 var indicatorModel = function (options) {
 
   var helpers = {% include assets/js/model/helpers.js %}
+  this.helpers = helpers;
 
   // events:
   this.onDataComplete = new event(this);
@@ -276,6 +277,7 @@ var indicatorModel = function (options) {
         showMap: this.showMap,
         precision: helpers.getPrecision(this.precision, this.selectedUnit, this.selectedSeries),
         dataSchema: this.dataSchema,
+        chartTitles: this.chartTitles,
       });
     }
 
--- a/_includes/assets/js/indicatorModel.js
+++ b/_includes/assets/js/indicatorModel.js
@@ -276,6 +276,7 @@ var indicatorModel = function (options) {
         indicatorId: this.indicatorId,
         showMap: this.showMap,
         precision: helpers.getPrecision(this.precision, this.selectedUnit, this.selectedSeries),
+        precisionItems: this.precision,
         dataSchema: this.dataSchema,
         chartTitles: this.chartTitles,
       });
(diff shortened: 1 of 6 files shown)


otis-bath:
@brockfanning - most of this is there just finding an example that works hard to come by. So have it here: http://uk-sdg-feature-branches.s3-website.eu-west-2.amazonaws.com/feature-map-series-unit-behaviors/3-3-3/ the graph titles, precision and footer fields change based on the series and there are geocodes for the series' 'Rate per 100,000 population' and 'Number of cases'. The change breakdowns option is showing however I am unable to change out of 'Number of cases'. This is the same as with: http://uk-sdg-feature-branches.s3-website.eu-west-2.amazonaws.com/feature-map-series-unit-behaviors/3-b-1/ where there are a number of different series all using 'Percentage (%)' as the Units however the change breakdowns will not let you switch.

brockfanning:
@otis-bath Thanks for catch that - the latest commit is meant to address that.

otis-bath (approved):
Looks good -OB

PR #1742, opened 2022-07-18
Documentation fixes related to analytics and production config

brockfanning:
These documentation fixes are meant to correct the production and analytics guidance according to the latest version.

### DIFF
--- a/docs/analytics.md
+++ b/docs/analytics.md
@@ -10,14 +10,18 @@ Out of the box, Open SDG sends information to Google Analytics for certain event
 
 ## Google Analytics ID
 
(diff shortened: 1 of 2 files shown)


brockfanning:
@LucyGwilliamAdmin @otis-bath Now that the site starter is using the new structure for `_config.yml` vs `_data/site_config.yml`, I think our documentation for the production environment needs to be updated. Do these changes make sense?

otis-bath:
@brockfanning Have just read through - looks good to me

otis-bath (approved):

PR #1765, opened 2022-08-04
Maintain focus when changing series

### LINKED ISSUE #1757: (A) Radio button focus issue
**High Level (A)**

When a radio button is selected, the page automatically updates. Focus has not been managed which means that keyboard only users need to press ‘tab’ to return to the radio group. Screen reader focus is taken to the beginning of the page and must navigate back down the page to find their place.
image
For example on: https://sdgdata.gov.uk/15-2-1/

This user comment explains the issue well: 
“I found when I was selecting the radio buttons if the focus is on a radio button, I am then able to use the cursor keys to move up or down to the next radio button. However, if I then try to use the cursor keys again to move up or down, it doesn’t work, and I must use the ‘tab’ key to get the focus back to where I want it.”

image

Effectively this is an issue of focus and control, we need to ensure that focus is taken back to the expected control so that users do not become disorientated.

brockfanning:
Q | A
--- | ---
Fixed issues | Fixes #1757 
Related version | 2.1.0-dev
Bugfix, feature or docs? |
* [ ] Includes tests
* [ ] Updated docs
* [ ] Updated config forms
* [ ] Added CHANGELOG entry

To recreate the problem:
1. Go to https://onsdigital.github.io/sdg-indicators/15-2-1/
2. Press the TAB key until "Woodland area net change" has focus
3. Press the down arrow to switch to "Above-ground biomass stock"
4. Press the down arrow again
5. Note that it does not switch to the next series, and instead just scrolls the page a bit

If you make a feature branch with this PR, after step 4 it should switch to the next series ("Percent of woodland within protected areas").

### DIFF
--- a/_includes/assets/js/view/seriesHelpers.js
+++ b/_includes/assets/js/view/seriesHelpers.js
@@ -3,7 +3,10 @@
  * @return null
  */
 function initialiseSerieses(args) {
-    var templateElement = $('#series_template');
+    var activeSeriesInput = $('#serieses').find(document.activeElement),
+        seriesWasFocused = (activeSeriesInput.length > 0) ? true : false,
+        focusedValue = (seriesWasFocused) ? $(activeSeriesInput).val() : null,
+        templateElement = $('#series_template');
     if (templateElement.length > 0) {
         var template = _.template(templateElement.html()),
             serieses = args.serieses || [],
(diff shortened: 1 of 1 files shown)


sarah-beasley (approved):
Thanks Brock, tested and it now goes through the different series instead of scrolling

PR #1767, opened 2022-08-05
Update title of map play/pause button dynamically

### LINKED ISSUE #1755: (A) Image links not available on map links
**High-level (A)**

https://sdgdata.gov.uk/3-3-3/
image

A text alternative that serves the equivalent purpose is not available for images. This means there is nothing to clearly describe the contents of the images to screen reader users. The ‘play’ link changes to a ‘pause’ link, and the icon acting as a visible label changes from a ‘play’ icon to a ‘pause’ icon to reflect this. However, the only text alternative to this icon is the ‘title’ of the link, but this does not change from ‘play’ to reflect the updated purpose of the control. Here we can either change this to show 'pause' if possible or potentially change to 'play/pause'

image

We need to ensure links have an accessible name which aligns with the visible label onscreen and describes the link purpose. Where icons are used to act as part of the label, ensure that the information conveyed via the icon is also programmatically determinable. This can be achieved by including additional text offering the same information as the icons. This can be visually hidden using CSS.

brockfanning:
Q | A
--- | ---
Feature branch/test site URL | [Link](http://uk-sdg-feature-branches.s3-website.eu-west-2.amazonaws.com/feature-site-map-slider-aria-fix/3-3-3/)
Fixed issues | Fixes #1755
Related version | 2.1.0-dev
* [ ] Includes tests
* [ ] Updated docs
* [ ] Updated config forms
* [ ] Added CHANGELOG entry

Testing notes: In order to test this, you can either inspect the map's play-button element in the browser's developer console, and confirm that its title attribute changes to "Pause" while it is playing. Or I believe you could also use a screenreader to test it, but I'm not sure exactly how it should behave.

### DIFF
--- a/_includes/assets/js/plugins/leaflet.yearSlider.js
+++ b/_includes/assets/js/plugins/leaflet.yearSlider.js
@@ -126,6 +126,12 @@
     });
     // Create the player.
     options.player = new L.TimeDimension.Player(options.playerOptions, options.timeDimension);
+    options.player.on('play', function() {
+      $('.timecontrol-play').attr('title', 'Pause');
+    });
+    options.player.on('stop', function() {
+      $('.timecontrol-play').attr('title', 'Play');
+    });
     // Listen for time changes.
     if (typeof options.yearChangeCallback === 'function') {
       options.timeDimension.on('timeload', options.yearChangeCallback);


sarah-beasley (approved):
Thanks Brock, have checked and the pause title now comes up when it's playing.

PR #1777, opened 2022-08-14
Fix for metadata content overflow on high magnification

### LINKED ISSUE #1761: (AA) Reflow magnification issue
When the content on a page is magnified to 400% the content begins to overlap in a few instances.

1. This is occurring on the archived indicators page currently. This implies this is an issue only occurring with the standalone-indicators layout. 
https://sdgdata.gov.uk/archived-indicators/
image
image

2. Secondly, this occurs on the indicator page when at 400% zoom. This can be seen here in the sources and metadata tabs where the text is overflowing the boxes:
https://sdgdata.gov.uk/15-2-1/
image
image
image
image

3. The final instance is on the user personas page 
image
https://sdgdata.gov.uk/user-personas/

We need the page to be viewable as a single column at 320px without loss of content or functionality. 
"Please note that for content which requires a two-dimensional layout for usage or meaning including tabular information, horizontal scrolling may be necessary in order for the presentation and structure to be preserved."

brockfanning:
This is meant to fix the second issue mentioned in #1761. It only has an effect on 500 pixels width and narrower, so this should only affect phones, and browsers on high zoom. To confirm I would recommend testing a variety of indicator metadata on:
1. Normal desktop
2. Desktop zoomed to 400%
3. Phones

Things to look out for include:
* Metadata that is overflowing outside of the column
* Metadata that seems too narrow to read

Q | A
--- | ---
Feature branch/test site URL | [Link](http://uk-sdg-feature-branches.s3-website.eu-west-2.amazonaws.com/feature-site-zoom-metadata-overflow-fix)
Fixed issues | Fixes #1761 
Related version | 2.1.0-dev
Bugfix, feature or docs? |
* [ ] Includes tests
* [ ] Updated docs
* [ ] Updated config forms
* [ ] Added CHANGELOG entry

### DIFF
--- a/_includes/components/indicator/metadata.html
+++ b/_includes/components/indicator/metadata.html
@@ -1,4 +1,5 @@
 <table class="table metadata-content">
+  <tbody>
   {% for indicator_metadata in site.data.schema %}
     {%- assign metadata_label = indicator_metadata.name | translate_metadata_field -%}
     {% if indicator_metadata.field.scope == include.scope %}
@@ -39,4 +40,5 @@
         {% endunless %}
       {% endif %}
   {% endfor %}
+  </tbody>
 </table>
--- a/_includes/components/indicator/metadata.html
+++ b/_includes/components/indicator/metadata.html
@@ -1,4 +1,4 @@
-<table class="table">
+<table class="table metadata-content">
   {% for indicator_metadata in site.data.schema %}
     {%- assign metadata_label = indicator_metadata.name | translate_metadata_field -%}
     {% if indicator_metadata.field.scope == include.scope %}
--- a/_sass/components/_metadata.scss
+++ b/_sass/components/_metadata.scss
@@ -6,7 +6,7 @@
             max-width: 1px;
         }
         > th {
-          width: 33%;  
+          width: 40%;  
         }
     }
 }
--- a/_sass/components/_metadata.scss
+++ b/_sass/components/_metadata.scss
@@ -5,5 +5,8 @@
             word-wrap: break-word;
             max-width: 1px;
         }
+        th {
+          width: 33%;  
+        }
     }
 }
--- a/_sass/components/_metadata.scss
+++ b/_sass/components/_metadata.scss
@@ -1,11 +1,11 @@
-.metadata-content {
+.metadata-content > tbody > tr {
     @media only screen and (max-width: 500px) {
         table-layout: fixed;
-        th, td {
+        > th, > td {
             word-wrap: break-word;
             max-width: 1px;
         }
-        th {
+        > th {
           width: 33%;  
         }
     }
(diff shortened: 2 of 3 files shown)


otis-bath:
@sarah-beasley this issue should be explained clearly enough in the ticket linked to this PR but if you have any other questions just let me know

sarah-beasley:
@brockfanning 
Hi Brock, 
I just tested some using your feature branch on desktop and found that indicator 8.10.1 now has an issue when zoomed at 400% when it didn't actually before - [LINK](http://uk-sdg-feature-branches.s3-website.eu-west-2.amazonaws.com/feature-site-zoom-metadata-overflow-fix/8-10-1/) 
The width of the metadata column headings is still quite wide for 400% too, 300% is much narrower which allows the main bulk of text to be more readable. Can the columns be adjusted more at 400% to allow for smaller headings but more bulk text?
Thanks,

Sarah

brockfanning:
@sarah-beasley If I understand right, you're pointing out that the left column is too wide and the right column is too narrow, at this high zoom? If so, I believe we can hardcode the ratio. Like for example, we can set the left column to 33% and the right column to 66%. Or something like that.

sarah-beasley:
> @sarah-beasley If I understand right, you're pointing out that the left column is too wide and the right column is too narrow, at this high zoom? If so, I believe we can hardcode the ratio. Like for example, we can set the left column to 33% and the right column to 66%. Or something like that.

Hi @brockfanning yes that's correct, I think the right column is a bit too narrow at the moment which makes it more difficult to read at 400% zoom.

brockfanning:
@sarah-beasley This is ready to check again, if you could refresh the feature branch.

brockfanning:
@sarah-beasley This is ready for another refresh.

My latest commit should help with the 8.10.1 issue - thanks for catching that.

I've also increased the left column to 40%. Do you think the right column is OK at 60%?

brockfanning:
@sarah-beasley My 2 cents on it - I don't think there can be a perfect solution, so it is a question between two evils - is it worse to have a cramped right column, or is it worse to have more mid-word wrapping in the left column? So we're just looking for the best balance - maybe that's 40/60, 33/66, 30/70, etc.

sarah-beasley:
thanks @brockfanning the fix worked for the 8.10.1, that's much better. I think you're right, it really depends what device you use too. on my laptop alone, the 40:60 is still not right for the left hand column, but on my usual monitor which is wider screen, it is fine. @otis-bath what do you think? should we keep at 40:60 (aka the latest feature branch (in the link))

otis-bath:
@sarah-beasley @brockfanning I think this is fine now, as Brock says we are never going to be able to get this absolutely perfect in all cases however it is a definite improvement and now users will only experience slight issues in rare use cases

sarah-beasley (approved):
Approving based on the 40:60 ratio of column widths and the 8.10.1 issue being fixed. thanks

Issue #1792, opened 2022-08-26
Possible problem with understanding the maps


MoBosse:
**Please complete the following information** 
Site repository URL: [https://github.com/DnsUpgradeEnvironment/dns-indicators](https://github.com/DnsUpgradeEnvironment/dns-indicators)
Open SDG version: 2.1.0-dev

**Describe the bug**
[Here](https://dnsupgradeenvironment.github.io/dns-indicators/en/3-1-cd/) we have an two time series ("adolescent smokers" and "adult smokers") we have two time series ("adolescent smokers" and "adult smokers") to be displayed on the same indicator page. To make them easier to compare, the two series are not named as "series" so that they can both be selected at the same time. Therefore, the chart/map title is not dynamic and applies to both series.
The problem is that only one of the series has geodata and is displayed in the map. The title is still "Proportion of adolescent and adult smokers", and since there is no other data to select in the map, only the unit is shown in the legend. The result is that the user has no way of telling whether the data shown in the map is for adolescents or adults.
![grafik](https://user-images.githubusercontent.com/50657629/186851547-5ae441fd-11ea-46a8-be86-002a61fc1fe0.png)

**To Reproduce**
1. Go to [https://dnsupgradeenvironment.github.io/dns-indicators/en/3-1-cd/](https://dnsupgradeenvironment.github.io/dns-indicators/en/3-1-cd/)
2. Click on map
3. See error

**Expected behavior**
I could think of two possible solutions:

1. There could be a separate title for maps that (if defined) overrides the chart title.
2. All disaggregation columns that are filled for the map datasets must be displayed in the map legend (even if the disaggregation cannot be changed)

brockfanning:
@MoBosse Both of those solutions seem reasonable - and in particular I think there is a clear need for the second one - displaying disaggregation columns even if they can't be changed. That's important information that should be displayed.

I believe the problem is here: https://github.com/open-sdg/open-sdg/blob/2.1.0-dev/_includes/assets/js/plugins/leaflet.disaggregationControls.js#L49-L79

If you replace all that code with `return disaggregations;` I think you'll see your "time series" column. But the problem is that you'll also see your "federal state" column. So the challenge, I think, is to figure out how we can hide the "federal state" column but still show the "time series" column.

I wonder if there is an efficient way for Open SDG to identify that a column corresponds to regions on the map. If so, we could just hide that column ("federal state" in this case) and display all the others.

brockfanning:
[...]

PR #1795, opened 2022-08-26
Better pruning of map disaggregations to hide region columns

brockfanning:
Testing instructions:

For any indicator with a map, add a new column in the data, and give each row the same value. For example add a new column called "Foo" and give each row a value of "Bar" in that column.

Without this PR, that new disaggregation value ("Foo: Bar") will *not* appear in the lower-left map disaggregation box. But with this PR, it should appear.

If you would like to test the button for changing the map disaggregations, the site configuration for that is "disaggregation_controls" under "map_options", as shown here: https://open-sdg.readthedocs.io/en/latest/maps/#map_options (it needs to be "true" for the button to appear)

Q | A
--- | ---
Feature branch/test site URL | [Link](http://uk-sdg-feature-branches.s3-website.eu-west-2.amazonaws.com/feature-map-identify-region-columns/3-3-3/)
Related version | 1.6.0-dev
Bugfix, feature or docs? |
* [ ] Includes tests
* [ ] Updated docs
* [ ] Updated config forms
* [ ] Added CHANGELOG entry

### DIFF
--- a/_includes/assets/js/plugins/leaflet.disaggregationControls.js
+++ b/_includes/assets/js/plugins/leaflet.disaggregationControls.js
@@ -46,9 +46,11 @@
         getVisibleDisaggregations: function() {
             var features = this.plugin.getVisibleLayers().toGeoJSON().features;
             var disaggregations = features[0].properties.disaggregations;
-            // The purpose of the rest of this function is to
-            // "prune" the disaggregations by removing any keys
-            // that are identical across all disaggregations.
+            // The purpose of the rest of this function is to identiy
+            // and remove any "region columns" - ie, any columns that
+            // correspond exactly to names of map regions. These columns
+            // are useful on charts and tables but should not display
+            // on maps.
             var allKeys = Object.keys(disaggregations[0]);
             var relevantKeys = {};
             var rememberedValues = {};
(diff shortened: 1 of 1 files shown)


brockfanning:
@MoBosse Is this PR working to fix the issue from #1792 ?

MoBosse:
@brockfanning Yes, this fixes the issue.

brockfanning:
@otis-bath @sarah-beasley I added some testing instructions here.

sarah-beasley (approved):
Thanks Brock, I managed to add the Foo Bar, add the disaggregation box to the map, see that Foo Bar wasn't there, add your PR in and then Foo Bar appeared in box!

sarah-beasley:
one thing i noticed in this feature branch was that on the home screen, the download data text looks different to normal. I wasn't sure this was correct? 
image

brockfanning:
@sarah-beasley I think you may see that on all feature-branches, until 2.1.0 is released. The reason is that the latest 2.1.0-dev code of open-sdg relies on the latest 2.1.0-dev code of sdg-translations. So I think it's OK to ignore that issue, as it will resolve itself when it is finally time to test the UK platform on 2.1.0.

PR #1797, opened 2022-08-31
Double focus outline (black in addition to yellow)

### LINKED ISSUE #1760: (AA) Non-text colour contrast requirements not met
**Medium level (AA)**

https://sdgdata.gov.uk/
Across the site, on the homepage, goals page, indicator page etc. when an object is selected a yellow border appears around the outside. The yellow (#ffdd00) border used to indicate when the links are focused fails colour contrast guidelines with a ratio of 1.3:1 against the white background.
image
image
image

We need this ratio to be at least 3:1

brockfanning:
This is an experimental pull-request that has this change:
1. Changes the focus outline of some elements from a single yellow box to a double black/yellow box. These include the goal tiles, the search box, the indicator chart, and possible other elements too (so testing is needed)

Q | A
--- | ---
Feature branch/test site URL | [Link](http://uk-sdg-feature-branches.s3-website.eu-west-2.amazonaws.com/feature-site-focus-background-black-and-yellow/)
Fixed issues | Fixes #1760 
Related version | 2.1.0-dev
Bugfix, feature or docs? |
* [ ] Includes tests
* [ ] Updated docs
* [ ] Updated config forms
* [ ] Added CHANGELOG entry

### DIFF
--- a/_sass/base/_buttons.scss
+++ b/_sass/base/_buttons.scss
@@ -1,4 +1,4 @@
-button:focus-visible {
+button:focus {
     @include focus;
 }
 .btn {
@@ -6,7 +6,7 @@ button:focus-visible {
     &:visited {
         color: $button-foregroundColor-visited;
     }
-    &:focus-visible {
+    &:focus {
         color: $button-foregroundColor-focus;
     }
 }
@@ -27,13 +27,13 @@ button:focus-visible {
     &:visited {
         color: $button-foregroundColor-visited !important;
     }
-    &:focus-visible {
+    &:focus {
         background: $button-color-focus !important;
         color: $button-foregroundColor-focus !important;
     }
 }
 body.contrast-high {
-    button:focus-visible {
+    button:focus {
         @include focusHighContrast;
     }
     .btn {
@@ -41,7 +41,7 @@ body.contrast-high {
         &:visited {
             color: $color-dark-highContrast !important;
         }
-        &:focus-visible {
+        &:focus {
             color: $color-dark-highContrast !important;
         }
     }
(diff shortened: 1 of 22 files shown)


otis-bath:
@brockfanning @phillipgardiner - So have this on the FB, know Phill had something to add on regarding GDS standards so I am not going to approve just yet. However the yellow and black border for the search function and when you click on the graph look good.

phillipgardiner:
@otis-bath @brockfanning 

The change to the focus outline of some elements from a single yellow box to a double black/yellow box is great and to me has the desired effect. 

However, looking at the GDS guidelines (which is what we've tend to use) elements are supposed to continue to receive focus weather it via keyboard or an mouse. I therefore think we should not be changing all our focus rules to focus-visible rules. You can see some examples here: https://design-system.service.gov.uk/components/select/ and https://design-system.service.gov.uk/components/checkboxes/

If we follow these guidelines it actually means that elements such as the dissagregation drops downs and the check boxes within them should also be using the double black/yellow box combination too (whereas currently they are just yellow).

brockfanning:
@phillipgardiner Thanks. I think the disaggregation dropdowns are getting the double box, but not the checkboxes. I'll try to add that.

To be sure about the focus-visible - I don't see any mention of it in the GDS guidelines. It is true that they are not _using_ focus-visible, but that is not surprising, because focus-visible only recently got full browser support, within the past year. So it's possible that focus-visible does not violate GDS guidelines, and instead they just haven't implemented it yet. Anyway I'm happy to split that out into a separate PR.

brockfanning:
@otis-bath This is ready for a refresh - I've added the double-focus on checkboxes, and I've split out the focus-visible stuff into a different PR (#1800).

phillipgardiner:
@otis-bath @brockfanning Yep looks good to me. Happy to discuss PR (https://github.com/open-sdg/open-sdg/pull/1800). on the tech call, just in case I am not understanding it correctly :)

PR #1800, opened 2022-09-03
Use focus-visible instead of focus

brockfanning:
This was split off from #1797 for separate discussion/testing.

Q | A
--- | ---
Feature branch/test site URL | [Link](http://uk-sdg-feature-branches.s3-website.eu-west-2.amazonaws.com/feature-site-focus-visible/)
Related version | 2.1.0-dev
Bugfix, feature or docs? |
* [ ] Includes tests
* [ ] Updated docs
* [ ] Updated config forms
* [ ] Added CHANGELOG entry

### DIFF
--- a/_sass/base/_buttons.scss
+++ b/_sass/base/_buttons.scss
@@ -1,4 +1,4 @@
-button:focus {
+button:focus-visible {
     @include focus;
 }
 .btn {
@@ -6,7 +6,7 @@ button:focus {
     &:visited {
         color: $button-foregroundColor-visited;
     }
-    &:focus {
+    &:focus-visible {
         color: $button-foregroundColor-focus;
     }
 }
@@ -27,13 +27,13 @@ button:focus {
     &:visited {
         color: $button-foregroundColor-visited !important;
     }
-    &:focus {
+    &:focus-visible {
         background: $button-color-focus !important;
         color: $button-foregroundColor-focus !important;
     }
 }
 body.contrast-high {
-    button:focus {
+    button:focus-visible {
         @include focusHighContrast;
     }
     .btn {
@@ -41,7 +41,7 @@ body.contrast-high {
         &:visited {
             color: $color-dark-highContrast !important;
         }
-        &:focus {
+        &:focus-visible {
             color: $color-dark-highContrast !important;
         }
     }
@@ -55,7 +55,7 @@ body.contrast-high {
         &:hover {
             background: $button-color-hover-highContrast !important;
         }
-        &:focus {
+        &:focus-visible {
             background: $color-highlight-highContrast !important;
         }
     }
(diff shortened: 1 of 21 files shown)


brockfanning:
@otis-bath This is not slated for 2.1.0 but do you mind creating a feature branch for this, so we can discuss?

otis-bath:
Hi @brockfanning - does this require any changes other than the remote_theme? Have this on a FB and as you can see the disaggregations don't look right http://uk-sdg-feature-branches.s3-website.eu-west-2.amazonaws.com/feature-site-focus-visible/1-2-1/

otis-bath:
We need to know if this is the GDS standard or not. Will follow up with GDS and do a bit of digging to see what standard we want to employ.

phillipgardiner:
@brockfanning @otis-bath @LucyGwilliamAdmin 

I received a response from GDS which I have pasted below.. I also raised it on the GDS design channel in slack and the feedback  there was:

- it would introduce an inconsistency of focus state according to the mode of interaction - so keep consistent
- there is nothing within the Web Content Accessibility Guidelines stopping you from having a different indicator to keyboard focus, but it 'might' confuse users (but no user testing has been done to date to find out)
- It may confuse users who happen to be using a mix of mouse and keyboard when navigating inputs

-----------------------------------------

Hi Phillip,
 
I don’t think I can give a definitive answer on whether focus states should be the same across everything. This isn’t something we’ve considered yet in the GOV.UK Design System team because we still support older browsers (for example, Internet Explorer) which do not support focus-visible. It’s likely that attempting to implement this would make our code more complex, and we’re not aware of any issues being caused by having the same focus states.
 
In order to remove the focus state for users using a mouse/pointing device, I think we’d need some clear evidence that things would be made better by removing them.
 
I hope this helps. If you do any kind of user testing on this, I think we’d be interested to see your findings!
 
Thanks,
 
Vanita
GOV.UK Design System team

-----------------------------------------------

I think from my perspective, I would say, keep how it is for now, primarily for three reasons:
1) To stay inline with the (current) GDS standard 
2) I'm not aware of any issues being caused by having the same focus states for both keyboard and mouse users (but happy to be corrected)
3) We don't have any clear evidence that things would be made better by removing it for mouse users (again happy to be corrected).

Thanks

brockfanning:
@phillipgardiner Ok great, thanks for investigating. My hunch is that mouse-users would prefer to not have the focus trigger when they click on things, but I don't have any evidence for that. I'll close this PR and we can revisit in the future if the user testing leads us back to this.

phillipgardiner:
@brockfanning Thanks Brock and I'm not disagreeing with your hunch. But like GDS have mentioned until such a time we have some concrete evidence on which we can base the change from their norm, I think we have to keep to their standard. Otherwise in things like our UK assessments we could fall foul of not abiding to the standards. Obviously it won't be me but can be revisited in the future as sounds like we might be leading the way with this hypothesis and can help inform GDS's approach.

PR #1808, opened 2022-09-09
Year-slider starts with the most recent year

christoph-LSN:
This is my first contribution to open-sdg

I hope I do everything right...

Starting the map wit the most recent year (as @MoBosse implementet for his SDG 1.8.0)

Example Indicator: [https://christoph-lsn.github.io/MT_Site/2-1-1/](https://christoph-lsn.github.io/MT_Site/2-1-1/)

Q | A
--- | ---
Fixed issues | Fixes #1650 
Related version | 2.1.0-dev
Bugfix, feature or docs? |
* [ ] Includes tests
* [ ] Updated docs
* [ ] Updated config forms
* [ ] Added CHANGELOG entry

### DIFF
--- a/_includes/assets/js/plugins/jquery.sdgMap.js
+++ b/_includes/assets/js/plugins/jquery.sdgMap.js
@@ -470,7 +470,8 @@
         plugin.setColorScale();
 
         plugin.years = _.uniq(availableYears).sort();
-        plugin.currentYear = plugin.years[0];
+        //Start the map with the most recent year
+        plugin.currentYear = plugin.years.slice(-1)[0];
 
         // And we can now update the colors.
         plugin.updateColors();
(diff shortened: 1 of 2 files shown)


brockfanning:
@christoph-LSN This _should_ be all that is needed, but for me the "handle" of the year slider still appears on the left, even though the the right-most year is displayed. It would be nice if we could figure out how to make the handle appear on the right. It seems like maybe a bug in the third-party library we're using there, but I wonder if there is some way to work around it.

LucyGwilliamAdmin:
@christoph-LSN  @christoph-LSN it seems that the issue brock has identified is only on initial load i.e. if you click either the left or right arrow, it goes to the correct position - not sure if this is any help in figuring out the solution?

brockfanning:
I guess that means it is just a visual issue - I'll merge this and then work on that. Thanks @christoph-LSN!

brockfanning (approved):

PR #1813, opened 2022-09-21
Use raw value for chart tooltip

### LINKED ISSUE #1796: alterDataDisplay gets confused with values > 1000
Site repository URL: [e.g. github.com/ONSdigital/sdg-data](https://github.com/DnsUpgradeEnvironment/dns-indicators)
Open SDG version: 2.1.0-dev

**Describe the bug**
In the chart tooltip, the displayed values are messed up for values bigger than 1000 (see screenshot) when a decimal separator is defined as ",". My guess is that the tooltip gets the data (say 2000), wants to display it as "2.000" with a dot as the thousands separator, then [this line](https://github.com/open-sdg/open-sdg/blob/2.1.0-dev/_includes/assets/js/view/dataHelpers.js#L32) in the alterDataDisplay function replaces the dot with a comma, thinking that the dot is a decimal separator.
Then (although I don't know why the function does this after replacing the dot) the value (now "2,000") is rounded to the defined number of decimal places (0 in this case) to "2" since the function interpretes the comma as decimal seperator.

**To Reproduce**
1. Go to [https://dnsupgradeenvironment.github.io/dns-indicators/12-2/](https://dnsupgradeenvironment.github.io/dns-indicators/12-2/)
2. Hover over any datapoints
3. See tooltip

**Screenshots**
To avoid misunderstandings: The number that should be shown here in the tooltip is one thousand nine hundred and eighty-five.
grafik

**Additional context**
Note that on the screenshot and on the linked indicator page the "alterDataDisplay" function is disabled for the data on the y-axis. Therefore, the dots (which correctly display the thousands) are displayed.

brockfanning:
Q | A
--- | ---
Feature branch/test site URL | [Link](http://uk-sdg-feature-branches.s3-website.eu-west-2.amazonaws.com/feature-site-chart-tooltip-use-raw/16-1-1/)
Fixed issues | Fixes #1796 
Related version | 2.1.0-dev
Bugfix, feature or docs? |
* [ ] Includes tests
* [ ] Updated docs
* [ ] Updated config forms
* [ ] Added CHANGELOG entry

### DIFF
--- a/_includes/assets/js/plugins/jquery.sdgMap.js
+++ b/_includes/assets/js/plugins/jquery.sdgMap.js
@@ -266,11 +266,21 @@
       opensdg.dataDisplayAlterations.forEach(function(callback) {
         value = callback(value);
       });
-      if (this._precision || this._precision === 0) {
-        value = Number.parseFloat(value).toFixed(this._precision);
+      if (typeof value !== 'number') {
+        if (this._precision || this._precision === 0) {
+          value = Number.parseFloat(value).toFixed(this._precision);
+        }
+        if (this._decimalSeparator) {
+          value = value.toString().replace('.', this._decimalSeparator);
+        }
       }
-      if (this._decimalSeparator) {
-        value = value.toString().replace('.', this._decimalSeparator);
+      else {
+        var localeOpts = {};
+        if (this._precision || this._precision === 0) {
+            localeOpts.minimumFractionDigits = this._precision;
+            localeOpts.maximumFractionDigits = this._precision;
+        }
+        value = value.toLocaleString(opensdg.language, localeOpts);
       }
       return value;
     },
--- a/_includes/assets/js/view/chartTypeBase.js
+++ b/_includes/assets/js/view/chartTypeBase.js
@@ -76,7 +76,7 @@ opensdg.chartTypes.base = function(info) {
                     backgroundColor: 'rgba(0,0,0,0.7)',
                     callbacks: {
                         label: function (tooltipItem) {
-                            return translations.t(tooltipItem.dataset.label) + ': ' + alterDataDisplay(tooltipItem.formattedValue, tooltipItem.dataset, 'chart tooltip');
+                            return translations.t(tooltipItem.dataset.label) + ': ' + alterDataDisplay(tooltipItem.raw, tooltipItem.dataset, 'chart tooltip');
                         },
                         afterBody: function () {
                             var unit = MODEL.selectedUnit ? translations.t(MODEL.selectedUnit) : MODEL.measurementUnit;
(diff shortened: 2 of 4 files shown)


MoBosse:
Works and fixes the problems described in issue #1796 
Thanks @brockfanning

brockfanning:
Thanks for testing @MoBosse!

@sarah-beasley As discussed if you can test for the UK platform - the thing to look out for are any regressions in the chart "tooltips" when hovering over data points. Ideally nothing will change for you -- decimal places, rounding, etc, should all be the same as the live site.

sarah-beasley (approved):
Thanks Brock, have tested on the UK site and there are no changes to the chart tooltips when you hover over the data points. have compared directly with the live site.

LucyGwilliamAdmin:
@sarah-beasley I think the indicator you've linked to uses decimal, but might also be worth checking an indicator that reports thousands to see if the comma is still showing as expected if not already done so?

sarah-beasley:
> @sarah-beasley I think the indicator you've linked to uses decimal, but might also be worth checking an indicator that reports thousands to see if the comma is still showing as expected if not already done so?

Thanks for pointing that out, just checked and the comma has been removed in the thousands as shown here - http://uk-sdg-feature-branches.s3-website.eu-west-2.amazonaws.com/feature-site-chart-tooltip-use-raw/6-a-1/

MoBosse:
@brockfanning If you consider to solve the missing comma issue by adding a configural thousands seperator: I found [this aproach](https://github.com/DnsUpgradeEnvironment/dns-indicators/blob/develop/_includes/assets/js/view/dataHelpers.js#L52-L54) and implemented it [here ](https://dnsupgradeenvironment.github.io/dns-indicators/16-1/) (at the moment with a white space as separator). It works pretty good as far as I can see.

brockfanning:
@sarah-beasley Can you try refreshing the feature branch when able?

@MoBosse I've been reading up on [toLocaleString](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Number/toLocaleString) and it seems like it might be a replacement for our `decimal_separator` option and also solve the thousands problem at the same time. Can you try out this PR? I've just added some code to try to use toLocaleString.

MoBosse:
Tested with several combinations of input-values with and without decimal places, values less and greater than 1000, and precision-decimal defined and not-defined and languages English and German. Worked perfectly on: 
- values on chart y axis
- values in tooltip
- values in tables
- target value in table

brockfanning:
@MoBosse That's great news. I'll add an update to the docs in this PR to mention that `decimal_separator` should no longer be needed in most cases.

sarah-beasley (approved):
Thanks for making the change Brock, the commas and decimals are showing perfectly now!

PR #1839, opened 2022-10-20
Use more fuzzy searching for maps

brockfanning:
Q | A
--- | ---
Feature branch/test site URL | [Link](http://uk-sdg-feature-branches.s3-website.eu-west-2.amazonaws.com/feature-site-map-fuzzy-fix/3-3-3/)
Related version | 2.1.0-dev
Bugfix, feature or docs? |
* [ ] Includes tests
* [ ] Updated docs
* [ ] Updated config forms
* [ ] Added CHANGELOG entry

### DIFF
--- a/_includes/assets/js/plugins/leaflet.searchAccessible.js
+++ b/_includes/assets/js/plugins/leaflet.searchAccessible.js
@@ -109,8 +109,9 @@
         // https://github.com/stefanocudini/leaflet-search/issues/264
         var firstSuggestion = this._tooltip.childNodes[0].innerText;
         var firstSuggestionLower = firstSuggestion.toLowerCase();
-        var userInputLower = this._input.value.toLowerCase();
-        if (firstSuggestionLower !== userInputLower && firstSuggestionLower.includes(userInputLower)) {
+        var userInput = this._input.value;
+        var userInputLower = userInput.toLowerCase();
+        if (firstSuggestion !== userInput && firstSuggestionLower.includes(userInputLower)) {
           this._input.value = firstSuggestion;
         }
       }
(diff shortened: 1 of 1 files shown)


otis-bath:
@brockfanning - just tested this, seems like its not working on the UK platform using the search button. Was this just to make it case insensitive as currently cannot get this working?

brockfanning:
@otis-bath No, I must have missed something, I'll check again.

sarah-beasley:
> @brockfanning - just tested this, seems like its not working on the UK platform using the search button. Was this just to make it case insensitive as currently cannot get this working?
> 
> Did this necessitate any change other than the version of Open SDG?

I've just tried and it did work for me on 3-3-3, just typed in 'lon' and then pressed search and it came up!

brockfanning:
@otis-bath @sarah-beasley I think that it was not working with full matches (eg "london") but I've pushed a fix that should hopefully fix that, if you could refresh the FB.

otis-bath (approved):
That's sorted it thanks @brockfanning

sarah-beasley:
Have given a refresh and think it all works great now, thanks @brockfanning! Just wondering whether we need to just double check the hover label still - should it be map search now that it works? we had some feedback from the DAC report too which suggested we remove the 'search' label given that it didn't work either, however this is the comment that was left:

'When typing northeast into the map search, JAWS announced that no options were available to me. I attempted to delete northeast given that I was advised that no results were present. In doing so a member of the support team advised me that the whole section of the page was highlighted meaning that I was unsure of my place on the page A member of the support team advised that when attempting to search again even though JAWS announces content from the map search, my focus was in the site search.'

Given this and the new fix that makes the search button better, do you think we would be better off specifying it to 'map search' instead?

brockfanning:
@sarah-beasley I should already be "Map search". I believe it is showing "Search" on the feature branch because that is not using the latest translations.

sarah-beasley:
ah that's great @brockfanning, thanks for confirming.

PR #1915, opened 2023-01-05
Oct 22 Docs revamp

otis-bath:
Plan on merging these changes proposed last year before doing the new sections looked at by James and myself, don't want to cause complications with v2.2.0 that's all - is this a better approach @brockfanning or should these be merged to 2.2.0-dev first?


docs/open-sdg-features.md
```diff
@@ -30,6 +43,18 @@ It is possible to "boost" one or more fields in the search index, so that more r
 
 It is also possible to index additional metadata fields, so that more fields are searched e.g. if you wanted the search to return pages where the graph title contains the search term. See the [search_index_extra_fields configuration option](https://open-sdg.readthedocs.io/en/latest/configuration/#search_index_extra_fields) for how to configure this.
 
+## Accessibility High Contrast version
+As well as the default contrast version, Open SDG also offers a high contrast version. By default two menu buttons show, to allow users to choose between the different contrast levels, for example on the [US site](https://sdg.data.gov/). Another approach is to use a contrast toggle button, for example on the [UK site](https://sustainabledevelopment-uk.github.io/).
+
+For guidance on how to use the more accessible contrast button, see the Configuration page [contrast_type](configuration.md) section.
+
+## Multilingual
+By default, the site is created in English but this can be changed and multiple languages can be added.
```
brockfanning, line 52: Technically there is not a "default" language. The site-starter does start off with English (and Spanish too I think) but that is just the way that the site-starter is configured. I might recommend this language here: `The Open SDG platform can be either monolingual or multilingual. All aspects of the platform are included in the translation system, including page content, indicator metadata, indicator data disaggregations, menus and interface elements, etc.`

docs/open-sdg-features.md
```diff
@@ -75,37 +100,12 @@ An example of using the **Not applicable** option is [Rwanda's Reporting status
...
-
-An example of menu items being added is [UK's site](https://sustainabledevelopment-uk.github.io/)
-
-Even though only four pages are linked to in the menu by default, there are other pages provided for use (e.g. Contacts, News) as well as the option to easily create your own pages.
-
-For guidance on how to add more pages to the menu, see the Configuration page [menu](configuration.md#menu) section.
-
-## Filter by disaggregation
-Open SDG platforms allow data to be displayed in a way in which it can be filtered by disaggregation. This allows user to compare different breakdowns for a particular indicator.
+## Targets on goal pages
```
brockfanning, line 103: This section is quite out of date - the goal now automatically shows targets (eg the UK page does show targets). As such I'm not sure it is needed to have a "Targets on goal pages" section here. Instead maybe we should just have a "Goal pages" section that talks about the goal pages.

We'll also need to remove this whole section: https://open-sdg.readthedocs.io/en/latest/customisation/#optional-feature-goal-page-layouts since it is out of date.

docs/open-sdg-features.md
```diff
@@ -30,6 +43,18 @@ It is possible to "boost" one or more fields in the search index, so that more r
 
 It is also possible to index additional metadata fields, so that more fields are searched e.g. if you wanted the search to return pages where the graph title contains the search term. See the [search_index_extra_fields configuration option](https://open-sdg.readthedocs.io/en/latest/configuration/#search_index_extra_fields) for how to configure this.
 
+## Accessibility High Contrast version
+As well as the default contrast version, Open SDG also offers a high contrast version. By default two menu buttons show, to allow users to choose between the different contrast levels, for example on the [US site](https://sdg.data.gov/). Another approach is to use a contrast toggle button, for example on the [UK site](https://sustainabledevelopment-uk.github.io/).
```
brockfanning, line 47: This is out of date also. There is no longer a "contrast_type" setting, and the only option is the toggle button. I do think it's worth describing here, but there doesn't need to be links to any docs.

brockfanning (changes requested):
@otis-bath A few specific comments below, but also a general one: It might be a little dangerous to link to real-world platforms as examples of the features, because it's possible for those platforms to change. I think screenshots are safer. However we don't want to commit a bunch of images to the repository. I wonder if it would be possible for this page to live on open-sdg.org, where we can add as many images as we'd like. Maybe something to discuss in future calls.

PR #1930, opened 2023-01-24
Year range per disaggregation

### LINKED ISSUE #1890: Fix year slider for map disaggregations
Germany have [a case on their platform](https://dns-indikatoren.de/en/8-5-ab/) where they have a series with two disaggregations (20-64 years and 60-64 years). For 20-64 years they have data up to 2021, for 60-64 years the series currently ends in 2019, but the time slider runs until 2021 no matter which disaggregation is chosen. Since the map initially shows the 60-64 years disaggregation and the most recent year, the user sees a blank map when the map is opened.

When integrating start values for maps, maybe it is possible to add a rule that makes the maps only show those years where there are values for the chosen series/unit/disaggregation - link in with issue #1889

brockfanning:
Q | A
--- | ---
Fixed issues | Fixes #1890 
Related version | 2.2.0-dev
Bugfix, feature or docs? |
* [ ] Includes tests
* [ ] Updated docs
* [ ] Updated config forms
* [ ] Added CHANGELOG entry

### DIFF
--- a/_includes/assets/js/plugins/jquery.sdgMap.js
+++ b/_includes/assets/js/plugins/jquery.sdgMap.js
@@ -328,6 +328,30 @@
       return [opensdg.remoteDataBaseUrl, 'geojson', subfolder, fileName].join('/');
     },
 
+    getYearSlider: function() {
+      var plugin = this,
+          years = plugin.years[plugin.currentDisaggregation];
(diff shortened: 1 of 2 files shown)


otis-bath (approved):
@Govej has had a look at building this and the fix works so happy to approve

PR #2016, opened 2023-06-29
Sticky header and first column on data tables

### LINKED ISSUE #2010: Pin the year column in table
When you add multiple disaggregations to the table, it is easy to get lost in the data as the horizontal scroll is so long. Suggestion to pin the year column so that when you scroll across, you always have the year as a reference.

**To Reproduce**
Add multiple disaggregations to the table and then scroll across - you will see you lose the reference year column

**Screenshots**
image

**Desktop (please complete the following information):**
 - Browser:  chrome
 - Version: [e.g. 22]

brockfanning:
This PR makes the header sticky, and also makes the "Year" column sticky. To test the sticky header, you would need to try an indicator with a lot of years (as you scroll down, you should see that the header is sticky). To test the sticky year column, you would need to try an indicator with a lot of disaggregations.

Q | A
--- | ---
Fixed issues | Fixes #2010 
Related version | 2.3.0-dev
* [ ] Includes tests
* [ ] Updated docs
* [ ] Updated config forms
* [ ] Added CHANGELOG entry

### DIFF
--- a/_includes/assets/js/view/tableHelpers.js
+++ b/_includes/assets/js/view/tableHelpers.js
@@ -189,6 +189,24 @@ function createTable(table, indicatorId, el, isProxy) {
                 var sortDirection = $(this).attr('aria-sort');
                 $(this).find('span[role="button"]').attr('aria-sort', sortDirection);
             });
+
+        let tableWrapper = document.querySelector('.dataTables_wrapper');
+        if (tableWrapper) {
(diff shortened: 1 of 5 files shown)


sarah-beasley (approved):

sarah-beasley:
Thanks Brock, worked really well! :)

PR #2025, opened 2023-07-13
Indicator iframes

### LINKED ISSUE #2023: Embed indicators onto other indicator pages
Allow embedding indicators on other pages (discussions/2013) embedded using an iframe. Should be able to generate pages for every indicator, generate a second page and the header and footer would be hidden - so could be used to embed on other sites. Think about removing the scroll, metadata tabs, etc from the embedded indicator?

brockfanning:
This PR (along with https://github.com/open-sdg/jekyll-open-sdg-plugins/pull/159) create iframe-friendly versions of each indicator. The URLs of the iframe-friendly versions is the same, but with "-frame" at the end. For example the iframe-friendly version of `/1-1-1` would be `/1-1-1-iframe`.

The "iframe-friendly" part is that these new versions of the indicators do not include the sitewide header and footer. They do include other parts of the indicator pages, like the indicator header, and the metadata tabs, which in theory could be removed as well, if that is what folks prefer (let me know in the comments below). They also include usage of the Pym.js library which can be used to avoid vertical scrollbars on the iframe.

Testing instructions:

In your site repository's Gemfile, point to my other branch in jekyll-open-sdg-plugins:

`gem 'jekyll-open-sdg-plugins', git: 'https://github.com/brockfanning/jekyll-open-sdg-plugins.git', branch: 'indicator-iframes'`

In any page in your site, add some code in order to embed an indicator. For example, to create a new page that embeds an indicator, you could create a file called "my-page.md" with something like this:

```
---
title: My page
permalink: /my-page/
layout: page
---
Indicator 1.1.1 should be embedded below:

<script type="text/javascript" src="https://pym.nprapps.org/pym.v1.min.js"></script>
<div id="my-iframe-container"></div>
<script>
    var pymParent = new pym.Parent('my-iframe-container', '{{ site.baseurl }}/1-1-1-iframe', {});
</script>
```

You could also simply do an iframe directly, however then you lose out on the benefit of the Pym.js library, which is what is supposed to prevent vertical scrollbars. If you wanted to do this, it would be something like this:

```
---
title: My page
permalink: /my-page/
layout: page
---
Indicator 1.1.1 should be embedded below:

<iframe src="{{ site.baseurl }}/1-1-1-iframe"></iframe>
```

(though you'd probably also need to play around with the height and width)

Note that the automated tests here will fail until the jekyll-open-sdg-plugins PR is merged.

Q | A
--- | ---
Fixed issues | Fixes #2023 
Related version | 2.3.0-dev
* [ ] Includes tests
* [ ] Updated docs
* [ ] Updated config forms
* [ ] Added CHANGELOG entry

### DIFF
--- a/_includes/components/indicator/metadata-panes.html
+++ b/_includes/components/indicator/metadata-panes.html
@@ -39,7 +39,7 @@
     {% else %}
         {% include components/indicator/metadata-panes-default.html %}
     {% endif %}
-    {% if site.environment == 'staging' %}
+    {% if site.environment == 'staging' and page.layout != 'indicator-iframe' %}
         <div role="tabpanel" class="tab-pane" id="edit">
             {% include components/indicator/edit-buttons.html %}
         </div>
--- a/_includes/components/indicator/metadata-tabs.html
+++ b/_includes/components/indicator/metadata-tabs.html
@@ -26,7 +26,7 @@
     {% else %}
         {% include components/indicator/metadata-tabs-default.html %}
     {% endif %}
-    {% if site.environment == 'staging' %}
+    {% if site.environment == 'staging' and page.layout != 'indicator-iframe' %}
         <li role="presentation" class="nav-item">
             <button class="nav-link" data-bs-toggle="tab" data-bs-target="#edit" aria-controls="edit" role="tab">{{ page.t.indicator.edit }}</a>
         </li>
--- a/_layouts/indicator-iframe.html
+++ b/_layouts/indicator-iframe.html
@@ -1,4 +1,5 @@
 {% include head.html %}
+{% include multilingual-js.html key="header" %}
 
 {% include components/indicator/fields-template.html %}
 {% include components/indicator/units-template.html %}
@@ -14,7 +15,6 @@
   {% include components/indicator/indicator-progress.html %}
   {% include components/indicator/indicator-content.html %}
   {% include components/indicator/indicator-main.html %}
-  {% include back-to-top.html %}
 </div>
 
 {% include scripts.html %}
--- a/_layouts/indicator-iframe.html
+++ b/_layouts/indicator-iframe.html
@@ -1,0 +1,27 @@
+{% include head.html %}
+
+{% include components/indicator/fields-template.html %}
+{% include components/indicator/units-template.html %}
+{% include components/indicator/series-template.html %}
+{% include multilingual-js.html key="indicator" %}
+{% include multilingual-js.html key="data" %}
+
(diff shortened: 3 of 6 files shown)


christoph-LSN:
@brockfanning how do I test this? pointing to brockfanning:indicator-iframes in my site config.yml? And then creating a page including _layouts/indicator-iframe.html?

brockfanning:
@christoph-LSN I've added testing instructions above. Thanks for trying it out!

christoph-LSN:
Thank you @brockfanning I will test it tomorrow morning!

christoph-LSN:
@brockfanning @MoBosse it works, more or less.

loading the pagen https://christoph-lsn.github.io/MT_Site/decision_makers/index.html first time half of the metadata tabs aren't shown. Inspecting the site or changing the size of the browser window the rest is shown.

how do I add more than one indicator on one page?

Adding a second Indicator shows only the last one.

https://github.com/christoph-LSN/MT_Site/blob/develop/_pages/decision_makers.html

````
<script type="text/javascript" src="https://pym.nprapps.org/pym.v1.min.js"></script>
<div id="my-iframe-container"></div>
<script>
    var pymParent = new pym.Parent('my-iframe-container', '{{ site.baseurl }}/1-1-1-iframe', {});
    var pymParent = new pym.Parent('my-iframe-container', '{{ site.baseurl }}/1-3-1-iframe', {});
</script>
````

**Comment by christoph-LSN** [comment]
@brockfanning 

Adding a second var and a second div works for two or more indicators.

```
<script type="text/javascript" src="https://pym.nprapps.org/pym.v1.min.js"></script>
<div id="my-iframe-container"></div>
<div id="my-iframe-container-2"></div>
<script>
    var pymParent = new pym.Parent('my-iframe-container', '{{ site.baseurl }}/1-1-1-iframe', {});
    var second = new pym.Parent('my-iframe-container-2', '{{ site.baseurl }}/1-3-1-iframe', {});
</script>
```

christoph-LSN:
@brockfanning don't get i hurry with this issue. I am three weeks off for cycling the Pyrenees....

christoph-LSN:
@brockfanning Thank you for your work on this issue. My decision makers decided to have a Word&Excel solution.... So I don't need this Iframe issue anymore

sarah-beasley:
Hi @christoph-LSN , thanks for the update. We'd be keen for your opinion as you've tested this, even if you won't now actually be using. As we would still like to include the functionality as part of our next Open SDG release if possible!

sarah-beasley (approved):

PR #2035, opened 2023-07-24
Add series and unit to chart/table CSV downloads

### LINKED ISSUE #2022: Add units and series to data download
When you download data from an indicator page, there is no indication of what series and units the numbers in the .csv file are referring to. Add a units column and a series column to the .csv file for when you download data in this way.

brockfanning:
Q | A
--- | ---
Feature branch/test site URL | [Link](http://uk-sdg-feature-branches.s3-website.eu-west-2.amazonaws.com/feature-site-unit-and-series-in-chart-download/1-4-1/)
Fixed issues | Fixes #2022 
Related version | 2.3.0-dev
Bugfix, feature or docs? |
* [ ] Includes tests
* [ ] Updated docs
* [ ] Updated config forms
* [ ] Added CHANGELOG entry

### DIFF
--- a/_includes/assets/js/view/chartHelpers.js
+++ b/_includes/assets/js/view/chartHelpers.js
@@ -139,7 +139,7 @@ function setPlotEvents(chartInfo) {
         $(VIEW._legendElement).html(generateChartLegend(VIEW._chartInstance));
     });
 
-    createDownloadButton(chartInfo.selectionsTable, 'Chart', chartInfo.indicatorId, '#chartSelectionDownload');
+    createDownloadButton(chartInfo.selectionsTable, 'Chart', chartInfo.indicatorId, '#chartSelectionDownload', chartInfo.selectedSeries, chartInfo.selectedUnit);
     createSourceButton(chartInfo.shortIndicatorId, '#chartSelectionDownload');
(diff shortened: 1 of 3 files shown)


sarah-beasley (changes requested):
Hi Brock, thanks for working on this! I've tested when there is only one series, and choosing a series - both work great showing the series and units column in the csv file. I've also tried selected disaggregations though and the series and units column did not show. e.g. - http://uk-sdg-feature-branches.s3-website.eu-west-2.amazonaws.com/feature-site-unit-and-series-in-chart-download/1-2-1/

brockfanning:
@sarah-beasley I think that's because the [data for 1.2.1](https://github.com/ONSdigital/sdg-data/blob/develop/data/indicator_1-2-1.csv) doesn't include a "Units" column, which is what Open SDG needs to identify the units. It does have a "Unit measure" column, but Open SDG doesn't know that this is the unit. So in order for the unit to show up in the download I believe the data would need to be changed, so that the column header is 'Units' instead of 'Unit measure'.

brockfanning:
@sarah-beasley Nice catch, I think that should be fixed and ready for another test now, if you refresh the feature branch.

sarah-beasley (approved):
Thanks Brock, think this looks all good now!

